PR #8, opened 2021-10-15
Taking dependencies from another repo

PooyaKharamesh:
Deliverable 413482: Dependencies between repositories

Allow the apps in a github-go repository to have dependencies on another repo (own or foreign)
Either:
current - current relased version
latest - latest build (not yet released)
or a specific version
Download and place apps correctly and include in -installApps when running Run-AlPipeline

### DIFF
--- a/Actions/ReadSecrets/ReadSecrets.ps1
+++ b/Actions/ReadSecrets/ReadSecrets.ps1
@@ -1,7 +1,8 @@
 Param(
     [string] $settingsJson = '{"keyVaultName": ""}',
     [string] $keyVaultName = "",
-    [string] $secrets = ""
+    [string] $secrets = "",
+    [bool] $updateSettingsWithValues = $false
 )
 
 $ErrorActionPreference = "Stop"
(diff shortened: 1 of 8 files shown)


Actions/ReadSecrets/ReadSecrets.ps1
```diff
@@ -128,20 +163,24 @@ try {
...
+        $outSecrets += @{ ""$($secretSplit[0])"" = """" }
+    }) -join ', ')"
     }
 
     $outSecretsJson = $outSecrets | ConvertTo-Json -Compress
-    Add-Content -Path $env:GITHUB_ENV -Value "RepoSecrets=$OutSecretsJson"
+    Add-Content -Path $env:GITHUB_ENV -Value "RepoSecrets=$outSecretsJson"
+
+    $outSettingsJson = $outSettings | ConvertTo-Json -Compress
+    Add-Content -Path $env:GITHUB_ENV -Value "Settings=$OutSettingsJson"
```
freddydk, line 181: The code here adds the settings to ENV and overrides the settings set by readsettings. I think that is fine.
You don't however set the output value, by writing this to host ::set-output -name=settings::settingsinjson
This means that the output defined in yaml is never set and you cannot reference the settings through the outputs.
ENV variables only works inside the same job, referencing the outputs through outputs is useful when using settings and secrets across jobs, one job can define the output of an action as the output from the job.
Right now, you don't use that as you use env:settings in runpipeline with.
I think you should set the output value here as well and use that and I actually think that both readsettings and readsecrets should stop emitting the full combined settings as ENV - and readsecrets should emit reposecrets as outputs as well instead of ENV. (sorry for being the one pulling you down the ENV path:-))

Actions/ReadSecrets/ReadSecrets.ps1
```diff
@@ -50,10 +75,20 @@ try {
...
+                }
+            } 
+
+            $outSettings.appDependencyProbingPaths
+        }
     }
     catch {
-        OutputError -message "Error reading from GitHub Secrets. Error was $($_.Exception.Message)"
-        exit
+        throw $_
```
freddydk, line 91: The reason for the try/catch is to add some context to where the error happens and to make errors just a tad more "user" freindly.

Actions/ReadSecrets/ReadSecrets.ps1
```diff
@@ -50,10 +75,20 @@ try {
                 }
             }
         }
+
+        if ($updateSettingsWithValues) {
+            $outSettings.appDependencyProbingPaths | 
+            ForEach-Object {
+                if ($($_.authTokenSecret)) {
+                    $_.authTokenSecret = GetGithubSecret -secretName $_.authTokenSecret
```
freddydk, line 83: This approach requires the secrets to be in github secrets and you throw an error if it isn't.
We should do the same as with the secretscollection - take the ones we can from Github secrets and optionally the rest from the KeyVault (if defined) only if appDependencyProbingpaths secrets are not found in either - we should output an error.
I do agree that missing secrets from probing paths should emit an error (which is different from what the secret collection does)

Templates/AppSource App/.github/workflows/CICD.yaml
```diff
@@ -110,18 +110,20 @@ jobs:
           project: ${{matrix.project}}
 
       - name: Read secrets
+        id: readsecrets
```
freddydk, line 113: I don't think the readsecrets id is used?

Templates/Per Tenant Extension/.github/workflows/CICD.yaml
```diff
@@ -110,18 +110,20 @@ jobs:
           project: ${{matrix.project}}
 
       - name: Read secrets
+        id: readsecrets
```
freddydk, line 113: I don't think the readsecrets id is used?

PR #18, opened 2021-11-02
Misc

freddydk:


freddydk:
Bugs found when creating the walkthrough

PR #44, opened 2021-12-03
Improving error message on adding a new app 

PooyaKharamesh:
if a malformed url is provided to the workflow it fails when it tries to commit the changes. Here is an improvement to make sure the correct error message is poped up.
image


Actions/AddExistingApp/AddExistingApp.ps1
```diff
@@ -20,6 +20,10 @@ function getfiles {
 
     $path = Join-Path $env:TEMP "$([Guid]::NewGuid().ToString()).app"
     Download-File -sourceUrl $url -destinationFile $path
+    if (!(Test-Path -Path $path)) {
+        throw "could'nt download the file."
```
freddydk, line 24: "could not" or "couldn't"

freddydk (approved):

PR #51, opened 2021-12-07
Minor improvement on error messages

PooyaKharamesh:
General improvement on error messages docs and ...

### DIFF
--- a/Actions/AddExistingApp/AddExistingApp.ps1
+++ b/Actions/AddExistingApp/AddExistingApp.ps1
@@ -3,7 +3,7 @@ Param(
     [string] $actor,
     [Parameter(HelpMessage = "The GitHub token running the action", Mandatory = $false)]
     [string] $token,
-    [Parameter(HelpMessage = "Specifies the parent telemetry scope for the Telemetry signal", Mandatory = $false)]
+    [Parameter(HelpMessage = "Specifies the parent telemetry scope for the telemetry signal", Mandatory = $true)]
     [string] $parentTelemetryScopeJson = '{}',
     [Parameter(HelpMessage = "Project name if the repository is setup for multiple projects", Mandatory = $false)]
     [string] $project = '.',
--- a/Actions/AddExistingApp/AddExistingApp.ps1
+++ b/Actions/AddExistingApp/AddExistingApp.ps1
@@ -187,11 +187,11 @@ try {
                         $SettingsJson.appFolders += @($folderName)
                     }
                 }
+
                 $SettingsJson | ConvertTo-Json -Depth 99 | Set-Content -Path $settingsJsonFile -Encoding UTF8
             }
             catch {
-                OutputError -message "$ALGoSettingsFile is wrongly formatted. Error is $($_.Exception.Message)"
-                exit
+                throw "$ALGoSettingsFile is malformed. Error: $($_.Exception.Message)"
             }
 
             # Modify workspace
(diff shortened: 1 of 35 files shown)


Actions/PipelineCleanup/PipelineCleanup.ps1
```diff
@@ -1,7 +1,7 @@
 Param(
     [Parameter(HelpMessage = "Project folder", Mandatory = $false)]
     [string] $project = ".",
-    [Parameter(HelpMessage = "Specifies the parent telemetry scope for the Telemetry signal", Mandatory = $false)]
+    [Parameter(HelpMessage = "Specifies the parent telemetry scope for the telemetry signal", Mandatory = $true)]
```
freddydk, line 4: If you set it to true here, you also need to set it to true in the action.yaml file

Actions/CheckForUpdates/CheckForUpdates.ps1
```diff
@@ -140,18 +140,13 @@ try {
...
 
     if (-not $update) {
         if (($updateFiles) -or ($removeFiles)) {
-            OutputWarning -message "Updated AL-Go System Files are available for your repository, please run the Update AL-Go System Files workflow"
+            OutputWarning -message "There are updates for your AL-Go system, run 'Update AL-Go System Files' workflow to download the latest version of AL-Go."
         }
         else {
-            Write-Host "No updated AL-Go System Files are available"
+            Write-Host "Your repository runs on the latest version of AL-Go System."
         }
```
freddydk, line 150: Technically it is wrong to call it the latest version of AL-Go System
A repo that was created using v0.1 might be up-to-date with v0.1, but a v0.2 exists.
This means that he would have the latest files in v0.1 but a newer version is available (not yet detected by the script)

Actions/AddExistingApp/action.yaml
```diff
@@ -13,7 +13,7 @@ inputs:
     required: false
     default: ${{ github.token }}
   parentTelemetryScopeJson:
-    description: Specifies the parent telemetry scope for the Telemetry signal
+    description: Specifies the parent telemetry scope for the telemetry signal
     required: false
```
freddydk, line 17: this needs to be true if you change the .ps1 required to be true - and then the default should be removed.

Actions/AddExistingApp/AddExistingApp.ps1
```diff
@@ -3,7 +3,7 @@ Param(
     [string] $actor,
     [Parameter(HelpMessage = "The GitHub token running the action", Mandatory = $false)]
     [string] $token,
-    [Parameter(HelpMessage = "Specifies the parent telemetry scope for the Telemetry signal", Mandatory = $false)]
+    [Parameter(HelpMessage = "Specifies the parent telemetry scope for the telemetry signal", Mandatory = $true)]
     [string] $parentTelemetryScopeJson = '{}',
```
freddydk, line 7: if it is mandatory - there shouldn't be a default.
I think you should leave those as mandatory false.

freddydk:
Unit tests are failing - but the job succeeds???
image

Actions/PipelineCleanup/PipelineCleanup.ps1
PooyaKharamesh: thinking again about it I dont believe that it should be mandatory. Ill fix it accordingly

Actions/CheckForUpdates/CheckForUpdates.ps1
PooyaKharamesh: I also thought about it. but if scripts is not detecting any new update it means that they run on the latest. IMO we should update the script to detect a new branch and report it?

Actions/AddExistingApp/action.yaml
PooyaKharamesh: Rethinking about it I believe it shouldnt be mandatory so Ill fix it accordingly

PR #83, opened 2022-04-13
Update AL-Go-Helper.ps1

pzentner:
Fixed Typo

### DIFF
--- a/Actions/AL-Go-Helper.ps1
+++ b/Actions/AL-Go-Helper.ps1
@@ -757,7 +757,7 @@ function CloneIntoNewFolder {
 
 function CommitFromNewFolder {
     Param(
-        [string] $severUrl,
+        [string] $serverUrl,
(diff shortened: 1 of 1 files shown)


freddydk:
Kind of strange that it works when there is a spelling mistake like this...:-)

pzentner:
I guess that means that it's not really needed?

I only ran into it, because I got a warning when running ScriptAnalyzer over it and it complained about a variable not being used. There are also a bunch of other complaints, but those are regarding usage of "foreach" instead if "foreach-object" and some others that can be ignored, IMO.

freddydk:
The reason why it works is, that this function is called in 4 places - every time from the global scope and every time calling the function with $serverUrl as the variable to transfer - meaning that the variable in the function is already defined in global scope.

Thanks for reporting - will merge the fix

PR #88, opened 2022-04-23
Modified .gitignore entry "BuildOutput.txt" to "BuildOutput.xml"

JLSolutionsNET:


JLSolutionsNET:
BuildOutput.xml is always generated after each local build. I have not yet noticed any BuildOutput.txt file.

BuildOutputXml

freddydk:
The problem here is that you have a file called .xml (this is clearly not an .xml file)
Is this file created when you build in VS Code?

AL-Go for GitHub does create a BuildOutput.txt during builds in workflows - you don't have anything in your vs code settings?

freddydk:
I can see that this happens when running localdevenv - will check why

freddydk:
The problem is in ContainerHelper.
Will fix this asap.

JLSolutionsNET:
No need to merge

PR #124, opened 2022-05-12
Fix incremental app version change

ashirokikh:
Hi folks,

This is to convert old version string to the Version type so that the Major and Minor properties can be updated.

**Steps to reproduce**

Run the Increment Version Number using `+0.1` as Updated Version Number parameter.

image

This also fixes the Create release workflow when the app version is incremented.

All tests passed.

Thanks!


freddydk (approved):

freddydk:
Thanks - will deploy to preview later today

PR #138, opened 2022-05-26
Minor Formatting change to Readme,md

BW-PA:
Just some markdown formatting to improve readability of the links in the See also section.  Judging by the way you had formatted it originally I think you intended to have it as a list anyway?  Can cancel this PR if that wasn't the case.

Also used this as an exercise for following your steps outlined in "Contributing"

Before
image

After
image


freddydk:
Thanks

Issue #131, opened 2022-05-18
Keyvaultsecrets with umlauts are passed incorrectly and unmasked


juliandittmann:
Steps to reproduce:
- Create a secret with umlauts (ö,ä,ü, ... )
- Run CI/CD workflow

Note: I used a keyvault.

The secrets are recieved correctly from the vault but somehow are passed incorrectly. 
This results in unmasking secrets and futher errors.

Example:
![image](https://user-images.githubusercontent.com/50719320/169031779-10f2dea3-790d-4755-8109-6e1337b0e68a.png)

freddydk:
I have spent a lot of time trying to figure out a way to make this work - and I will have to give up.
I do see this as a bug in GitHub, which I assume they will fix at some point in time.
At the same time, it is crucial that secrets are not revealed by the system, so from the next version of AL-Go - you will get this error if a secret contains characters, which are outside the range of 32..126
![image](https://user-images.githubusercontent.com/10775043/170807870-c0981cc1-56dd-44f5-8157-ee82ba0634e5.png)

One workaround would be to implement support for some secrets could be base64 encoded (like provide ::<base64string>:: as your secret value and I will decode when used). This would be quite a lot of work, and might not be needed. I will assess based on feedback.

freddydk:
I will mark the bug as fixed, as it will no longer reveal your secrets, knowing that you will have to change your URL to NOT contain special characters.

freddydk:
Released in preview

freddydk:
Shipped in v1.5

PR #157, opened 2022-06-08
Removing umlaut limitation in secrets

juliandittmann:
Intro: 

Secrets with umlauts were forbidden due to #131.

Change:
 
I removed the limitation by converting secrets to base64 and back.
The idea was mentioned by @freddydk in #131.

I am looking for a codereview.


freddydk:
Yeah, I tried something similar, but was unable to find a mechanism where the secret isn't exposed.
The big problem is really that when the secret is used later - it might be exposed in the log - and since we need the secret as non-base64. In your case - it might be that it doesn't show up exactly in this area - but then it might show up in a different area.
And since exposing secrets might reveal classified information, I found no other way than blocking secrets with special characters and I am afraid that this PR doesn't solve the problem fundamentally.

The problem (as I see it) is a bug in GitHub.

juliandittmann:
I also think it is definitely a bug in GitHub. I tracked down the problem to [yaml](https://github.com/microsoft/AL-Go-PTE/blob/11679de548a38fbe01c8484b77c2f9d6fe9a4bb4/.github/workflows/CICD.yaml#L120). 

> SecretsJson: ${{ env.RepoSecrets }}

The secrets are correctly formatted in env.RepoSecrets. After they are passed to the actions it breaks. I dont think i can fix the Github bug soon.


Another idea would be handling the secrets with[outputs](https://github.com/marketplace/actions/set-output). Also just a workaround. Other actions for example [Azure/get-keyvault-secrets](https://github.com/Azure/get-keyvault-secrets) have taken this way. I think it is quite unhandy to manage secrets like this and it is lacking extensibility.

Note: I obviously have secrets with umlauts that i can not change anymore and it is not the _licensefileUrl_.

freddydk:
Which secret are we talking about?

juliandittmann:
My codeSignCertificatePassword contains umlauts.

Meanwhile i submitted the bug to GitHub. I will post updates here?

freddydk:
That is a real concern - I will identify a solution asap

juliandittmann:
Update from GitHub support: 

> 
> Thank you for reaching out to GitHub Support! I apologize for the delayed response here.
> 
> I did a bit of testing with this scenario today and noticed that the behavior you are experiencing with umlauts is specific to Powershell. The following Actions workflow outputs characters in powershell, which looks to return various symbols instead of the umlauts - just like you are seeing in your workflow:
> 
> ```
> on: [push]
> 
> defaults:
>   run:
>     shell: powershell
> 
> jobs:
>   output:
>     runs-on: windows-latest
>     steps:
>       - run: echo 'aäoöuü'
> ```
> ```
> > Run echo 'aäoöuü'
> aÃ¤oÃ¶uÃ¼ 
> 
> ```
> 
> 
> From some other posts I found regarding this interaction, it appears that powershell's InputEncoding and OutputEncoding need to be configured to `UTF-8`, which looks to be the default on the Actions runner:
> 
> ```
> >$OutputEncoding = [console]::OutputEncoding
> >$InputEncoding = [console]::InputEncoding
> >  
> >echo $OutputEncoding
> >echo $InputEncoding
> 
> BodyName          : utf-8
> EncodingName      : Unicode (UTF-8)
> HeaderName        : utf-8
> WebName           : utf-8
> WindowsCodePage   : 1200
> IsBrowserDisplay  : True
> IsBrowserSave     : True
> IsMailNewsDisplay : True
> IsMailNewsSave    : True
> IsSingleByte      : False
> EncoderFallback   : System.Text.EncoderReplacementFallback
> DecoderFallback   : System.Text.DecoderReplacementFallback
> IsReadOnly        : True
> CodePage          : 65001
> 
> BodyName          : utf-8
> EncodingName      : Unicode (UTF-8)
> HeaderName        : utf-8
> WebName           : utf-8
> WindowsCodePage   : 1200
> IsBrowserDisplay  : True
> IsBrowserSave     : True
> IsMailNewsDisplay : True
> IsMailNewsSave    : True
> IsSingleByte      : False
> EncoderFallback   : System.Text.EncoderReplacementFallback
> DecoderFallback   : System.Text.DecoderReplacementFallback
> IsReadOnly        : True
> CodePage          : 65001
> ```
> I'm still going to look a bit into what can be done at the workflow-level to potentially set the necessary encoding, but this doesn't look to be exclusive to Actions, just Powershell.

freddydk:
If I try this in various PowerShell "runners" on my laptop:
image
and in PowerShell ISE:
image
and in VS Code
image

freddydk:
an interesting observation in a workflow:

Write-Host with a literal in code does not work:
image

If the literal is Base64 - then the output works:
image

So, it doesn't seem to be an output problem - the string seems to have a wrong value in the inline script

freddydk:
Now I think I know what is wrong...
image
Whenever you use string literals in PowerShell, then the string becomes the binary representation of the UTF8 string.
The string contains 9 bytes (as chars) which are the 9 bytes, which the UTF8 string consists of.

I might be able to create a fix for this and allow secrets with special characters.

freddydk:
I actually think that your original suggestion of carrying the secrets as base64 encoded is good and likely the right approach.
It just couldn't stand alone, we also had to escape some characters.
I was then led off in all kinds of directions due to other challenges with encoding/decoding.

juliandittmann:
I also went full cycle and I am not yet satisfied. 
I will keep in contact with GitHub support. Perhaps a solution will emerge there. 
Your investigation was very helpful to mine. I'll probably sink a few more hours into it.

Meanwhile I updated the pullrequest.

freddydk:
You found most places:-)

I have submitted PR to main with support for this - running end 2 end tests now to see if anything breaks.
Will deploy this to preview if/when all tests are passing.

juliandittmann:
Implemented with #166

freddydk:
Thanks for your investigations, persistence and ideas - as you can see - your original PR was the right idea:-)

PR #217, opened 2022-09-29
PublishToAppsource doesn't work on github hosted runner.

jonaswre:
> Error: Deliver action failed. Error: The term 'New-AzStorageContext' is not recognized as the name of a cmdlet, function, script file, or operable program.

This is the error I got when running "Publish To AppSource" Action. 

I've had the same issue in DevOps a few days ago. 
The issue is, the correct module is not in the path.

This PR adds the path from the following documentation to the PSModulePath. 

https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md#azure-powershell-modules

### DIFF
--- a/Actions/Deliver/Deliver.ps1
+++ b/Actions/Deliver/Deliver.ps1
@@ -27,6 +27,11 @@ $bcContainerHelperPath = $null
 
 # IMPORTANT: No code that can fail should be outside the try/catch
 
+# Load the right module if on hosted runner.
+if(Test-Path -Path "C:\Modules\az_7.5.0") {
+    $env:PSModulePath += ';C:\Modules\az_7.5.0'
+}
+
 try {
     . (Join-Path -Path $PSScriptRoot -ChildPath "..\AL-Go-Helper.ps1" -Resolve)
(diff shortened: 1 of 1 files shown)


freddydk:
This error is actually in ContainerHelper.
Everywhere else - I test whether New-AzContext is available and if it is, I set an alias:

```
    if (!(get-command New-AzureStorageContext -ErrorAction SilentlyContinue)) {
        Set-Alias -Name New-AzureStorageContext -Value New-AzStorageContext
        Set-Alias -Name Set-AzureStorageBlobContent -Value Set-AzStorageBlobContent
    }
```

And then I use New-AzureStorageContext.

The GitHub hosted runners have the AzureStorage module pre-loaded - most self-hosted runners have the newer AzStorage module.

The Next containerhelper will be out in an hour and will fix this issue.

(Hope that I can get to your PRs on containerhelper next week - sorry for the wait)

freddydk:
Fixed in the latest containerhelper

jonaswre:
Thanks for the quick fix Freddy.

After I opened the PR I realized that there might be other situation where a fix might be nessary. And setting an alias is probably the best solution in this case. 

Regarding my PR in navcontainerhelper. What settings do I have to set to run the Actions in my fork? Is there documention that I haven't found as well?

PR #218, opened 2022-10-03
Add updateBranch parameter to CheckForUpdates action

mazhelez:
Issue: Running the "Update AL-Go System Files" with `directCommit` set to 'Y' always creates a PR that targets the main branch insted of the branch the action was invoked on.

Solution: Add an `updateBranch` parameter to `CheckForUpdates` action and corresponding script to use as a target branch when creating a PR with the updates. By default, the `updateBranch` should be the branch the action is run on.

### DIFF
--- a/Actions/CheckForUpdates/CheckForUpdates.ps1
+++ b/Actions/CheckForUpdates/CheckForUpdates.ps1
@@ -11,6 +11,8 @@ Param(
     [string] $templateBranch = "",
     [Parameter(HelpMessage = "Set this input to Y in order to update AL-Go System Files if needed", Mandatory = $false)]
     [bool] $update,
+    [Parameter(HelpMessage = "Set the branch to update", Mandatory = $false)]
+    [string] $updateBranch,
     [Parameter(HelpMessage = "Direct Commit (Y/N)", Mandatory = $false)]
     [bool] $directCommit    
 )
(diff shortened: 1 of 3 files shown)


freddydk:
Shouldn't there be a change in UpdateGitHubGoSystemFiles.yaml from both the PTE and the AppSource templates as well?
If you made the changes in your versions of the AL-Go-PTE and/or AL-Go-AppSource, then the change would be collected to f.ex.
https://github.com/microsoft/AL-Go/blob/main/Templates/Per%20Tenant%20Extension/.github/workflows/UpdateGitHubGoSystemFiles.yaml

mazhelez:
> Shouldn't there be a change in UpdateGitHubGoSystemFiles.yaml from both the PTE and the AppSource templates as well? If you made the changes in your versions of the AL-Go-PTE and/or AL-Go-AppSource, then the change would be collected to f.ex. https://github.com/microsoft/AL-Go/blob/main/Templates/Per%20Tenant%20Extension/.github/workflows/UpdateGitHubGoSystemFiles.yaml

No, no changes are needed there. The default branch will be `github.ref_name` (the branch the workflow is invoked on).
Initially, I didn't want to add it as a parameter for action `CheckForUpdates` because why would anyone want to use another branch. However, it has to be a parameter for the script `CheckForUpdates` and there is a test that verifies that both the action and the script have the same parameters.

freddydk (approved):

freddydk:
Got it, thanks

freddydk:
End 2 End tests are failing because no NextMajor is available at this time - will wait for that.
(Should happen today)

PR #305, opened 2022-11-24
Add GitHub pages to AL-Go

freddydk:
Deploy documentation from the AL-Go repo in GitHub pages like this:
image

(theme can be changed if we decide to)


README.md
```diff
@@ -54,7 +53,7 @@ contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additio
 
 ## See also
 
-- [Developing Extensions in AL](../developer/devenv-dev-overview.md)
+- [Developing Extensions in AL](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-dev-overview)
```
mazhelez, line 56:
```suggestion
- [Developing Extensions in AL](https://learn.microsoft.com/dynamics365/business-central/dev-itpro/developer/devenv-dev-overview)
```
The locale part can be omitted, an automatic one will be picked based on the current location. For me, it picks `en-gb`.

Also, it's generally good practice to use forward links.

Scenarios/UseAzureKeyVault.md
```diff
@@ -1,11 +1,12 @@
...
+*Prerequisites: A completed [scenario 6](UpdateAlGoSystemFiles.md), an Azure KeyVault and you will need to follow the guidelines on how to connect to an Azure KeyVault as specified here: [https://docs.microsoft.com/en-us/azure/developer/github/github-key-vault](https://docs.microsoft.com/en-us/azure/developer/github/github-key-vault). Add your KeyVault name to the the JSON construct from this walkthrough (using **“keyVaultName” : “{your keyvault name}”**) and add this JSON construct as a repository secret called AZURE_CREDENTIALS. You can also specify the KeyVault name in the AL-Go settings file if you do not wait to mess with the JSON construct.*
```
mazhelez, line 2: Isn't `doc.microsoft.com` currently migrated to `learn.microsoft.com`? Or was it `go.microsoft.com`? 🤔

freddydk: Correct - https://learn.microsoft.com/azure/developer/github/github-key-vault would be better

freddydk: made the change

EmilDamsbo:
How come we are separating the docs for AL-Go to this repo? Shouldn't it be in the official Learn channel? IMO this is splitting information unnecessarily.

freddydk:
> How come we are separating the docs for AL-Go to this repo? Shouldn't it be in the official Learn channel? IMO this is splitting information unnecessarily.

The docs have always been in the repo and copied to learn. For an open-source offering like AL-Go for GitHub, we cannot imagine NOT having any documentation right here in the repo and GitHub pages is the standard way of hosting documentation for GitHub repos.

I think the way we should implement this is that the documentation on GitHub will be constantly updates (meaning it will be for the preview version) and we will update learn when we ship, so that will be the official documentation.

freddydk:
One more thing.

Today, the workshop (https://aka.ms/algoworkshop) is a PPT placed in the repo as well - I want to change that to .md, backed by a list of youtube videos.

freddydk:
Thinking some more - I do also get Emils point - the docs doesn't look like Microsoft docs and we should solve 
Let's merge the PR (all the changes to the .md files are good) and disable the workflow.
Then I will have a chat with Susanne about the strategy for this going forward.

EmilDamsbo:
> Thinking some more - I do also get Emils point - the docs doesn't look like Microsoft docs and we should solve Let's merge the PR (all the changes to the .md files are good) and disable the workflow. Then I will have a chat with Susanne about the strategy for this going forward.

I can agree with this approach. No further comments.

EmilDamsbo (approved):

mazhelez (approved):

PR #327, opened 2022-12-26
Restructure end 2 end tests

freddydk:
This PR adds a third E2E testing option: Scenarios (added to Test AL-Go End-2-End and Test AL-Go Upgrades)
You can also select which of these to run AND you can select which version of BcContainerHelper should be used for running the tests. Previously, this was always preview.
- Test AL-Go End-2-End tests an end-2-end scenario for all combinations of Public/Private, Windows/Linux, Single/Multiproject and PTE/AppSource.
- Test AL-Go Upgrade tests that all prior released versions of AL-Go for GitHub can be upgraded to the latest version and run CI/CD successfully.
- Test AL-Go scenarios are for cross-cutting scenarios, like:
  - UseDependencies - create a repo with multiple projects and set UseDependencies to modify CI/CD and other build workflows to build projects in the right order
  - GitHubPackages - create 3 repositories using GitHub Packages as dependency resolver and check that artifacts are built properly
  - buildModes - create a repository, set buildModes and test that generated artifacts are as expected.

Additionally, another step in the E2E.yaml has been added, which displays the repo-URL before running the runtests:
image
Clicking this link while the test is running allows you to follow the test progress in the target repo.

Furthermore, if you want to run end-2-end tests locally, you need to import BcContainerHelper and then either need to set some global variables:

```
$global:E2EgitHubOwner = "businesscentralapps"
$global:SecureE2EPAT = $E2EPATSecret.SecretValue
$global:SecureAdminCenterApiToken = $AdminCenterApiCredentialsSecret.SecretValue
$global:SecureLicenseFileUrl = $LicenseFileUrlSecret.SecretValue
$global:SecureInsiderSasToken = $InsiderSasTokenSecret.SecretValue
```

Or you need to invoke the runtests.ps1 script with these parameters, setting the right parameters.

SecureE2EPAT is a SecureString, containing a Personal Access token with these permissions: _admin:org, delete:packages, delete_repo, repo, workflow, write:packages_
AdminCenterApiToken is a securestring containing an AuthContext, created by New-AlGoAuthContext with access to your admin center api
SecureLicenseFileUrl is a securestring containing a direct download url for a build license file
SecureInsiderSasToken is a securestring containing the Insider Sas Token from https://aka.ms/collaborate
E2EgithubOwner is the organization in which all the repositories will be created and run (this should not be your user account)
When running all tests, the organization might hit limits if the organization is a free account.

Lastly, this PR also cleans up CRLF / LF and JSON formatting issues when working cross platforms.
GitHub natively uses LF and up until now, AL-Go scripts have been converting to CRLF and tried to ensure consistency.
PS5 formats JSON differently than PS7.
With this PR - everything uses LF and PS7 JSON formatting.

GitHubOwner secret is no longer used - instead there is a parameter on the end 2 end testing workflow, specifying which organization should be used. Also OrgPAT secret is no longer used for access - instead E2EPAT is used (which needs access to this org.). Default is to use the "current" github owner as githubOwner - and if this is the case, max-parallel will be 8 - making sure that there is sufficient vacant workflows to avoid deadlocking.
There is a file in e2eTests called orgmap.json, which maps known orgs for ensuring that we don't try to create tmp repos under f.ex. microsoft.


.github/workflows/E2E.yaml
```diff
@@ -73,36 +89,81 @@ jobs:
...
             $releasesJSon = "[$($releases | ConvertTo-Json -compress)]"
           }
           else {
             $releasesJSon = $releases | ConvertTo-Json -compress
           }
           Add-Content -Path $env:GITHUB_OUTPUT -Value "releases=$releasesJson"
           Write-Host "releases=$releasesJson"
-          Add-Content -Path $env:GITHUB_OUTPUT -Value "releasesCount=$($releases.Count)"
-          Write-Host "releasesCount=$($releases.Count)"
+          $scenarios = @(Get-ChildItem -path (Join-Path $ENV:GITHUB_WORKSPACE "e2eTests/scenarios/*/runtest.ps1") | ForEach-Object { $_.Directory.Name })
```
mazhelez, line 122: Does `$_.Directory.Name` work well in PS in ubuntu?

Actions/AL-Go-Helper.ps1
```diff
@@ -1687,23 +1688,65 @@ Function AnalyzeProjectDependencies {
...
+                $depProject = $projects | Where-Object { $_ -ne $project -and $appDependencies."$_".apps -contains $dependency }
+                # Add this project and all projects on which that project has a dependency to the list of dependencies for the current project
+                $depProject | ForEach-Object {
+                    $_
+                    if ($projectDependencies.Value.ContainsKey($_)) {
+                        $projectDependencies.value."$_"
+                    }
+                }
             } | Select-Object -Unique)
+            # foundProjects now contains all projects that the current project has a dependency on
```
mazhelez, line 1728: I cannot find `foundProjects`.

.github/workflows/E2E.yaml
freddydk: Yes, works fine

Actions/AL-Go-Helper.ps1
freddydk: Fixed - should have been foundDependencies

Internal/Deploy.ps1
```diff
@@ -199,7 +201,7 @@ try {
                     New-Item $srcFilePath -ItemType Directory | Out-Null
                 }
                 Write-Host "$dstFile -> $srcFile"
-                $lines = ([string](Get-Content -Raw -path $dstFile)).Split("`n")
+                $lines = ([string](Get-Content -Raw -path $dstFile -Encoding UTF8)).Replace("`r", "").TrimEnd("`n").Split("`n")
```
mazhelez, line 204: I see this pattern repeated ```.Replace("`r", "").TrimEnd("`n").Split("`n")``` in a couple of places.

Consider adding `Get-ContentLF` to match `Set-ContentLF`.

e2eTests/e2eTestHelper.psm1
```diff
@@ -304,9 +402,13 @@ function CommitAndPush {
 
 function MergePRandPull {
     Param(
+        [string] $repository,
         [string] $branch = "main"
     )
 
+    if (!$repository) {
+        $repository = $defaultRepository
```
mazhelez, line 410: Can't `$defaultRepository` be set as default value for `$repository` param?

mazhelez (approved):
All good on my side. Left a couple of minor comments.

Also, do not forget to update the ARM template for the self-hosted runners to include PS7 by default.

e2eTests/e2eTestHelper.psm1
freddydk: Default parameters are only used if the parameter isn't specified.
In this case, the repository might be transferred as empty.

Internal/Deploy.ps1
freddydk: Good idea - have done so.
Re-running end-2-end tests to ensure I didn't make a copy/paste error

Issue #215, opened 2022-09-26
LocalDevEnv.ps1 fails with "Illegal characters in path."


buzzwick:
We created an AL-Go repository for our AppSource extension code. When I clone that repo and run LocalDevEnv.ps1 from inside VSCode (we do this routinely for Docker containers), it opens a separate PowerShell window that looks like it loads the BC modules but then immediately closes. If I open PowerShell as admin and run the script, it runs but fails consistently as below. Questions:
1. Any clue as to the source of the "illegal character"?
2. I can create a Docker container with the latest BC version--but do I need to use this script for the container if I am working with code from the AL-Go repository?
3. Is there some reason this script does not run within the VSCode PowerShell terminal?

  _                     _   _____             ______
 | |                   | | |  __ \           |  ____|
 | |     ___   ___ __ _| | | |  | | _____   __ |__   _ ____   __                                                         | |    / _ \ / __/ _` | | | |  | |/ _ \ \ / /  __| | '_ \ \ / /                                                         | |____ (_) | (__ (_| | | | |__| |  __/\ V /| |____| | | \ V /                                                          |______\___/ \___\__,_|_| |_____/ \___| \_/ |______|_| |_|\_/                                                                                                                                                                                  This script will create a docker based local development environment for your project.                                  
NOTE: You need to have Docker installed, configured and be able to create Business Central containers for this to work.
If this fails, you can setup a cloud based development environment by running cloudDevEnv.ps1

All apps and test apps will be compiled and published to the environment in the development scope.
The script will also modify launch.json to have a Local Sandbox configuration point to your environment.

Checking ..\.github\AL-Go-Settings.json
Checking .AL-Go\settings.json
Checking ..\.github\.settings.json
Checking .AL-Go\.settings.json
Checking .AL-Go\dbaldwin.settings.json
Checking System Requirements

Container name
--------------
Please enter the name of the container to create (default bcserver) bcappsource
bcappsource selected

Authentication mechanism for container
--------------------------------------
a Windows Authentication
b Username/Password authentication

Select authentication mechanism for container (default b)
Username/Password authentication selected

LicenseFileUrl
--------------
When developing PTEs, you can optionally specify a developer licensefile with permissions to object IDs of your dependant apps

Local path or a secure download URL to license file  (default none) "C:\_CAI Licenses\CAIv20.flf"
"c:\_cai licenses\caiv20.flf" selected

Downloading BcContainerHelper latest version
Using Expand-Archive
[...]

PR #337, opened 2023-01-07
Run LocalDevEnv and CloudDevEnv in VS Code

freddydk:
With the latest changes in ContainerHelper to support PS7, we can now run localDevEnv and cloudDevEnv directly in VS Code (instead of spawning a PS5.1 powershell prompt with all the problems that gives).
Fix for Issue #215

### DIFF
--- a/Templates/AppSource App/.AL-Go/localDevEnv.ps1
+++ b/Templates/AppSource App/.AL-Go/localDevEnv.ps1
@@ -15,35 +15,6 @@ Param(
 $ErrorActionPreference = "stop"
 Set-StrictMode -Version 2.0
 
-$pshost = Get-Host
-if ($pshost.Name -eq "Visual Studio Code Host") {
-    $executionPolicy = Get-ExecutionPolicy -Scope CurrentUser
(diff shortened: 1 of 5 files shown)


Templates/AppSource App/.AL-Go/localDevEnv.ps1
```diff
@@ -15,35 +15,6 @@ Param(
 $ErrorActionPreference = "stop"
 Set-StrictMode -Version 2.0
 
-$pshost = Get-Host
```
mazhelez, line 18: There a lot of common functionality in these scripts for AppSource App and PTE templates. Is there a way to extract such common functionality so that it's not duplicated?

freddydk: Majority of the template repositories are the same - we could have a common and then have all things, which are the same reside in common and only special things be in AppSource App or PTE folders - meaning that when deploying, we would grab Common + PTE to become PTE and Common + AppSource App to become AppSource App.
Make sense?

PR #339, opened 2023-01-08
Buildmodulesseperately

freddydk:
AnalyzeRepo shows a false warning when a folder contains both folders and files.


mazhelez (approved):
PR title could be better 😄

Issue #328, opened 2022-12-29
Multi AppSource Apps in one repo.


jonaswre:
Im trying to publish multiple apps in one repo to appsource.

You can only enter one productId per .AL-Go/settings.json. So I tried to make the repo multi project but now I get the following error.

```Error: RunPipeline action failed. Error: For AppSource Apps with AppSourceCop enabled, you need to specify AppSourceCopMandatoryAffixes in .AL-Go\settings.json Stacktrace: at AnalyzeRepo, D:\a\_actions\microsoft\AL-Go-Actions\v2.2\AL-Go-Helper.ps1: line 539 at <ScriptBlock>, D:\a\_actions\microsoft\AL-Go-Actions\v2.2\RunPipeline\RunPipeline.ps1: line 79 at <ScriptBlock>, D:\a\_temp\d965a261-a9b2-4f1c-91d7-c20ad8eabeb2.ps1: line 1 at <ScriptBlock>, <No file>: line 1```

https://github.com/microsoft/AL-Go-Actions/blob/53470cc1bf0524559a158d90b093170d7fb3ec76/AL-Go-Helper.ps1#L539

It seems like the analyse-repo part doesn't respect multi project AppSource Repos. 

I'm doing something wrong here?

**Comment by jonaswre** [comment]
I found a solution

- .AL-Go/settings.json - AppSourceCopMandatoryAffixes set to any value
- app1/.AL-Go/settings.json - AppSourceCopMandatoryAffixes set for the app1
- app2/.AL-Go/settings.json - AppSourceCopMandatoryAffixes set for the app2

But this feels like cheating and this solution drasticly increases the build time.

Whats the reasoning to only allow one appsource app per project?

**Comment by freddydk** [comment]
One project is one shippable entity, you can have multiple apps in one project, but they would be under one appsource app.

Seems like a bug that you cannot have a multi-project repo with 2 appsource app projects - the above "solution" makes the root into a third project - and is likely not what you want.

**Comment by jonaswre** [comment]
If you set appFolders to an empty error no harm 

Is there anyway to eliminate wasted build time here? Maybe we can cache container images if all projects have the same requirements?

Because if you capsulate your code you will often create bridge apps, which are independently deployable, but are still tightly coupled.

**Comment by jonaswre** [comment]
There is no "good" way to get dependencies between projects within the same run.

The following option is problematic

```
  "appDependencyProbingPaths":  [
        {
            "repo":  "https://github.com/<org>/<repo>",
[...]

PR #341, opened 2023-01-09
Issue #328

freddydk:
Issue #328 Allow multiple projects in AppSource App template
Change E2E.yaml to skip entire jobs instead of steps.
In E2E - Test-ArtifactsFromRun did not fail the build - it does now (and it also works properly now)


Actions/Deliver/Deliver.ps1
```diff
@@ -199,7 +198,7 @@ try {
             #Calculate the folders per artifact type
             'Apps', 'TestApps', 'Dependencies' | ForEach-Object {
                 $artifactType = $_
-                $singleArtifactFilter = "*-$refname-$artifactType-*.*.*.*";
+                $singleArtifactFilter = "$project-$refname-$artifactType-*.*.*.*";
```
mazhelez, line 201: Using `$project` might not work as it is supposed to be a path, right? Like `root/folder/subfolder/projectFolder`.

Maybe use `$projectName`? (From above: `$projectName = $project -replace "[^a-z0-9]", "-"`).

However, when we calculate artifacts names in CICD, the logic is slightly different: `$($project.Replace('\','_'))` (see [here](https://github.com/microsoft/AL-Go-PTE/blob/main/.github/workflows/CICD.yaml#:~:text=%24(%24project.Replace(%27%5C%27%2C%27_%27)))).

freddydk: Good catch.
$project should be the name of the published artifact (/ and \ converted to _).
In CI/CD it will today only have a \ but soon - they might have / as well - have added the same conversion all over.
Should refactor and create a function for this, but in many of the areas, no function library is available.

projectName is for blob storage and other things where all special characters are stripped off - for output purposes.

Actions/Deliver/Deliver.ps1
```diff
@@ -98,12 +97,14 @@ try {
 
     $projectList | ForEach-Object {
         $thisProject = $_
+        # $project should be the project part of the artifact name generated by CI/CD
```
mazhelez, line 100:
```suggestion
        # $project should be the project part of the artifact name generated from the build
```

CI/CD is a bit too vague here. After all, _Deliver_ is a part of CI/CD.

mazhelez (approved):

Issue #318, opened 2022-12-09
Make it possible to avoid syncing certain workflows


mazhelez:
Currently, if a workflow is deleted, the next run of `Update AL-Go System Files` will bring it back, polluting the view of workflow/actions.

There should be a configuration to avoid syncing workflows that are not in use.
`

freddydk:
I agree that a setting in the repo settings file, which lists unused workflows (specified by workflow filenames) would be the way to go. The Update AL-Go System Files would then remove these files if they exist (instead of comparing their content)

freddydk:
No, this should be a list of unused workflows - NOT a list of used workflows.
So, if the partner adds their own workflow or AL-Go comes out with a new workflow - they would be included (until you remove them if you like)

jonaswre:
Sorry I was just asking in general wether added action are removed on update. 

Wasn't suggesting any changes.

freddydk:
I know:-)
Actions added into an AL-Go repo are left alone by the update action (unless there is an action with the same name added in the AL-Go template).
AL-Go only touches known workflows.

PR #342, opened 2023-01-09
Issue #318

freddydk:
Issue #318
Add new Repo setting: unusedALGoSystemFiles in which you can specify an array of System Files, which won't be updated, but instead will be removed.
Example here: https://github.com/BusinessCentralApps/BingMaps.AppSource/blob/main/.github/AL-Go-Settings.json

### DIFF
--- a/Actions/CheckForUpdates/CheckForUpdates.ps1
+++ b/Actions/CheckForUpdates/CheckForUpdates.ps1
@@ -316,12 +316,15 @@ try {
                 }
 
                 $dstFile = Join-Path $dstFolder $fileName
+                $dstFileExists = Test-Path -Path $dstFile -PathType Leaf
                 if ($unusedALGoSystemFiles -contains $fileName) {
-                    if (Test-Path -Path $dstFile -PathType Leaf) {
+                    # file is not used by ALGo, remove it if it exists
+                    # do not add it to $updateFiles if it does not exist
+                    if ($dstFileExists) {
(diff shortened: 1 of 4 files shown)


Actions/CheckForUpdates/CheckForUpdates.ps1
```diff
@@ -309,9 +315,13 @@ try {
                     $srcContent = Get-ContentLF -Path $srcFile
                 }
 
-
                 $dstFile = Join-Path $dstFolder $fileName
-                if (Test-Path -Path $dstFile -PathType Leaf) {
+                if ($unusedALGoSystemFiles -contains $fileName) {
+                    if (Test-Path -Path $dstFile -PathType Leaf) {
```
mazhelez, line 320: This check is duplicated below.

Consider moving `$unusedALGoSystemFiles -contains $fileName` as a subcheck instead.

mazhelez (approved):
Any way to add a quick test with the new setting?

Actions/CheckForUpdates/CheckForUpdates.ps1
freddydk: Moving that to a subcheck - then I would have to check for the unusedALGoSystemFiles before adding the new file to updateFiles. Will make a small update and add a comment

freddydk:
> Any way to add a quick test with the new setting?

Today, the only way really to test these settings is end 2 end.
I will work on adding CI tests to these things and add tests to this

PR #350, opened 2023-01-12
PublishToEnvironment to honor DeployTo settings

freddydk:
Previously, Publish To Environment would ignore the setting.

I will add some tests to this area before refactoring to also support the new GitHub variables (later)

### DIFF
--- a/Templates/Per Tenant Extension/.github/workflows/PublishToEnvironment.yaml
+++ b/Templates/Per Tenant Extension/.github/workflows/PublishToEnvironment.yaml
@@ -87,12 +87,19 @@ jobs:
           $ErrorActionPreference = "STOP"
           Set-StrictMode -version 2.0
           $envName = '${{ steps.envName.outputs.envName }}'
+          $deployToSettingStr = [System.Environment]::GetEnvironmentVariable("DeployTo$envName")
+          if ($deployToSettingStr) {
+            $deployToSetting = $deployToSettingStr | ConvertFrom-Json
+          }
+          else {
+            $deployToSetting = [PSCustomObject]@{}
+          }
           $authContext = $null
           "$($envName)-AuthContext", "$($envName)_AuthContext", "AuthContext" | ForEach-Object {
             if (!($authContext)) {
               $authContext = [System.Environment]::GetEnvironmentVariable($_)
               if ($authContext) {
-                Write-Host "Using $_ secret"
+                Write-Host "Using $_ secret as AuthContext"
               }
             }            
           }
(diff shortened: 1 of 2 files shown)


Templates/Per Tenant Extension/.github/workflows/PublishToEnvironment.yaml
```diff
@@ -87,48 +87,69 @@ jobs:
...
-                Write-Host "Using $_ secret"
-              }
-            }            
+          if ($deployToSetting.PSObject.Properties.name -eq "EnvironmentName") {
+            $environmentName = $deployToSetting.EnvironmentName
+          }
+          else {
+            $environmentName = $null
+            "$($envName)-EnvironmentName", "$($envName)_EnvironmentName", "EnvironmentName" | ForEach-Object {
+              if (!($EnvironmentName)) {
```
mazhelez, line 116:
```suggestion
              if (!($environmentName)) {
```

Stick to one casing.

Templates/Per Tenant Extension/.github/workflows/PublishToEnvironment.yaml
```diff
@@ -87,48 +87,69 @@ jobs:
...
           $environmentName = [Convert]::ToBase64String([System.Text.Encoding]::UTF8.GetBytes(($environmentName + '${{ matrix.environment }}'.SubString($envName.Length)).ToUpperInvariant()))
-
-          $projects = [System.Environment]::GetEnvironmentVariable("$($envName)-projects")
-          if (-not $projects) {
-            $projects = [System.Environment]::GetEnvironmentVariable("$($envName)_Projects")
+          if ($deployToSetting.PSObject.Properties.name -eq "projects") {
+            $projects = $deployToSetting.projects
+          }
+          else {
+            $projects = [System.Environment]::GetEnvironmentVariable("$($envName)-projects")
```
mazhelez, line 133: This fallback mechanism looks similar to the one above (that determines the environment name).

Consider unifying it in a separate function that can be reused.

mazhelez (approved):
Left non-blocking comments.

Generally, avoid adding more complexity, especially to scripts that are part of the yaml-s (as I think it's hard to analyse such scripts). Testing and readability take a hit here.

Templates/Per Tenant Extension/.github/workflows/PublishToEnvironment.yaml
freddydk: Agree
I want to move the functionality into the Deploy Action and remove things from the yaml file altogether.

PR #355, opened 2023-01-14
Ignore Quotes, plings and spaces when entering values in localDevEnv and cloudDevEnv

freddydk:
Issue #215

### DIFF
--- a/Templates/AppSource App/.AL-Go/cloudDevEnv.ps1
+++ b/Templates/AppSource App/.AL-Go/cloudDevEnv.ps1
@@ -59,7 +59,8 @@ if (-not $environmentName) {
     $environmentName = Enter-Value `
         -title "Environment name" `
         -question "Please enter the name of the environment to create" `
-        -default "$($env:USERNAME)-sandbox"
+        -default "$($env:USERNAME)-sandbox" `
+        -trimQuotesAndSpaces
 }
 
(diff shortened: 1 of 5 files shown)


Templates/AppSource App/.AL-Go/cloudDevEnv.ps1
```diff
@@ -59,7 +59,8 @@ if (-not $environmentName) {
     $environmentName = Enter-Value `
         -title "Environment name" `
         -question "Please enter the name of the environment to create" `
-        -default "$($env:USERNAME)-sandbox"
+        -default "$($env:USERNAME)-sandbox" `
+        -trimQuotesAndSpaces
```
mazhelez, line 63: Why don't you simply add `$environmentName.trim(@('"',"'",' '))` after `Enter-Value`?

Adding complexity to a commodity function doesn't seem justified here, in my opinion.

If you insist on adding a new parameter, why about `$trimCharacters = @()` to cover potential future cases where we might want to trim parathesis `(` `)` or some other characters?

mazhelez (approved):
Added a comment with my opinion.

Also, not sure if `Enter-Value` is tested somewhere. If it is, let's add another test to cover the new parameter.
If it is not, let's add a test to cover the new parameter. ;)

Issue #356, opened 2023-01-15
Publish to AppSource fails in multi project repo


jonaswre:
```
Error: Deliver action failed. Error: AppSourceProductId needs to be specified in AppD/.AL-Go/settings.json in order to deliver to AppSource Stacktrace: at <ScriptBlock>, D:\a\_actions\microsoft\AL-Go-Actions\v2.3\Deliver\Deliver.ps1: line 418 at <ScriptBlock>, D:\a\_actions\microsoft\AL-Go-Actions\v2.3\Deliver\Deliver.ps1: line 98 at <ScriptBlock>, D:\a\_temp\6c84dd42-f6bd-4b08-bab2-b1ad63498722.ps1: line 2 at <ScriptBlock>, <No file>: line 1
```

This is the content of AppD/.AL-Go/settings.json
```
{
    "country": "de",
    "AppSourceCopMandatoryAffixes": [
       "TEST"
    ],
    "appFolders": [
        "app"
    ],
    "testFolders": [],
    "bcptTestFolders": [],
    "AppSourceProductId": "5f3a3117-d878-44b0-ad21-9986c8582cd9"
}
```
Its still based on the https://github.com/byndit/AL-Go-Issue138

P.S. obviously  I deployed our product not that repo. The structure is just the same.

kb-dynaway:
@freddydk  To not create addtional thread. Is it possible to publish to AppSource multiple apps from the same repository? I have in one repo 8 apps that need to be published to AppSource.

Is it possible to have multiple AppSourceProductsIds?

jonaswre:
@kb-dynaway 

@freddydk anwsered this in https://github.com/microsoft/AL-Go/issues/328

freddydk:
From v2.3 - yes, the AppSourceProductId needs to be in the project settings file.
But apparently there is another bug, which I have a fix for in a PR - when this makes it in, it will be added to preview

freddydk:
This fix is now available in AL-Go preview

janlehmkuhl:
Hi, I don't know if my problem is related to this customization, but we can't publish our AppSource app anymore, although all parameters are set as requested. Attached are the logs and two screenshots.

![settings json](https://user-images.githubusercontent.com/94679485/219421710-531aa318-9f56-4ff8-baa5-d4a96e79f439.png)
![errormsg](https://user-images.githubusercontent.com/94679485/219421718-0a2760a3-be75-40d3-b8d5-a52d6d1cf806.png)
[...]

Issue #171, opened 2022-06-27
Adding Test App not showing in Workspace


tscottjendev:
Should adding a Test app update the workspace?  It does not for me, wondering if this is intentional.

freddydk:
It should add it to the workspace, yes.
Will check

tscottjendev:
I see you noted you can't reproduce.  Here is the steps I did for this repo.
Created the repo from the AL-GO-PTE repo template
Added a new APP, changing the WorkSpace from "." to PTEFLOW as I wanted to test multiple apps (using the sample code).
after trying to add new app without the "PTEFLOW" workspace name, I added a new Test App (without the sample code).
The test app does not show in the workspace.

Happy to add you to the repo if that would assist you in your investigation.

tscottjendev:
yes, this field 
![image](https://user-images.githubusercontent.com/102963427/175923172-079279bb-bf6a-4b8d-9ceb-57f17369c451.png)

freddydk:
Yep - that results in a multi-project repo, which really is intended for having totally unrelated (or maybe even conflicting) apps in the same repository - basically every project ships seperatly.
If the apps have inter-dependencies - you should keep the . and have all apps in the same root project,

I will check the issue

PR #357, opened 2023-01-16
Refactor ReadSettings and add tests

freddydk:
Issue #171 create a workspace file when creating a project
Issue #356 Publish to AppSource fails in multi project repo

ReadSettings has been refactored to allow organization wide settings to be added as well. CI Tests have been added to cover ReadSettings.

The primary change is to ensure that $baseFolder is always the $baseFolder and $projectFolder is always a project folder (they are the same in a single project repo).
ReadSettings and more now have two parameters: baseFolder and project - which clearly specifies when to read project settings files and when to read repo settings files only
Comments have been added to ReadSettings and MergeCustomObjectIntoOrderedDictionary

### DIFF
--- a/Actions/AL-Go-Helper.ps1
+++ b/Actions/AL-Go-Helper.ps1
@@ -367,42 +367,6 @@ function MergeCustomObjectIntoOrderedDictionary {
     }
 }
 
-function GetBaseFolder {
-    Param(
-        [string] $folder
-    )
-
-    if (!(Test-Path (Join-Path $folder '.github') -PathType Container)) {
-        $folder = (Get-Item -Path $folder).Parent.FullName
-        if (!(Test-Path (Join-Path $folder '.github') -PathType Container)) {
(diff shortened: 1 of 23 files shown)


Actions/AL-Go-Helper.ps1
```diff
@@ -1781,3 +1814,40 @@ Function AnalyzeProjectDependencies {
...
     }
 }
+
+function GetBaseFolder {
+    Param(
+        [string] $folder
+    )
+
+    if (!(Test-Path (Join-Path $folder '.github') -PathType Container)) {
+        $folder = (Get-Item -Path $folder).Parent.FullName
```
mazhelez, line 1824: Wasn't there an issue with `Parent` in linux?

freddydk: The problem i discovered in Linux was with Get-ChildItem - but it might have been something else.
The above code works fine

mazhelez (approved):

PR #365, opened 2023-01-19
Migration docs

freddydk:
The docs are viewable in markdown here:

https://github.com/freddydk/AL-Go/blob/migrationDocs/Scenarios/MigrateFromAzureDevOpsWithoutHistory.md
https://github.com/freddydk/AL-Go/blob/migrationDocs/Scenarios/MigrateFromAzureDevOpsWithHistory.md

and the readme:

https://github.com/freddydk/AL-Go/tree/migrationDocs


Scenarios/MigrateFromAzureDevOpsWithHistory.md
```diff
@@ -0,0 +1,59 @@
+# #B Migrate a repository from Azure DevOps to AL-Go for GitHub with history
+*This walkthrough explains how to migrate a repository from Azure DevOps to AL-Go for GitHub **while preserving the full commit history**. As a sample, I will use a repository, which was setup using the CI/CD Hands On Lab, including scripts and pipelines from this.*
+
+*If you do not want to preserve the full commit history, you should use [Scenario A](MigrateFromAzureDevOpsWithoutHistory.md) or [Scenario 11](SetupCiCdForExistingAppSourceApp.md)*
+
+***Note: This walkthrough will leave your existing Azure DevOps repository untouched and you can decide to keep working in that if you like.***
+
+1. Start out by navigating to your **Azure DevOps** repository. Click the **Clone** button and click **Generate Git Credentials**.
```
mazhelez, line 8: The first few steps are generic for any repo migration. Any chance to simply link to some official docs so that we won't have to keep these steps up-to-date. I doubt the steps would change, but who knows.

mazhelez (approved):

Scenarios/MigrateFromAzureDevOpsWithHistory.md
freddydk: For these two migration steps - the only line that is the same is the Note - and I actually think that line should be in all - I don't like to click a link to read info like this.

PR #366, opened 2023-01-19
Action to import repositories based on the OldHelloWorld

jonaswre:
The action allows a user to import a repository into an al-go based repo.  

There are some possible improvements: 
1. Merge the workspace file to keep old workspace configuration. 
2. Add folders to the need configurations.
3. Be more permissive in which folders need to exist in the import.

### DIFF
--- a/Actions/AL-Go-Helper.ps1
+++ b/Actions/AL-Go-Helper.ps1
@@ -823,11 +823,8 @@ function AnalyzeRepo {
         [string] $project = '.',
         [string] $insiderSasToken,
         [switch] $doNotCheckArtifactSetting,
-        [switch] $doNotCheckAppDependencyProbingPaths,
         [switch] $doNotIssueWarnings,
-        [string[]] $includeOnlyAppIds,
-        [string] $server_url = $ENV:GITHUB_SERVER_URL,
-        [string] $repository = $ENV:GITHUB_REPOSITORY
+        [string[]] $includeOnlyAppIds
     )
 
     $settings = $settings | Copy-HashTable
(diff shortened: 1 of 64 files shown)


jonaswre:
@freddydk this should help partners migrating to al-go

freddydk:
Very nice:-)
What happens if you run the workflow twice?
Or run the workflow on a repo, where there already is code?

freddydk:
How do you handle authentication (if you are migrating a private repo) - I guess we would have to add username and password to the inputs - then that should work, right?

jonaswre:
> Very nice:-) What happens if you run the workflow twice? Or run the workflow on a repo, where there already is code?

Well it will merge both repos but keeps the history of the imported one. I thourgt about a confirmation input (Y/N) defaulting to N explaining the danger. 

Maybe you could see if you are ahead from the al-go template if so don't allow it?

jonaswre:
> How do you handle authentication (if you are migrating a private repo) - I guess we would have to add username and password to the inputs - then that should work, right?

Nope DevOps and Github allow to pass credentials in clone url. {pat}@github.com

freddydk:
> > Very nice:-) What happens if you run the workflow twice? Or run the workflow on a repo, where there already is code?
> 
> Well it will merge both repos but keeps the history of the imported one. I thourgt about a confirmation input (Y/N) defaulting to N explaining the danger.
> 
> Maybe you could see if you are ahead from the al-go template if so don't allow it?

Yeah - I think it is a good idea - will probably add the workflow directly to the templates instead of having it as an action - and then check that the repo is empty and auto-add it to unusedWorkflows by Update AL-Go if the repo isn't empty.

jonaswre:
> > > Very nice:-) What happens if you run the workflow twice? Or run the workflow on a repo, where there already is code?
> > 
> > 
> > Well it will merge both repos but keeps the history of the imported one. I thourgt about a confirmation input (Y/N) defaulting to N explaining the danger.
> > Maybe you could see if you are ahead from the al-go template if so don't allow it?
> 
> Yeah - I think it is a good idea - will probably add the workflow directly to the templates instead of having it as an action - and then check that the repo is empty and auto-add it to unusedWorkflows by Update AL-Go if the repo isn't empty.

And then you need to "private" fork the template and run the workflow from the template.  Otherwise I see potential to leak information about the app you are migrating. 

It might be better to find an indicator for changes since creation from the template.
For example: 
Number of branches == 1
Number of commits == 1
Number of .al files == 0
.github/AL-Go-Settings.json exists

Basicly meens AL-Go Repo without branches/commits or code.
This should keep the user from nuking code/history.

jonaswre:
Or to be totally safe. Create a new Import-AL-Go Template which has only this action. 

And add a new input "Template Repository URL" to this action. 

This would take care of all.

jonaswre:
@freddydk I think this is the best solution. I gave it a try https://github.com/byndit/AL-Go-Import this works great and avoids all the dangers.

freddydk:
could be https://aka.ms/AL-Go-Import
I totally like the idea

jonaswre:
@freddydk the repo can now be deloyed using. The Deploy action. But one important todo would still be to migrate to a powershell action to support the @ notation for referencing branches. And automaticlly registering app/test apps would be cool.

Issue #371, opened 2023-01-19
'There are updates for your AL-Go system' and 'No previous release found' for builds on release branches 


AskeHolst:
When merging PRs to a release/* branch, there is an output of 'There are updates for your AL-Go system'.
Could this be due to the release branch being created when 2.2 was current? The main branch is on 2.3.
Probably this is nothing to worry about, but incorrect warnings are annoying (we try to run a "no warnings" policy).

We also get a 'No previous release found', but there are multiple previous releases. Is this by design due to it being hard to determine the correct previous build in a release branch? Builds on main do find the previous release.

This is not critical, we are back to a partially manual process for hotfixes due to #360 anyway. So this is as much a heads-up about a potential issue.

freddydk:
The release branch should locate the released bits from this release as previous release
will investigate together with 360 today

freddydk:
In the Runpipeline job, could you locate the section with Locating Previous release and send the output of that?
![image](https://user-images.githubusercontent.com/10775043/213861339-166169fb-0090-4f1e-8d3e-32e2ed806bf2.png)

AskeHolst:
The output is short:

```
Locating previous release
  Analyzing releases https://api.github.com/repos/prymeGlobal/[REDACTED]/releases
  Warning: No previous release found
```

I added you as a collaborator in case you need to have a look. Feel free to change settings or run workflows.

freddydk:
I don't think I have your email - could you shoot me an email and explain to me how you created the release branch?
Manually - or through the Create Release workflow?

freddydk:
The release branch was called release/2.3 - I created a release/2.3.0 from the same tag - now it finds the previous release.
I would expect 2.3 to find the same - let me investigate that.
If the branch is created in Create Release - it should have the name of the tag (2.3.0)

freddydk:
Yeah, there definitely is a bug here - not only the branch name.

freddydk:
Have submitted a fix for the known bugs in release branches.
The "There are updates" will not be removed - if your release branch is running on an old version of AL-Go - you should be notified.
You can run the Update AL-Go System files on this.
I will create an item in the backlog for supporting that the Update AL-Go System Files can update all release branches as well.

AskeHolst:
How would you run Update AL-Go System files in a release branch? 
[...]

Issue #360, opened 2023-01-17
Creating a release and deploying from a release branch


AskeHolst:
A few days ago release "2.3" was created in a repo, build 2.3.8.0 was published to AppSource (low build no. as we migrated the repo to GH just before 2.3). Main branch incremented to 2.4.
Development ongoing in 2.4, now at 2.4.14.0

We needed to hotfix 2.3 after release. 
Changes were made on the release/2.3 branch which resulted in build 2.3.16.0.
So far so good.

What would be the way to use the "Publish to AppSource" workflow to pick up 2.3.16.0 and submit it? 
I get errors that it does not exist, I guess because it was build of the release/2.3 branch.

```
Artifacts 2.3.16.0
Projects:
.
Using AppSourceContext
Project '[REDACTED]'
Analyzing artifacts
https://api.github.com/repos/prymeGlobal/[REDACTED]/actions/artifacts?per_page=100&page=1
project ''[REDACTED]'
Error: Deliver action failed. Error: ERROR: Could not find any Apps artifacts for projects *, version 2.3.16.0 Stacktrace: at <ScriptBlock>, D:\a\_actions\microsoft\AL-Go-Actions\v2.3\Deliver\Deliver.ps1: line 169 at <ScriptBlock>, D:\a\_actions\microsoft\AL-Go-Actions\v2.3\Deliver\Deliver.ps1: line 150 at <ScriptBlock>, D:\a\_actions\microsoft\AL-Go-Actions\v2.3\Deliver\Deliver.ps1: line 98 at <ScriptBlock>, D:\a\_temp\0f90fede-a49d-4e64-80b2-47ed0dd49904.ps1: line 2 at <ScriptBlock>, <No file>: line 1
```

Similarly, I cannot get "Create Release" workflow to complete.
```
projects:
- .
Repo Artifacts count: 41
Downloaded Artifacts count: 41
Analyzing artifacts for project [REDACTED]
Search for [REDACTED]-main-Apps-2.3.16.0
Error: No artifacts found for this project
Error: Process completed with exit code 1.
```

Not urgent, I did it manually, but I would be nice to understand how it is supposed to work.

freddydk:
Will investigate

freddydk:
The root cause of this is the same as #371

PR #374, opened 2023-01-23
Release Branches

freddydk:
A Partner filed two bugs on release branches - #360 and #371

I created an end-2-end scenario test, which tested how release branches was intended to work.

This did unfortunately reveal a few bugs in release branches handling, which all are fixed in this PR.
- Release branches was named after the full tag (1.0.0), even though subsequent hotfixes released from this branch would be 1.0.x
- Release branches named 1.0 wasn't picked up as a release branch
- Release notes contained the wrong changelog
- The previous release was always set to be the first release from a release branch
- SemVerStr could not have 5 segments after the dash
- Release was created on the right SHA, but the release branch was created on the wrong SHA

Also added some CI tests for the SemVer conversion functions and added comments to the code for these.

This PR also contains a fix for #376

And it contains updates to documentation on settings:
- CICDPushBranches
- CICDPullRequestBranches
- CICDSchedule
- UpdateGitHubGoSystemFilesSchedule


Actions/Github-Helper.psm1
```diff
@@ -407,6 +468,39 @@ function GetReleases {
...
+    # Default github sorting will return the latest historically created release as the latest release - not the highest version
+    $releases = GetReleases -token $token -api_url $api_url -repository $repository
+
+    # Get Latest release
+    $latestRelease = $releases | Where-Object { -not ($_.prerelease -or $_.draft) } | Select-Object -First 1
+    if ($ref -like 'release/*') {
+        try {
+            # If release branch, get the latest release from that the release branch
+            # This is given by the latest release with the same major.minor as the release branch
+            $semVerObj = SemVerStrToSemVerObj -semVerStr $ref.SubString(8) -allowMajorMinorOnly
```
aholstrup1, line 491: Consider something like `$ref.Replace("release/","")` instead of `$ref.SubString(8)`. I find that a bit more readable personally

Actions/Github-Helper.psm1
```diff
@@ -407,6 +468,39 @@ function GetReleases {
...
+            $latestRelease = $releases | Where-Object {
+                $releaseSemVerObj = SemVerStrToSemVerObj -semVerStr $_.tag_name
+                $semVerObj.Major -eq $releaseSemVerObj.Major -and $semVerObj.Minor -eq $releaseSemVerObj.Minor
+            } | Select-Object -First 1
+        }
+        catch {
+            throw "Release branch name '$ref' is not a valid semantic version string (https://semver.org)"
```
aholstrup1, line 498: Consider removing the try catch block here? We're already throwing this error from the SemVerStrToSemVerObj function

freddydk: Agree - done

Actions/Github-Helper.psm1
```diff
@@ -304,11 +304,26 @@ function invoke-git {
...
+#   Patch: the patch version number
+#   Addt0: the first additional segment (zzz means not specified)
+#   Addt1: the second additional segment (zzz means not specified)
+#   Addt2: the third additional segment (zzz means not specified)
+#   Addt3: the fourth additional segment (zzz means not specified)
+#   Addt4: the fifth additional segment (zzz means not specified)
+#
+# Returns the SemVer string
+#   #   [v]major.minor.patch[-addt0[.addt1[.addt2[.addt3[.addt4]]]]]
 function SemVerObjToSemVerStr {
```
kosioreka, line 322: What's the need of keeping so many properties in versioning? Addt0 etc...

mazhelez (approved):

PR #382, opened 2023-02-02
CICD workflow separation

aholstrup1:
Separate CICD workflow into two workflows:
- _PR Build_, that runs on pull requests
- _Official Build_, that runs on pushes to certain branches (e.g. `main`).


Tests/VerifyPRChanges.Action.Test.ps1
```diff
@@ -0,0 +1,62 @@
+Describe 'VerifyPRChanges Action Tests' {
+
+    BeforeAll {
+        $scriptPath = Join-Path $PSScriptRoot "..\Actions\VerifyPRChanges\VerifyPRChanges.ps1" -Resolve
+    }
+
+    It 'should fail if the PR is from a fork and changes a script' {
+        Mock -CommandName Invoke-WebRequest -MockWith { 
+            '{ "files": [{ "filename": "Actions/AL-Go-Helper.ps1", "status": "modified" }] }'
```
mazhelez, line 9: Tests for _added_ and _deleted_ are missing.

Actions/VerifyPRChanges/VerifyPRChanges.ps1
```diff
@@ -0,0 +1,37 @@
...
+    }
+    $url = "$($githubApiUrl)/repos/$($prBaseRepository)/compare/$baseSHA...$headSHA"
+    $response = Invoke-WebRequest -UseBasicParsing -Headers $headers -Uri $url | ConvertFrom-Json
+    Write-Host "Files Changed:"
+    $response.files | ForEach-Object {
+      $filename = $_.filename
+      Write-Host "- $filename $_.status"
+      $extension = [System.IO.Path]::GetExtension($filename)
+      $name = [System.IO.Path]::GetFileName($filename)
+      if ($extension -eq '.ps1' -or $extension -eq '.yaml' -or $extension -eq '.yml' -or $name -eq "CODEOWNERS") {
```
mazhelez, line 30: There aren't any yet, but what about `.psm1` files?

Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
```diff
@@ -16,74 +16,240 @@ permissions:
   pull-requests: read
 
 jobs:
-  PullRequestHandler:
+  PregateCheck:
```
mazhelez, line 19: Shouldn't this job simply have `if: ${{ github.event.pull_request.base.repo.full_name != github.event.pull_request.head.repo.full_name}}` ?

Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
```diff
@@ -16,74 +16,240 @@ permissions:
...
+
+  Initialization:
+    needs: [ PregateCheck ]
+    runs-on: [ windows-latest ]
+    outputs:
+      telemetryScopeJson: ${{ steps.init.outputs.telemetryScopeJson }}
+      settings: ${{ steps.ReadSettings.outputs.SettingsJson }}
+      projects: ${{ steps.ReadSettings.outputs.ProjectsJson }}
+      projectCount: ${{ steps.ReadSettings.outputs.ProjectCount }}
+      environments: ${{ steps.ReadSettings.outputs.EnvironmentsJson }}
```
mazhelez, line 37: `environments` and `environmentCount` are probably not needed.

mazhelez (commented):
Great first iteration 👍 

Needs a few touches here and there.

Actions/VerifyPRChanges/VerifyPRChanges.ps1
freddydk: We will need to allow .yml or .yaml files in other directories in the repo - and also .ps1 files.
I think we should disallow these files in .github (and subfolders) and in .AL-Go folders (where we automatically will run them)

Example: Our upcoming Test Task Recorder will deliver tests in .yaml format and also the ALDoc tool will produce documentation in .yaml format.

Templates/Per Tenant Extension/.github/workflows/CICD.yaml
```diff
@@ -1,25 +1,11 @@
...
-  workflow_run:
-    workflows: ["Pull Request Handler"]
-    types:
-      - completed
   push:
-    paths-ignore:
-      - '**.md'
-      - '.github/workflows/*.yaml'
-      - '!.github/workflows/CICD.yaml'
     branches: [ 'main', 'release/*', 'feature/*' ]
```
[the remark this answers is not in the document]
freddydk: We should not change the defaults - people can do that in settings.
If people are used to these branches, things will change over an update if we remove a branch.

Actions/VerifyPRChanges/VerifyPRChanges.ps1
mazhelez: Sounds like a separate task :)

Templates/AppSource App/.github/workflows/CICD.yaml
```diff
@@ -51,111 +36,21 @@ jobs:
...
-              }
-              Write-Host "Copying $path to $newFolder"
-              Copy-Item $path -Destination $newFolder -Force
-            }
-          }
-          Remove-Item -Path $prfolder -recurse -force
 
       - name: Initialize the workflow
         id: init
-        uses: microsoft/AL-Go-Actions/WorkflowInitialize@main
```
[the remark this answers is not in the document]
mazhelez: How would it work actually? Because these are new actions. They won't exist in _microsoft/AL-Go-Actions_ until the PR is merged and then deployed.

Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
```diff
@@ -0,0 +1,44 @@
...
+    [Parameter(HelpMessage = "Settings from repository in compressed Json format", Mandatory = $true)]
+    [string] $settingsJson,
+    [Parameter(HelpMessage = "Name of the built project", Mandatory = $true)]
+    [string] $project,
+    [Parameter(HelpMessage = "Buildmode used when building the artifacts", Mandatory = $true)]
+    [string] $buildMode,
+    [Parameter(HelpMessage = "Name of the branch the workflow is running on", Mandatory = $true)]
+    [string] $branchName,
+    [Parameter(HelpMessage = "Setting this switch will write github outputs to environment variables", Mandatory = $false)]
+    [switch] $runLocally
```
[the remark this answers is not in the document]
mazhelez: Another approach would be to set an environment variable `localALGoActions` to `true` whenever we ran actions locally (e.g. in tests). That way we won't have to pollute almost every action with an extra parameter.

Templates/AppSource App/.github/workflows/CICD.yaml
```diff
@@ -1,25 +1,11 @@
...
 
 on:
   workflow_dispatch:
-  workflow_run:
-    workflows: ["Pull Request Handler"]
-    types:
-      - completed
   push:
-    paths-ignore:
-      - '**.md'
```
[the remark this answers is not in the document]
mazhelez: I believe the idea is that we cannot be sure what should and what shouldn't trigger a build. 
So, by default, we do it for every pushed change.

Here's a scenario to consider: The .md files in the projects contain documentation, so on every new build+deliver an official documentation somewhere is updated.

Templates/AppSource App/.github/workflows/CICD.yaml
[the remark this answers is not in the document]
freddydk: correct - when I grab this SHA in my repo and run deploy - the new actions will be deployed and everything is fine.
Everything in microsoft/AL-Go* needs to point to microsoft/AL-Go* - else it will not be picked up by deploy/collect

Templates/AppSource App/.github/workflows/CICD.yaml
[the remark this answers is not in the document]
freddydk: Good point - we should probably create a setting (like CICDPushBranches) called CICDPathsIgnore which is an array, we merge into pathsignore during Update AL-Go System Files and default that to the files that were there.

Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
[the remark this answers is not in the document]
freddydk: GitHub Outputs are (by default) written to the file pointed out by $ENV:GITHUB_OUTPUT.
During tests, shouldn't we just set that environment variable to a temp filename - and then read that file to check whether the intended outcome is written?
This way we wouldn't have to have any code-change in the action to run it under tests?

Actions/VerifyPRChanges/VerifyPRChanges.ps1
freddydk: Yeah OK - we can include that in the PR for those things.

Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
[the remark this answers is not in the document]
aholstrup1: Good idea :) I'll update that

Templates/Per Tenant Extension/.github/workflows/CICD.yaml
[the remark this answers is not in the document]
aholstrup1: This comment isn't really related to the PR so I'll close it. I tend to agree with Maria, that I would probably also expect this to only include main and release

Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
aholstrup1: Good point :) Thank you

Templates/AppSource App/.github/workflows/CICD.yaml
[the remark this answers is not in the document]
aholstrup1: As Maria said, I'm using my own AL-GO-Actions right now because these code changes don't exist in microsoft/AL-GO :) I'll make sure to revert them back before merging the PR. 
Perhaps we should consider checking this in the gates, so we won't accidently end up merging private AL-GO-Actions into the repo

Templates/AppSource App/.github/workflows/CICD.yaml
```diff
@@ -246,59 +141,30 @@ jobs:
...
-            }
-          }
-          while ($step -ge 1) {
-              Add-Content -Path $env:GITHUB_OUTPUT -Value "projects$($step)Json="
-              Add-Content -Path $env:GITHUB_OUTPUT -Value "projects$($step)Count=0"
-              Write-Host "projects$($step)Json="
-              Write-Host "projects$($step)Count=0"
-              $step--
-          }
+        uses: aholstrup1/AL-Go-Actions/DetermineBuildOrder@main
```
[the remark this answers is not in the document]
aholstrup1: Same as https://github.com/microsoft/AL-Go/pull/382#discussion_r1095314511

Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
aholstrup1: Use 'If not failed'

Templates/AppSource App/.github/workflows/CICD.yaml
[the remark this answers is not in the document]
aholstrup1: Reverted to defaults

Actions/VerifyPRChanges/VerifyPRChanges.ps1
```diff
@@ -0,0 +1,39 @@
...
+    [Parameter(HelpMessage = "The URL of the GitHub API", Mandatory = $false)]
+    [string] $githubApiUrl,
+    [Parameter(HelpMessage = "The token to use for the GitHub API", Mandatory = $false)]
+    [string] $githubToken
+
+)
+
+$ErrorActionPreference = "STOP"
+Set-StrictMode -version 2.0
+if ($prHeadRepository -ne $prBaseRepository) {
```
mazhelez, line 19: This check is already covered in the workflow itself.

Templates/AppSource App/.github/workflows/CICD.yaml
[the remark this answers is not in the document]
freddydk: The end-2-end tests will deploy the 3 repos and run the tests on them - meaning that when running the end-2-end tests on your SHA - some will use the actions in your repo.
You need to run the end-2-end tests after all references is microsoft/AL-Go-xxx
I noticed that a few tests are failing with a wrong number of actions run - I guess that is because we are running fewer actions when combining PR handler with build. This probably needs to be fixed in the tests.

Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
```diff
@@ -16,74 +16,203 @@ permissions:
...
+          shell: ${{ needs.Initialization.outputs.githubRunnerShell }}
+          parentTelemetryScopeJson: ${{ needs.Initialization.outputs.telemetryScopeJson }}
+          Project: ${{ matrix.project }}
+          ProjectDependenciesJson: ${{ needs.Initialization.outputs.projectDependenciesJson }}
+          settingsJson: ${{ env.Settings }}
+          SecretsJson: ${{ env.RepoSecrets }}
+          buildMode: ${{ matrix.buildMode }}
+
+      - name: Calculate Artifact names
+        uses: aholstrup1/AL-Go-Actions/CalculateArtifactNames@main
```
freddydk, line 145: Running the end-2-end tests will use aholstrup1/AL-Go-Actions and NOT what you have in the PR here.

Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
```diff
@@ -0,0 +1,41 @@
...
+    [string] $project,
+    [Parameter(HelpMessage = "Buildmode used when building the artifacts", Mandatory = $true)]
+    [string] $buildMode,
+    [Parameter(HelpMessage = "Name of the branch the workflow is running on", Mandatory = $true)]
+    [string] $branchName
+)
+
+function Set-EnvVariable([string] $name, [string] $value) {
+    Write-Host "Assigning $value to $name"
+    Add-Content -Path $env:GITHUB_OUTPUT -Value "$name=$value"
```
freddydk, line 14: I thought that if your action doesn't have any outputs defined in action.yaml, I don't think outputting anything to GITHUB_OUTPUT will have any meaning.
I saw that you defined outputs in CI/CD grabbing values from this action - does that really work?

Actions/TelemetryHelper.psm1
```diff
@@ -14,7 +14,7 @@ $signals = @{
     "DO0082" = "AL-Go action ran: AnalyzeTests"
```
freddydk, line 14: Today all the telemetry functions are used through BcContainerHelper.
We should be able to add to telemetry without containerhelper and then add DO numbers for new actions here,
Will add that as a separate refactoring workitem - we should not hold this PR up for that I think

Actions/ReadSettings/ReadSettings.ps1
```diff
@@ -117,6 +117,75 @@ try {
...
+        }
+
+        return $filesChanged
+    }
+
+    function Get-ProjectsToBuild($settings, $projects, $baseFolder, $token) {
+        if ($settings.alwaysBuildAllProjects) {
+            Write-Host "Building all projects because alwaysBuildAllProjects is set to true"
+            return $projects
+        } elseif ($ENV:GITHUB_WORKFLOW -eq 'CI/CD') {
```
freddydk, line 147: We should have a different way to determine this than the name of the workflow.
CI/CD is the name property in the yaml file and if we at some point in time decide to change that, we will not see this.
IMO - this value should come in a parameter

Actions/VerifyPRChanges/VerifyPRChanges.ps1
```diff
@@ -0,0 +1,34 @@
...
+  [Parameter(HelpMessage = "Base commit of the PR", Mandatory = $false)]
+  [string] $baseSHA,
+  [Parameter(HelpMessage = "Head commit of the PR", Mandatory = $false)]
+  [string] $headSHA,
+  [Parameter(HelpMessage = "The name of the repository the PR is going to", Mandatory = $false)]
+  [string] $prBaseRepository,
+  [Parameter(HelpMessage = "The URL of the GitHub API", Mandatory = $false)]
+  [string] $githubApiUrl,
+  [Parameter(HelpMessage = "The token to use for the GitHub API", Mandatory = $false)]
+  [string] $githubToken
```
freddydk, line 11: In other actions, the token is just called token and placed first in the order.
I think we should use the same name here (unless it is a different token - and I don't think it is)

freddydk (changes requested):

Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
aholstrup1: You're right :( I discovered this yesterday as well. I can define the outputs for CalculateArtifactNames in the action.yaml but for DetermineBuildOrder this might not be possible, since our outputs names are variable e.g. projects1json, projects2json

Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
aholstrup1: All my changes should be deployed to aholstrup1/AL-Go-Actions so it should work, no?

freddydk: Yes - it should work.
and... - I will run all end-2-end tests before release to preview anyway - but we should probably have a test, which ensures that all references to AL-Go-Actions are microsoft (as you mentioned)

Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
freddydk: and here you need to add the outputs to the test as well (see other actions with output)

Actions/ReadSettings/ReadSettings.ps1
aholstrup1: We decided to build all projects unless the workflow is triggered by pull_request or pull_request_target

Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
aholstrup1: Outputs added to CalculateArtifactNames

Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
```diff
@@ -16,74 +16,231 @@ permissions:
...
+        uses: actions/checkout@v3
+        with:
+          lfs: true
+
+      - name: Initialize the workflow
+        id: init
+        uses: aholstrup1/AL-Go-Actions/WorkflowInitialize@main
+        with:
+          shell: powershell
+          eventId: "DO0091"
```
freddydk, line 60: use DO0104 and add that to telemetryhelper as PullRequestHandler

Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
```diff
@@ -16,74 +16,231 @@ permissions:
...
+        uses: actions/checkout@v3
+
+      - name: Finalize the workflow
+        id: PostProcess
+        uses: aholstrup1/AL-Go-Actions/WorkflowPostProcess@main
         with:
-          name: Pull_Request_Files
-          path: '${{ steps.ChangedFiles.outputs.prfolder }}/'
+          shell: powershell
+          eventId: "DO0091"
```
freddydk, line 245: use DO0104

Scenarios/EnablingTelemetry.md
```diff
@@ -34,7 +34,7 @@ Here is a list of the telemetry signals for different workflows:
 | Event ID | Description |
 | :-- | :-- |
 | DO0090 | AL-Go workflow ran: AddExistingAppOrTestApp |
-| DO0091 | AL-Go workflow ran: CiCd |
+| DO0091 | AL-Go workflow ran: CICD |
```
freddydk, line 37: Add DO0104 to telemetryHelper and here (as suggested otherwere) - and add the other missing entries as well:-)

freddydk (changes requested):

freddydk:
I have requested a few more changes, I am OK with merging after this.
After merging - I will still run a number of tests before deploying to preview - and then next week we can deploy to preview,

Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
```diff
@@ -0,0 +1,41 @@
+Param(
+    [Parameter(HelpMessage = "Settings from repository in compressed Json format", Mandatory = $true)]
+    [string] $settingsJson,
+    [Parameter(HelpMessage = "Name of the built project", Mandatory = $true)]
+    [string] $project,
+    [Parameter(HelpMessage = "Buildmode used when building the artifacts", Mandatory = $true)]
```
mazhelez, line 6:
```suggestion
    [Parameter(HelpMessage = "Build mode used when building the artifacts", Mandatory = $true)]
```

Actions/ReadSettings/ReadSettings.ps1
```diff
@@ -19,6 +19,64 @@ Param(
...
+    $filesChanged = @($response.files | ForEach-Object { $_.filename })
+
+    return $filesChanged
+}
+
+function Get-ProjectsToBuild($settings, $projects, $baseFolder, $token) {
+    if ($settings.alwaysBuildAllProjects) {
+        Write-Host "Building all projects because alwaysBuildAllProjects is set to true"
+        return $projects
+    } elseif ($ENV:GITHUB_EVENT_NAME -notin @("pull_request_target", "pull_request")) {
```
mazhelez, line 41: No need for the `elseif` if the `if` returns.

Actions/ReadSettings/ReadSettings.ps1
```diff
@@ -19,6 +19,64 @@ Param(
...
+        Write-Host "Building all projects because this is not a pull request"
+        return $projects
+    }
+    else {
+        $filesChanged = @(Get-ChangedFiles -token $token)
+        if ($filesChanged.Count -eq 0) {
+            Write-Host "Building all projects"
```
mazhelez, line 48: Building all projects, because...? 😕

Actions/ReadSettings/ReadSettings.ps1
```diff
@@ -19,6 +19,64 @@ Param(
...
+            return $projects
+        }
+        elseif ($filesChanged -like '.github/*.json') {
+            Write-Host "Changes to Repo Settings, building all projects"
```
mazhelez, line 52:
```suggestion
            Write-Host "Changes to repo settings, building all projects"
```

Actions/ReadSettings/ReadSettings.ps1
```diff
@@ -19,6 +19,64 @@ Param(
...
+            return $projects
+        }
+        else {
+            Write-Host "Modified files:"
+            $buildProjects = @()
+            $filesChanged | Out-Host
+            $buildProjects = @($projects | Where-Object {
+                $checkProject = $_
+                $buildProject = $false
+                if (Test-Path -path (Join-Path $baseFolder "$checkProject/.AL-Go/settings.json")) {
```
mazhelez, line 66:
```suggestion
                if (Test-Path -Path (Join-Path $baseFolder "$checkProject/.AL-Go/settings.json")) {
```

Actions/ReadSettings/ReadSettings.ps1
```diff
@@ -202,6 +207,8 @@ try {
                 Write-Host "BuildOrderDepth=$($buildOrder.Count)"
             }
         }
+        Write-Host "Projects to build: $($buildProjects -join ', ')"
```
mazhelez, line 210: This is output already in `Get-ProjectsToBuild` so I suggest to remove one of the logging (probably the one in `Get-ProjectsToBuild`).

Actions/ReadSettings/ReadSettings.ps1
```diff
@@ -202,6 +207,8 @@ try {
...
+        Write-Host $buildProjects
```
mazhelez, line 211: Also, is this line of logging needed?

Actions/VerifyPRChanges/VerifyPRChanges.ps1
```diff
@@ -0,0 +1,34 @@
...
+  [Parameter(HelpMessage = "The token to use for the GitHub API", Mandatory = $false)]
+  [string] $token,
+  [Parameter(HelpMessage = "Base commit of the PR", Mandatory = $false)]
+  [string] $baseSHA,
+  [Parameter(HelpMessage = "Head commit of the PR", Mandatory = $false)]
+  [string] $headSHA,
+  [Parameter(HelpMessage = "The name of the repository the PR is going to", Mandatory = $false)]
+  [string] $prBaseRepository,
+  [Parameter(HelpMessage = "The URL of the GitHub API", Mandatory = $false)]
+  [string] $githubApiUrl
```
mazhelez, line 11: So, this parameter is here because it's needed in the in PS and there `github.api_url` is not in the context, right?

I don't particularly like the approach of having parameters to sake of matching YAML to PS. Also, with a bit more complex actions (that's not just calling a PS script), this whole idea collapses. Up to you to decide what's better, but in my opinion, it's not worth polluting the actions so that we can test that they call PS with the same parameters.

Actions/VerifyPRChanges/VerifyPRChanges.ps1
```diff
@@ -0,0 +1,34 @@
...
+}
+$url = "$($githubApiUrl)/repos/$($prBaseRepository)/compare/$baseSHA...$headSHA"
+$response = Invoke-WebRequest -UseBasicParsing -Headers $headers -Uri $url | ConvertFrom-Json
+Write-Host "Files Changed:"
+$response.files | ForEach-Object {
+  $filename = $_.filename
+  Write-Host "- $filename $_.status"
+  $extension = [System.IO.Path]::GetExtension($filename)
+  $name = [System.IO.Path]::GetFileName($filename)
+  if ($extension -eq '.ps1' -or $extension -eq '.yaml' -or $extension -eq '.yml' -or $name -eq "CODEOWNERS") {
```
mazhelez, line 29: What about everything in .github folder? Didn't we agree that's a no-go for PRs?

mazhelez (approved):
Left some non-blocking comments. 

After fixing the telemetry, let's merge this cutie 🚀

Actions/ReadSettings/ReadSettings.ps1
aholstrup1: Now that we only do incremental builds for PRs, this shouldn't be needed

Actions/VerifyPRChanges/VerifyPRChanges.ps1
aholstrup1: I suppose in this case it's such a simple parameter, that we could just put it in the script instead of passing it as a parameter

I agree on your point about the Yaml tests though

PR #392, opened 2023-02-10
Verify origin of AL-GO actions in PR 

aholstrup1:
To make sure we don't accidently add references to forks of AL-GO-Actions, we should have a gate that checks for references to AL-GO-Actions. In this gate we should check if all workflows that are using AL-GO-Actions, are using them from microsoft/AL-GO-Actions

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -20,3 +20,13 @@ jobs:
       - name: Run Tests
         run: |
           . (Join-Path "." "Tests/runtests.ps1")
+  
+  TestALGOActionOrigins:
+    name: Test ALGO Action Origins
+    runs-on: [ ubuntu-latest ]
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Test AL-GO Action Origins
+        run: |
+          . (Join-Path "." "Tests/ActionOrigin.Test.ps1")
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -20,13 +20,8 @@ jobs:
       - name: Run Tests
         run: |
           . (Join-Path "." "Tests/runtests.ps1")
-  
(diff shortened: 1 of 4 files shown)


.github/workflows/CI.yaml
```diff
@@ -20,3 +20,13 @@ jobs:
       - name: Run Tests
         run: |
           . (Join-Path "." "Tests/runtests.ps1")
+  
+  TestALGOActionOrigins:
```
freddydk, line 24: Why not just create another step in the Test job - instead of spinning up another runner, checking out all source code and running this in parallel

freddydk: or actually - since you placed the test in the tests folder and celled it xxx.test.ps1 - it should already be executed by the runtests command above - is this job needed at all`?

aholstrup1: That's true. The idea was to split this from the others since it's not actually testing a specific action. Thus, in some PRs (like https://github.com/microsoft/AL-Go/pull/382) I'm okay with this test failing right up until I want to merge
I see your point though, so I'll remove the second step :) We can always change it later if we feel like it

Tests/TestActionsHelper.psm1
```diff
@@ -140,5 +140,37 @@ function YamlTest {
...
+    }
+}
+
+function TestAllWorkflowsInPath {
+    param(
+        [Parameter(Mandatory)]
+        [string]$Path
+    )
+
+    $workflows = Get-ChildItem -Path $Path -Filter "*.yaml" -Recurse
```
mazhelez, line 168:
```suggestion
    $workflows = Get-ChildItem -Path $Path -Filter "*.yaml" -File -Recurse

```

mazhelez (approved):
Should we also forbid actions comming from microsoft/AL-Go/Actions?

PR #394, opened 2023-02-13
Fix issues with Pull Request Handler

aholstrup1:
This PR is fixing three issues with PR Handler: 
1. Getting the correct head sha 
2. Fix for using the wrong parameter name in action 
3. Stop skipping build task when pregate check is skipped


mazhelez (approved):
Lgtm 😄 

Just check if `&& !failure()` is needed for other jobs.

PR #404, opened 2023-02-16
Enable concurrency groups in Pull Request Handler

aholstrup1:
Enable concurrency groups in Pull Request Handler. This way there's only one PR Build running at a time per PR, and if there's a new push to the PR while a PR Build is running, that PR build will be cancelled.

### DIFF
--- a/Actions/CheckForUpdates/CheckForUpdates.ps1
+++ b/Actions/CheckForUpdates/CheckForUpdates.ps1
@@ -270,7 +270,7 @@ try {
                                     # Example (depth 1):
                                     #    needs: [ Initialization ]
                                     #    if: needs.Initialization.outputs.projects1Count > 0
-                                    $if = "if: (!failure()) && needs.Initialization.outputs.projects$($_)Count > 0"
+                                    $if = "(!failure()) && (!cancelled()) && needs.Initialization.outputs.projectCount > 0"
                                 }
                                 else {
                                     # Subsequent build jobs needs to have a dependency on all previous build jobs
--- a/Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
+++ b/Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
@@ -6,6 +6,10 @@ on:
       - '**.md'
     branches: [ 'main' ]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref }}
+  cancel-in-progress: true
+
 defaults:
   run:
     shell: powershell
@@ -38,7 +42,7 @@ jobs:
 
   Initialization:
     needs: [ PregateCheck ]
-    if: (!failure())
+    if: (!failure() && !cancelled())
     runs-on: [ windows-latest ]
     outputs:
       telemetryScopeJson: ${{ steps.init.outputs.telemetryScopeJson }}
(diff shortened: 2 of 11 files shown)


Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
```diff
@@ -114,7 +118,7 @@ jobs:
 
   Build:
     needs: [ Initialization ]
-    if: (!failure()) && needs.Initialization.outputs.projectCount > 0
```
freddydk, line 117: This fix is also needed in CheckForUpdates.ps1, here:
https://github.com/microsoft/AL-Go/blob/3c2112110a6843827e77458a068579fb18215620/Actions/CheckForUpdates/CheckForUpdates.ps1#L273

and the syntax should be the same as in line 289.
The same syntax should be used above.

If this is not fixed, you will get updates are available immediately after starting

This change should also be done in CICD.yaml, Current.Yaml, NextMinor.yaml and NextMajor.yaml.

freddydk (changes requested):

Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
```diff
@@ -6,6 +6,10 @@ on:
       - '**.md'
     branches: [ 'main' ]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref }}
```
mazhelez, line 10: Is `github.head_ref` unique across forks?

freddydk: concurrency is not active across forks

mazhelez: I meant if `github.head_ref` identifies the head repo.

If I submit a PR from the _main_ branch of my fork, would that cancel the build on your PR, that is from the _main_ branch on your fork?

aholstrup1: Good question.. I'll investigate

aholstrup1: Updated this to use github.event.pull_request.number. Then it shouldn't matter if the branch is from the local repo or from a fork

Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
freddydk: Still missing the current.yaml workflow

freddydk (changes requested):

Actions/CheckForUpdates/CheckForUpdates.ps1
```diff
@@ -270,7 +270,7 @@ try {
                                     # Example (depth 1):
                                     #    needs: [ Initialization ]
                                     #    if: needs.Initialization.outputs.projects1Count > 0
-                                    $if = "if: (!failure()) && needs.Initialization.outputs.projects$($_)Count > 0"
+                                    $if = "(!failure()) && (!cancelled()) && needs.Initialization.outputs.projectCount > 0"
```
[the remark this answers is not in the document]
freddydk: I did change this (wanted to suggest a change, but it just changed it)

freddydk (approved):

PR #411, opened 2023-02-23
Use appFolders and testFolders to determine project dependencies

mazhelez:
When scanning for apps in an AL-Go project, the process should look into `appFolders` and `testFolders` (if defined).

### DIFF
--- a/Actions/AL-Go-Helper.ps1
+++ b/Actions/AL-Go-Helper.ps1
@@ -683,7 +683,7 @@ function AnalyzeRepo {
         }
     }
 
-    Write-Host "Checking appFolders and testFolders"
+    Write-Host "Checking appFolders, testFolders and bcptTestFolders"
     $dependencies = [ordered]@{}
     $appIdFolders = [ordered]@{}
     1..3 | ForEach-Object {
@@ -940,8 +940,6 @@ function AnalyzeRepo {
     }
 
     Write-Host "Checking project dependencies"
-
-
 
     Write-Host "Checking appDependencyProbingPaths"
     if ($settings.appDependencyProbingPaths) {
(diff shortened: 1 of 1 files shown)


Actions/AL-Go-Helper.ps1
```diff
@@ -1785,11 +1785,21 @@ Function AnalyzeProjectDependencies {
     $projects | ForEach-Object {
         $project = $_
         Write-Host "- $project"
-        $apps = @()
-        $folders = @(Get-ChildItem -Path (Join-Path $baseFolder $project) -Recurse | Where-Object { $_.PSIsContainer -and (Test-Path (Join-Path $_.FullName 'app.json')) } | ForEach-Object { $_.FullName.Substring($baseFolder.Length+1) } )
+
+        # Read project settings
```
freddydk, line 1789: You can use:

```
$projectSettings = ReadSettings -baseFolder $baseFolder -project $thisProject
$projectSettings = AnalyzeRepo -settings $projectSettings -baseFolder $baseFolder -project $thisProject -doNotCheckArtifactSetting -doNotIssueWarnings
$folders = @($projectSettings.appFolders+$projectSettings.testFolders+$projectSettings.bcptTestFolders)
```

to get projectSettings (and folders) with appFolders and testFolders calculated for you.

freddydk (changes requested):

Actions/AL-Go-Helper.ps1
```diff
@@ -1785,11 +1785,21 @@ Function AnalyzeProjectDependencies {
...
+            Write-Host "No apps or tests folders found for project $project. Scanning for apps in the project folder."
+            $folders = @(Get-ChildItem -Path (Join-Path $baseFolder $project) -Recurse | Where-Object { $_.PSIsContainer -and (Test-Path (Join-Path $_.FullName 'app.json')) } | ForEach-Object { $_.FullName.Substring($baseFolder.Length+1) } )
+        }
+
+        Write-Host "Folders containing apps are $($folders -join ',' )"
+
         $unknownDependencies = @()
         $apps = @()
-        $sortedFolders = Sort-AppFoldersByDependencies -appFolders $folders -baseFolder $baseFolder -WarningAction SilentlyContinue -unknownDependencies ([ref]$unknownDependencies) -knownApps ([ref]$apps)
+        Sort-AppFoldersByDependencies -appFolders $folders -baseFolder $baseFolder -WarningAction SilentlyContinue -unknownDependencies ([ref]$unknownDependencies) -knownApps ([ref]$apps)
```
freddydk, line 1802: If you do not set a variable, you should add **| Out-Null**

freddydk (changes requested):

Actions/AL-Go-Helper.ps1
bc-external-contributor: Thanks. I will try it out.

mazhelez: With `AnalyzeRepo` _appFolders_, _testFolders_ and _bcptTestFolders_ are calculated relative to the AL-Go project's folder.

However, `Sort-AppFoldersByDependencies` expects them relative to the root folder (`$baseFolder`).

freddydk: You are right - it was more to have a single place to enumerate and find appFolders etc. (which I apparently didn't honor myself in this example).
You could then instead of setting folders directly use:
```
$folders = @($projectSettings.appFolders+$projectSettings.testFolders+$projectSettings.bcptTestFolders | ForEach-Object { Join-Path $project $_ } )
```
If you decide not to go with this approach, then note that $folders in your code can be null - and $folders.count will fail instead of being 0.

freddydk (changes requested):

Actions/AL-Go-Helper.ps1
mazhelez: Should be alright now.

Issue #385, opened 2023-02-04
Some properties in settings files are case sensitive


freddydk:
Most properties in the settings files are not case sensitive, but some are.
AppSourceProductId is - needs to be spelled with capital A.

should be case insensitive.

This is caused by this strange behavior of PowerShell:
![image](https://user-images.githubusercontent.com/10775043/216756612-2c982a6f-5aca-45e5-b118-62afcc409945.png)
When converting an OrderedDictionary to a HashTable - it becomes case sensitive (even though the ordered dictionary is not and the normal hashtable is also not).

PR #419, opened 2023-02-23
Issue #385 Ensure case insensitivity

freddydk:
Due to what I see as an error in PowerShell, ordered dictionaries cast into a HashTable will become a case sensitive HashTable. This PR will copy the potentially wrong HashTable to ensure we have a case insensitive HashTable.


mazhelez:
This issue seems like it could re-appear if we keep using `[ordered]` and `[hashtable]`.

Any general solution ideas?

mazhelez (approved):

freddydk:
> This issue seems like it could re-appear if we keep using `[ordered]` and `[hashtable]`.
> 
> Any general solution ideas?

I have only found this when transferring an ordered direction into a [hashtable] parameter.
Currently we only use Ordered in settings and a few other places to make predictable output.
but no, I don't have a general solution idea.

PR #423, opened 2023-02-27
DetermineProjectsToBuild action

mazhelez:
- Separate calculating AL-Go projects to build and project dependencies into a separate action, _**DetermineProjectsToBuild**_ (please comment on the name).
- Calculate `buildDimensions` to use in the `Build` job. The current dimensions are AL-Go project and build mode. This allows to have build modes per project.
- Fix issue with _thisBuild_ artifacts

Pending:
- [x] Tests (there are some, but need to add more )
- [x] Propagate template changes to AL-Go-AppSource and other workflows
- [x] Address _CalculateArtifactsNames_ usage in _NextMinor_, _NextMajor_ and _Current_ workflow


Templates/Per Tenant Extension/.github/workflows/CICD.yaml
```diff
@@ -28,19 +28,17 @@ jobs:
...
       githubRunnerShell: ${{ steps.ReadSettings.outputs.GitHubRunnerShell }}
       checkRunId: ${{ steps.CreateCheckRun.outputs.checkRunId }}
-      projectDependenciesJson: ${{ steps.ReadSettings.outputs.ProjectDependenciesJson }}
-      buildOrderJson: ${{ steps.ReadSettings.outputs.BuildOrderJson }}
-      buildOrderDepth: ${{ steps.ReadSettings.outputs.BuildOrderDepth }}
-      buildModes: ${{ steps.ReadSettings.outputs.BuildModes }}
+      projects: ${{ steps.LoadProjects.outputs.ProjectsJson }}
+      projectDependenciesJson: ${{ steps.LoadProjects.outputs.ProjectDependenciesJson }}
+      buildOrderJson: ${{ steps.LoadProjects.outputs.BuildOrderJson }}
+      buildOrderDepth: ${{ steps.LoadProjects.outputs.BuildOrderDepth }}
```
mazhelez, line 41: Is this one used?

Templates/Per Tenant Extension/.github/workflows/CICD.yaml
```diff
@@ -275,12 +251,20 @@ jobs:
           project: ${{ matrix.project }}
           buildMode: ${{ matrix.buildMode }}
           branchName: ${{ github.ref_name }}
+          
+      - name: Calculate Sanitized Project name
```
mazhelez, line 255: This is a fix for when an AL-Go project is not in the root folder (so `matrix.project` contains slashes).

Think of a better way to handle this.

Templates/Per Tenant Extension/.github/workflows/CICD.yaml
```diff
@@ -206,15 +183,14 @@ jobs:
 
   Build:
     needs: [ Initialization ]
-    if: (!failure()) && (!cancelled()) && needs.Initialization.outputs.projectCount > 0
+    if: (!failure()) && (!cancelled()) && fromJson(needs.Initialization.outputs.buildOrderJson)[0].projectsCount > 0
```
mazhelez, line 186: Seems a bit convoluted.

Better ideas of how this part should look?

Actions/ReadSettings/action.yaml
```diff
@@ -21,10 +21,6 @@ inputs:
     description: Project folder
     required: false
     default: '.'
-  getProjects:
```
aholstrup1, line 24: This is not so much a comment as it is a question; Do we consider this a breaking change in AL-GO? Generally, is it a breaking change to remove a parameter from an AL-GO-Action?

Templates/Per Tenant Extension/.github/workflows/CICD.yaml
mazhelez: I'll add it to _CalculateArtifactNames_

Templates/Per Tenant Extension/.github/workflows/CICD.yaml
mazhelez: Remove `buildOrderDepth`

Templates/Per Tenant Extension/.github/workflows/CICD.yaml
mazhelez: Added to _CalculateArtifactNames_

Actions/AL-Go-Helper.ps1
```diff
@@ -103,6 +103,22 @@ function ConvertTo-HashTable() {
     $ht
 }
 
+function ConvertTo-JsonArray() {
+    [CmdletBinding()]
+    Param(
+        [Parameter(Mandatory=$true, ValueFromPipeline=$true)] 
+        $array
+    )
+    if ($array.Count -le 1) {
```
mazhelez, line 112: "less or equal" or "equal" here?

Do we expect empty array to be serialized as "[]" or as ""?

freddydk: []
as
'[]' | ConvertFrom-Json returns an empty array
and
'' | ConvertFrom-Json returns null

Actions/CheckForUpdates/CheckForUpdates.ps1
```diff
@@ -243,59 +243,50 @@ try {
...
                                 1..$depth | ForEach-Object {
                                     # All build job needs to have a dependency on the Initialization job
                                     $needs = @('Initialization')
                                     if ($_ -eq 1) {
                                         # First build job needs to have a dependency on the Initialization job only
                                         # Example (depth 1):
                                         #    needs: [ Initialization ]
-                                        #    if: needs.Initialization.outputs.projects1Count > 0
-                                        $if = "if: (!failure()) && (!cancelled()) && needs.Initialization.outputs.projects$($_)Count > 0"
+                                        #    if: fromJson(needs.Initialization.outputs.buildOrderJson)[0].projectsCount > 0
```
mazhelez, line 266:
```suggestion
                                        #    if: (!failure()) && (!cancelled()) && fromJson(needs.Initialization.outputs.buildOrderJson)[0].projectsCount > 0
```

Actions/CheckForUpdates/CheckForUpdates.ps1
```diff
@@ -243,59 +243,50 @@ try {
...
+                                        $if = "if: (!failure()) && (!cancelled()) && fromJson(needs.Initialization.outputs.buildOrderJson)[$($_ - 1)].projectsCount > 0"
                                     }
                                     else {
                                         # Subsequent build jobs needs to have a dependency on all previous build jobs
-                                        # Example (depth 2):
+                                        # Example (depth 3):
```
mazhelez, line 271:
```suggestion
                                        # Example (depth 2):
```

Templates/Per Tenant Extension/.github/workflows/CICD.yaml
```diff
@@ -60,8 +58,26 @@ jobs:
...
-          getProjects: 'Y'
           getEnvironments: '*'
+          
+      - name: Load Build Projects
+        id: loadProjects
+        uses: microsoft/AL-Go-Actions/LoadProjects@main
+        with:
+          shell: powershell
+
+      - name: Verify workflow depth
```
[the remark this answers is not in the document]
mazhelez: 😓

I'll try 😄

Actions/AL-Go-Helper.ps1
mazhelez: I removed `ConvertTo-JsonArray`.

It turned out `ConvertTo-Json $array` does its jobs properly. `$array | ConvertTo-Json` unwraps the array, so there might be side-effects if there is a single value in the array.

Templates/Per Tenant Extension/.github/workflows/CICD.yaml
mazhelez: No better alternative for now.

Templates/Per Tenant Extension/.github/workflows/Current.yaml
```diff
@@ -152,27 +147,11 @@ jobs:
...
-        if: success() || failure()
-        run: |
-          $ErrorActionPreference = "STOP"
-          Set-StrictMode -version 2.0
-          $settings = '${{ env.Settings }}' | ConvertFrom-Json
-          $project = '${{ matrix.project }}'
-          if ($project -eq ".") { $project = $settings.repoName }
-          'TestResults','BcptTestResults','BuildOutput','ContainerEventLog' | ForEach-Object {
-            $name = "$($_)ArtifactsName"
-            $value = "$($project.Replace('\','_').Replace('/','_'))-$_-Current-$([DateTime]::UtcNow.ToString('yyyyMMdd'))"
```
mazhelez, line 171: `CalculateArtifactsNames` action does not calculate the names like that. Need to be altered.

Templates/Per Tenant Extension/.github/workflows/CICD.yaml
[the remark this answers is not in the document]
mazhelez: I tried to think of a way, but an action that just compares two numbers seems a bit of an overkill.

Anyway, it can be addressed in a follow-up PR if you feel strongly about it. I'll resolve this comment now.

Actions/DetermineProjectsToBuild/action.yaml
```diff
@@ -0,0 +1,35 @@
...
+    description: An array of objects that determine that build order, including build dimensions
+    value: ${{ steps.determineProjectsToBuild.outputs.BuildOrderJson }}
+runs:
+  using: composite
+  steps:
+    - name: Determine Projects to Build
+      shell: ${{ inputs.shell }}
+      id: determineProjectsToBuild
+      run: try { 
+          $changedFiles = ${{ github.action_path }}/GetModifiedFiles.ps1 -token ${{ github.token }}
```
aholstrup1, line 29: Did you verify this works? I actually thought the `github.token` was inaccessible to GitHub Actions. From the [docs](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context): 

_Note: This context property is set by the Actions runner, and is only available within the execution steps of a job. Otherwise, the value of this property will be null._

Might be okay since you're accessing it from the execution steps of a GitHub Action?

mazhelez: I'm double-checking...

mazhelez: Tested in BuildOrder project (see [here](https://github.com/mazhelez/BuildOrder/pull/2)).

Churn-based build in the PR Handler works as expected.

aholstrup1 (approved):

mazhelez:
Maybe some release notes?

Actions/CheckForUpdates/CheckForUpdates.ps1
```diff
@@ -243,59 +243,53 @@ try {
                         }
                     }
 
-                    # CICD, Current, NextMinor and NextMajor workflows all include a build step.
+                    # PullRequestHandler, CICD, Current, NextMinor and NextMajor workflows all include a build step.
                     # If the dependency depth is higher than 1, we need to add multiple dependent build jobs to the workflow
                     if ($baseName -eq 'PullRequestHandler' -or $baseName -eq 'CICD' -or $baseName -eq 'Current' -or $baseName -eq 'NextMinor' -or $baseName -eq 'NextMajor') {
                         $yaml.Replace('env:/workflowDepth:',"workflowDepth: $depth")
```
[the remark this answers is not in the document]
freddydk: This is the first run after upgrading an old repo with UseProjectDependencies: `https://github.com/BusinessCentralApps/buildorder/actions/runs/4348419232`
The workflow passes, but nothing gets built.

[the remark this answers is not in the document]
freddydk: Re-running update will give:
`https://github.com/BusinessCentralApps/buildorder/actions/runs/4348570838`
Now - there are no updates and it looks like things are building...

freddydk:
Two issues I have encountered.
The Pull Request Build after updating AL-Go System Files doesn't build anything (when multiproject)
https://github.com/BusinessCentralApps/buildorder/actions/runs/4348567214

When single-project - it does.
https://github.com/BusinessCentralApps/BingMaps.PTE/actions/runs/4351096290

Since the PR Build runs on target - then nothing actually changed, so that is probably correct - maybe we should detect this in pregate check or maybe when detecting that no projects are to be build in initialize, we could add a build summary step explaining that no projects are changed - skipping all builds.

On the PR build above on single project, no BuildOutput artifacts was created.
Testing whether BuildOutput artifacts are created by a Multi-project PR build here:
https://github.com/BusinessCentralApps/buildorder/actions/runs/4351407874

mazhelez:
> The Pull Request Build after updating AL-Go System Files doesn't build anything (when multiproject) https://github.com/BusinessCentralApps/buildorder/actions/runs/4348567214

Can I see the PR itself? Is it [this one](https://github.com/BusinessCentralApps/buildorder/pull/54/files)
I am asking because the PR adds new action in the _Initialization_ job, _Determine Projects To Build_.
image

But in this case, no projects were picked up because the of the modified files:
> 5 modified file(s): .github/workflows/CICD.yaml, .github/workflows/Current.yaml, .github/workflows/NextMajor.yaml, .github/workflows/NextMinor.yaml, .github/workflows/PullRequestHandler.yaml

So, this was the "follow-up" PR.

> When single-project - it does.
https://github.com/BusinessCentralApps/BingMaps.PTE/actions/runs/4351096290

Based on the time, I assume the PR is [this one](https://github.com/BusinessCentralApps/BingMaps.PTE/pull/212/files)
The PR changes the template and the worfklows, but the workflows will run from the _main_ branch. Hence, the result.

It is interesting, because the somehow it detects changes in the workflow but it doesn't run them:
image

freddydk:
> Based on the time, I assume the PR is [this one](https://github.com/BusinessCentralApps/BingMaps.PTE/pull/212/files)

Yes, you can also see the PR build # and the PR # in the list:
image

> But in this case, no projects were picked up because the of the modified files:

But wouldn't you expect that if CICD.yaml (when running CI/CD), PullRequestHandler.yaml (when running PR Build) and/or AL-Go-Settings.json was changed - we should re-build all?

mazhelez:
> But wouldn't you expect that if CICD.yaml (when running CI/CD), PullRequestHandler.yaml (when running PR Build) and/or AL-Go-Settings.json was changed - we should re-build all?

I would, but in this case only workflow files were modified:
> 5 modified file(s): .github/workflows/CICD.yaml, .github/workflows/Current.yaml, .github/workflows/NextMajor.yaml, .github/workflows/NextMinor.yaml, .github/workflows/PullRequestHandler.yaml
(I assume that's because even if we run the build, it will run without the changes anyway.)

A project is built only if there are modified files in its .AL-Go folder or any of the app folders.
All projects are built is there are modified files, matching `.github/*.json`.
There are other cases but none of the criteria is met, so technically there are no projects to build.

freddydk:
> > But wouldn't you expect that if CICD.yaml (when running CI/CD), PullRequestHandler.yaml (when running PR Build) and/or AL-Go-Settings.json was changed - we should re-build all?
> 
> I would, but in this case only workflow files were modified:
> 
> > 5 modified file(s): .github/workflows/CICD.yaml, .github/workflows/Current.yaml, .github/workflows/NextMajor.yaml, .github/workflows/NextMinor.yaml, .github/workflows/PullRequestHandler.yaml
> > (I assume that's because even if we run the build, it will run without the changes anyway.)
> 
> A project is built only if there are modified files in its .AL-Go folder or any of the app folders. All projects are built is there are modified files, matching `.github/*.json`. There are other cases but none of the criteria is met, so technically there are no projects to build.

Got it - but shouldn't we add a criteria to build all if cicd.yaml and/or pullrequesthandler.yaml was changed?

mazhelez:
> Got it - but shouldn't we add a criteria to build all if cicd.yaml and/or pullrequesthandler.yaml was changed?

We can, but even if these files are changed, the build would pick up the old ones in the case of a PR (target branch). So not much value here.
In any other case (_CICD_, _Current_, etc.) all projects are built anyway.

If we ever change the trigger on "Pull Request Build" to be `pull_request` (not `pull_request_target`) then a build-all is a must, if pullrequesthandler.yaml is modified in a PR.

What should we settle on?

RELEASENOTES.md
```diff
@@ -2,13 +2,24 @@
...
 In v2.4 and earlier, the PullRequestHandler would trigger the CI/CD workflow to run the PR build.
 Now, the PullRequestHandler will perform the build and the CI/CD workflow is only run on push (or manual dispatch)
 
+### Build modes per project
+Build modes can now be specified per project
+
+### New Actions
+- **DetermineProjectsToBuild** is used to determine which projects to build in PullRequestHandler, CI/CD, Current, NextMinor and NextMajor workflows.
+- **CalculateArtifactNames** is used to calculate artifact names in PullRequestHandler, CI/CD, Current, NextMinor and NextMajor workflows.
+- **VerifyPRChanegs** is used to verify whether a PR contains changes, which are not allowed form a fork.
```
mazhelez, line 21:
```suggestion
- **VerifyPRChanges** is used to verify whether a PR contains changes, which are not allowed from a fork.
```

RELEASENOTES.md
```diff
@@ -2,13 +2,24 @@
...
 
+### **NOTE:** When upgrading to this version
+When upgrading to this version form earlier versions of AL-Go for GitHub, you will need to run the Update AL-Go System Files workflow twice if you have the UseProjectDependencies set to true.
+
 ### Issues
 - Issue [#391](https://github.com/microsoft/AL-Go/issues/391) Create release action - CreateReleaseBranch error
 
 ### Changes to Pull Request Process
 In v2.4 and earlier, the PullRequestHandler would trigger the CI/CD workflow to run the PR build.
 Now, the PullRequestHandler will perform the build and the CI/CD workflow is only run on push (or manual dispatch)
```
mazhelez, line 13:
```suggestion
Now, the PullRequestHandler will perform the build and the CI/CD workflow is only run on push (or manual dispatch) and will perform a complete build.
```

RELEASENOTES.md
```diff
@@ -2,13 +2,24 @@
 
 Note that when using the preview version of AL-Go for GitHub, you need to Update your AL-Go system files, as soon as possible when told to do so.
 
+### **NOTE:** When upgrading to this version
+When upgrading to this version form earlier versions of AL-Go for GitHub, you will need to run the Update AL-Go System Files workflow twice if you have the UseProjectDependencies set to true.
```
mazhelez, line 6:
```suggestion
When upgrading to this version form earlier versions of AL-Go for GitHub, you will need to run the _Update AL-Go System Files_ workflow twice if you have the `useProjectDependencies` setting set to _true_.
```

PR #426, opened 2023-02-28
Allow GetBaseFolder from any depth  

aholstrup1:
Currently, GetBaseFolder throws an error if the AL-GO project is not in the root of the repository. This change will allow localDevEnv to work, even when that is not the case.


Actions/AL-Go-Helper.ps1
```diff
@@ -1914,17 +1914,17 @@ function GetBaseFolder {
     Param(
         [string] $folder
     )
+    
+    $CurrentDir = Get-Location
```
mazhelez, line 1918: You can use the following pattern here to save one variable and be make it more resilient to potential errors.

```
Push-Location
try {
    doSomething
}
finally {
    Pop-Location
}
```

mazhelez (approved):

freddydk:
Only reason for not doing this would be if we are in doubt whether people have git installed when running this on their laptop and I guess that when they have a clone of the repo, we can assume that they have.

Actions/AL-Go-Helper.ps1
```diff
@@ -1914,17 +1914,20 @@ function GetBaseFolder {
     Param(
         [string] $folder
     )
+    
+    Push-Location $folder
+    try {
+        $baseFolder = invoke-git rev-parse --show-toplevel -returnValue
```
freddydk, line 1920: You can add -silent to avoid that the invoke-git prints out the result on the host.
I am OK with either.

freddydk (approved):

Issue #427, opened 2023-02-28
External dependencies in multi project repos that use useProjectDependencies


jonaswre:
I've had to switch a single project/multi app repo to a multi project/single app repo to be able to deploy to app source. 

I've now encountered an issue where builds triggered from a PR wont install/download dependencies which come from another repo (using nuget). 

In both runs, external dependecies are correctly identified. 

But in the run from the PR this line is never executed. 

https://github.com/microsoft/navcontainerhelper/blob/81f0eba49b606545316fd560a42a3f0f2da16d91/AppHandling/Run-AlPipeline.ps1#L1051.

The pipeline skips this section and then fails the compilation with a 404 when trying to download the symbols. 

![image](https://user-images.githubusercontent.com/3731681/221972140-af147cbf-ff26-4525-a35c-3c23f1f3d096.png)
The failed build

![image](https://user-images.githubusercontent.com/3731681/221972212-b40e0a51-48c7-42ff-b8bb-d05d70d2f402.png)
The succeeded build

And the changes in the PR are only a few al files nothing major.

jonaswre:
@freddydk 

And I can't revert to main...

Downloading BcContainerHelper preview version from Blob Storage
BcContainerHelper version 4.0.14-preview902
BC.HelperFunctions emits usage statistics telemetry to Microsoft
Using template from https://github.com/microsoft/AL-Go-AppSource@main
Using ArchiveUrl https://api.github.com/repos/microsoft/AL-Go-AppSource/zipball/main
Analyzing projects
- BeyondCatalog
Folders containing apps are .\BeyondCatalog\app,.\BeyondCatalog\test
- BeyondDataImporterCatalogBridge
Folders containing apps are .\BeyondDataImporterCatalogBridge\app
- BeyondRecordSearchCatalogBridge
Folders containing apps are .\BeyondRecordSearchCatalogBridge\app
Analyzing dependencies
- BeyondCatalog
Set ProjectDependencies for BeyondCatalog to 
No additional dependencies found, add project to build order
- BeyondDataImporterCatalogBridge
Set ProjectDependencies for BeyondDataImporterCatalogBridge to BeyondCatalog
Found dependencies to projects: BeyondCatalog
- BeyondRecordSearchCatalogBridge
Set ProjectDependencies for BeyondRecordSearchCatalogBridge to BeyondCatalog
Found dependencies to projects: BeyondCatalog
#1 - build projects: BeyondCatalog
[...]

PR #428, opened 2023-03-01
Issue #427 - Dependencies to GitHub Packages are not included in PR handler

freddydk:
1. Update AL Go System files cannot downgrade due to missing Outputs under Initialization in old version.
2. PR handler needs access to GitHubPackagesContext in order to resolve dependencies

### DIFF
--- a/Actions/CheckForUpdates/CheckForUpdates.ps1
+++ b/Actions/CheckForUpdates/CheckForUpdates.ps1
@@ -250,14 +250,16 @@ try {
                         if ($depth -gt 1) {
                             # When there are multiple build jobs, we need to add build job specific project list (and count) to the output of the Initialization job
                             $initializationOutputs = $yaml.Get('jobs:/Initialization:/outputs:/')
-                            $addOutput = @()
-                            1..$depth | ForEach-Object {
-                                $addOutput += @(
-                                  "projects$($_): `${{ steps.BuildOrder.outputs.projects$($_)Json }}"
-                                  "projects$($_)Count: `${{ steps.BuildOrder.outputs.projects$($_)Count }}"
-                                )
+                            if ($initializationOutputs) {
+                                $addOutput = @()
+                                1..$depth | ForEach-Object {
+                                    $addOutput += @(
+                                      "projects$($_): `${{ steps.BuildOrder.outputs.projects$($_)Json }}"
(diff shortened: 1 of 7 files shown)


Actions/CheckForUpdates/CheckForUpdates.ps1
```diff
@@ -250,14 +250,16 @@ try {
                         if ($depth -gt 1) {
                             # When there are multiple build jobs, we need to add build job specific project list (and count) to the output of the Initialization job
                             $initializationOutputs = $yaml.Get('jobs:/Initialization:/outputs:/')
-                            $addOutput = @()
-                            1..$depth | ForEach-Object {
-                                $addOutput += @(
-                                  "projects$($_): `${{ steps.BuildOrder.outputs.projects$($_)Json }}"
-                                  "projects$($_)Count: `${{ steps.BuildOrder.outputs.projects$($_)Count }}"
-                                )
+                            if ($initializationOutputs) {
```
mazhelez, line 253: If there is no output, shouldn't the process create it?

mazhelez (approved):

Actions/CheckForUpdates/CheckForUpdates.ps1
freddydk: No, this is when downgrading - the old PR handler doesn't have the Outputs section.

freddydk:
But... - there might be another problem
I have access to a partner repo, which uses GitHubPackages for dependency resolution, but in PR's we don't deliver to NuGet/GitHubPackages - and we shouldn't. We shouldn't deliver PR artifacts like that.
Will check this fix with their repo first.

Actions/CheckForUpdates/CheckForUpdates.ps1
freddydk: But the same problem exists with the build part, downgrading a repo now to test this

freddydk:
> But... - there might be another problem I have access to a partner repo, which uses GitHubPackages for dependency resolution, but in PR's we don't deliver to NuGet/GitHubPackages - and we shouldn't. We shouldn't deliver PR artifacts like that. Will check this fix with their repo first.

No, there is no other known problem in this. For apps from the same repo we use the thisbuild artifacts - not github - this fix solved the issue from the partner.

aholstrup1 (approved):

PR #430, opened 2023-03-02
Apps without subfolders return false warnings

aholstrup1:
When an app folder doesn't have subfolder, we return a warning

### DIFF
--- a/Actions/AL-Go-Helper.ps1
+++ b/Actions/AL-Go-Helper.ps1
@@ -733,7 +733,7 @@ function AnalyzeRepo {
 
             # Check if there are any folders matching $folder
             # Test-Path $folder -PathType Container will return false if any files matches $folder (beside folders)
-            if (-not ((Test-Path $folder) -and (Get-ChildItem $folder -Directory))) {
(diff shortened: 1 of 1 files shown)


Actions/AL-Go-Helper.ps1
```diff
@@ -733,7 +733,7 @@ function AnalyzeRepo {
 
             # Check if there are any folders matching $folder
             # Test-Path $folder -PathType Container will return false if any files matches $folder (beside folders)
-            if (-not ((Test-Path $folder) -and (Get-ChildItem $folder -Directory))) {
+            if (-not ((Test-Path $folder) -and (Get-ChildItem $folder))) {
```
freddydk, line 736: This code checks whether there is a folder matching $folder (which can have wildcards)
Removing -Directory, it will return true if there is a file matching that pattern as well.
I do recognize that the code fails if there is no subfolder AND folder doesn't have wildcards,
Maybe a better fix is:

if (Get-Item $folder | Where-Object { $_ -is [System.IO.DirectoryInfo] })

then you can also remove line 735

freddydk (changes requested):

PR #431, opened 2023-03-07
Add "Publish artifacts - container event log" and "Publish artifacts - build output" to PullRequest Build workflow

mazhelez:
Both _Publish artifacts - build output_ and _Publish artifacts - container event log_ steps should be part of the _PullRequest Build_ workflow as it contains a build job.
Most probably they were omitted when CICD workflow was split.


freddydk:
LGTM

PR #452, opened 2023-03-24
Run PR Build based on merge commit 

aholstrup1:
Sometimes it can be an issue that the PR Build is not running with the latest changes from the base branch. This build https://github.com/microsoft/ALAppExtensions/actions/runs/4500507606 is complaining about breaking changes even though, the PR doesn't touch the SignedXml codeunit. The issue here is that the head branch doesn't contain all the latest changes from the base branch. 

A fix for this could be to use the merge branch created by GitHub since it contains the changes from both head and base branch.

### DIFF
--- a/Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
+++ b/Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
@@ -32,7 +32,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           lfs: true
-          ref: ${{ github.event.pull_request.head.sha }}
+          ref: refs/pull/${{ github.event.number }}/merge
 
       - uses: microsoft/AL-Go-Actions/VerifyPRChanges@main
         with:
@@ -57,7 +57,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           lfs: true
-          ref: ${{ github.event.pull_request.head.sha }}
+          ref: refs/pull/${{ github.event.number }}/merge
 
       - name: Initialize the workflow
         id: init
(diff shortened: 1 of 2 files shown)


freddydk:
Does this always exist when creating a PR?

mazhelez (approved):
Nice one!

aholstrup1:
The ref should always exist, unless the PR is created with a merge conflict. Once the merge conflict is resolved, the ref will be created. 

Once the ref is created it is updated on every push into the head branch, unless there are merge conflicts between the head and base branch. Once the merge conflicts are resolved it will again contain the latest changes from the head branch and the base branch. 

If there are merge conflicts, the default behavior of the pull_request_target trigger is to trigger a build anyway. Once/if we switch to the pull_request trigger, a workflow won't be submitted until the merge conflict is resolved (https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request).

freddydk:
> The ref should always exist, unless the PR is created with a merge conflict. Once the merge conflict is resolved, the ref will be created.
> 
> Once the ref is created it is updated on every push into the head branch, unless there are merge conflicts between the head and base branch. Once the merge conflicts are resolved it will again contain the latest changes from the head branch and the base branch.
> 
> If there are merge conflicts, the default behavior of the pull_request_target trigger is to trigger a build anyway. Once/if we switch to the pull_request trigger, a workflow won't be submitted until the merge conflict is resolved (https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request).

So, what happens right now (with PR_target) if there is a PR conflict?
What is the experience?

freddydk (approved):

PR #454, opened 2023-03-27
Fix licenseFileUrl = "none" param in localDevEnv 

akoniecki:
Fix for localDevEnv.ps1, enabling the script execution with parameter licenseFileUrl = "none".


freddydk:
Is this a fix for when you actually type "none" in the licenseFileUrl field?
I think the idea is that you just press enter if you do not have a license file - that should just work I think?

akoniecki:
@freddydk, exactly! But this a problem when you want to run the script unattended. The parameter value of "none" was previously there but was implemented incorrectly and it was impossible to make use of it.

freddydk:
got it, thanks

freddydk (approved):

PR #456, opened 2023-03-30
Create AL-GO Action for signing 

aholstrup1:
This PR creates an AL-GO Action for signing files using a certificate stored in Azure Key Vault. The signing action also works in the case where the private key is stored in an HSM (Hardware Security Module)

### 1 - Service Principal
a. Create a service principal in Azure 
b. Create a client secret for the service principal 
c. Note down the **Client ID** and **Client Secret** of the service principal as well as the **Tenant ID**

### 2 - Azure Key Vault Setup 
a. Set up a key vault in Azure. To leverage HSMs, you must create a _Premium Tier_ keyvault (https://learn.microsoft.com/en-us/azure/key-vault/general/overview)
b. Give your service principal the following rights: 
Key: Sign
Certificate: Get
c. Note down the **Azure Key Vault URI**

### 3 - Generating / Importing your certificate 
a. Follow the guides on how to generate a certificate into Azure Key Vault (https://learn.microsoft.com/en-us/azure/key-vault/certificates/create-certificate) or importing a certificate into the keyvault (https://learn.microsoft.com/en-us/azure/key-vault/keys/hsm-protected-keys). Note: When setting up your certificate, you'll need to select:
-> Advanced Policy Configuration
 -> Exportable Private Key: No
 -> Key Type: Choose an HSM key type
Note down the **Certificate Name**


aholstrup1:
TODOs: 

- [ ] Move PR description into documentation

Actions/Sign/Sign.ps1
```diff
@@ -0,0 +1,61 @@
...
+$bcContainerHelperPath = $null
+
+# IMPORTANT: No code that can fail should be outside the try/catch
+
+try {
+    . (Join-Path -Path $PSScriptRoot -ChildPath "..\AL-Go-Helper.ps1" -Resolve)
+
+    $BcContainerHelperPath = DownloadAndImportBcContainerHelper -baseFolder $ENV:GITHUB_WORKSPACE 
+
+    Register-NavSip
```
mazhelez, line 36: This looks like it fits better into another step of the action.

Actions/Sign/Sign.ps1
```diff
@@ -0,0 +1,61 @@
...
+
+    AzureSignTool sign --file-digest $FileDigest `
+        --azure-key-vault-url $AzureKeyVaultURI `
+        --azure-key-vault-client-id $AzureKeyVaultClientID `
+        --azure-key-vault-tenant-id $AzureKeyVaultTenantID `
+        --azure-key-vault-client-secret $AzureKeyVaultClientSecret `
+        --azure-key-vault-certificate $AzureKeyVaultCertificateName `
+        --timestamp-rfc3161 "$TimestampService" `
+        --timestamp-digest $TimestampDigest `
+        --verbose `
```
mazhelez, line 52: How chatty is _AzureSignTool_ with the verbose flag on? Link to test logs?

Actions/Sign/Sign.ps1
```diff
@@ -0,0 +1,61 @@
...
+        $Files
+}
+catch {
+    OutputError -message "AnalyzeTests action failed.$([environment]::Newline)Error: $($_.Exception.Message)$([environment]::Newline)Stacktrace: $($_.scriptStackTrace)"
```
mazhelez, line 56: AnalyzeTests? Copy-paste much? 😄 

But that begs the question... any tests around here? Or it's not worth it...

Actions/Sign/action.yaml
```diff
@@ -0,0 +1,77 @@
+name: PowerShell script
```
mazhelez, line 1: PowerShell script? Isn't that supposed to be the name of the action? 🤔

Actions/Sign/action.yaml
```diff
@@ -0,0 +1,77 @@
...
+  steps:
+    - name: Install Azure SignTool
+      shell: ${{ inputs.shell }}
+      run: dotnet tool install --global AzureSignTool --version 4.0.1
+
+    - name: run
+      shell: ${{ inputs.shell }}
+      id: Sign
+      env:
+        _azureKeyVaultURI: ${{ inputs.azureKeyVaultURI }}
```
mazhelez, line 49: This isn't really needed.

From the docs:

> When you specify an input in a workflow file or use a default input value, GitHub creates an environment variable for the input with the name `INPUT_<VARIABLE_NAME>`.

 https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#inputs
 
 Clarification: I'm not suggesting you use `INPUT_<VARIABLE_NAME>`, I suggest you use, for example, `${{ inputs.azureKeyVaultURI }}` instead of `_azureKeyVaultURI`

mazhelez:
> TODOs:
> 
> * [ ]  Move PR description into documentation

You can try use [tasks list](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/about-task-lists) on your PR.

Actions/Sign/Sign.ps1
aholstrup1: No it's probably not necessary. It was nice for debugging but in production it probably won't be needed

Actions/Sign/Sign.ps1
aholstrup1: Not sure I see a need for this. I personally prefer to have the code in Powershell files rather than embedded in yaml

mazhelez: Then why is installing the signing tool a separate task, embedded in the yaml? 😉
It's a one-liner as well: `dotnet tool install --global AzureSignTool --version 4.0.1`

My point is not to have it directly in the yaml, but to achieve good visual separation in the logs (GitHub Actions do that pretty well when there are multiple steps):
-- Install signing tool
-- Register NAV SIP
-- Sign files

Actions/AL-Go-Helper.ps1
```diff
@@ -1952,3 +1952,44 @@ function GetProject {
     }
     $project
 }
+
+function Get-NavSipFromArtifacts() {
```
mazhelez, line 1956: Generally, let's try to avoid adding functions in here as such helper files tend to grow very quickly.

In this case, do you think that this functionality (`Get-NavSipFromArtifacts` or `Register-NavSip`) would be reused in other places, other than signing? If not, fear not to simply move these functions in scripts, specific for the action itself.

mazhelez (approved):

Actions/AL-Go-Helper.ps1
```diff
@@ -1952,3 +1952,44 @@ function GetProject {
...
+    
+    return Join-Path $navSipTempFolder "navsip.dll" -Resolve
+}
+
+function Register-NavSip() {
+    $navsipPath = Get-NavSipFromArtifacts
+    $navSip32Path = "C:\Windows\System32"
+    $navSip64Path = "C:\Windows\SysWow64"
+
+    $navSip32Path, $navSip64Path | ForEach-Object {
```
freddydk, line 1981: I don't think the SysWow64 is needed here - I don't think we don't ship the NavSip in 32 bit anymore

freddydk (dismissed):

Actions/AL-Go-Helper.ps1
```diff
@@ -1952,3 +1952,44 @@ function GetProject {
...
     $project
 }
+
+function Get-NavSipFromArtifacts() {
+    #TODO: It would be nice with a different approach here - This downloads a lot of unnecessary stuff
+    $artifactTempFolder = Join-Path $([System.IO.Path]::GetTempPath()) ([System.IO.Path]::GetRandomFileName())
+    $navSipTempFolder = Join-Path $([System.IO.Path]::GetTempPath()) ([System.IO.Path]::GetRandomFileName())
+
+    try {
+        Download-Artifacts -artifactUrl (Get-BCArtifactUrl -type Sandbox) -includePlatform -basePath $artifactTempFolder | Out-Null
```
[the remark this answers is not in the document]
aholstrup1: I agree. I discussed this with @mazhelez as well. I think we should put the NavSip on a more lightweight artifact along with the compiler and the platform.

Actions/Sign/action.yaml
```diff
@@ -0,0 +1,62 @@
...
+                                      -AzureKeyVaultClientSecret "${{ inputs.azureKeyVaultClientSecret }}" `
+                                      -AzureKeyVaultTenantID "${{ inputs.azureKeyVaultTenantID }}" `
+                                      -AzureKeyVaultCertificateName "${{ inputs.azureKeyVaultCertificateName }}" `
+                                      -TimestampService "${{ inputs.timestampService }}" `
+                                      -TimestampDigest "${{ inputs.digestAlgorithm }}" `
+                                      -FileDigest "${{ inputs.digestAlgorithm }}" `
+                                      -PathToFiles "${{ inputs.pathToFiles }}" `
+                                      -ParentTelemetryScopeJson "${{ inputs.parentTelemetryScopeJson }}"
+        } catch { 
+          Write-Host "::Error::Unexpected error when running action ($($_.Exception.Message.Replace("`r",'').Replace("`n",' ')))"; exit 1
```
mazhelez, line 57: The Sign.ps1 itself has a try-catch, would this `catch` actually catch anything?

aholstrup1: I suppose it would catch errors in the invocation of the script. It also seems to be a convention we have in AL-GO Actions :)

Actions/Sign/Sign.psm1
```diff
@@ -0,0 +1,41 @@
+function Get-NavSipFromArtifacts
+(
+    [string] $NavSipDestination
+) 
+{
+    #TODO: It would be nice with a different approach here - This downloads a lot of unnecessary stuff
```
mazhelez, line 6: Are we going to look for TODOs in the code once in a while? I think not... better create a work item to track this new approach.

mazhelez (approved):

Actions/Sign/Sign.psm1
aholstrup1: Making a note of it :)

PR #468, opened 2023-04-09
Add support for building apps without containers

freddydk:
This PR contains:
- New Action: DetermineArtifactUrl for determining which artifacts to use for a build (seperated out from AnalyzeRepo)
- New Setting UseCompilerFolder to specify that RunPipeline should use the CompilerFolder functionality from ContainerHelper instead of Containers (this setting is temporary and at some point in the future, this will be the default)
- All build workflows will call DetermineArtifactUrl before calling RunPipeline, to determine the artifacts up front.
- CI/CD and PullRequest will additionally cache the necessary pieces of the artifacts in GitHub cache
- Using UseCompilerFolder might (still) make builds slower. If you need to publish apps and run tests, it still needs a container and as such, nothing much saved. (will be solved by a future PR).
- ALAppExtensions and BusinessCentralApps can already today take advantage of UseCompilerFolder for speed optimization.

Missing pieces for this PR.
- [x] Run Complete Test Suite Containerless
- [x] Add Tests for new Action
- [x] Documentation

Next step on performance is to split Signing and Testing from Building the app.

### DIFF
--- a/e2eTests/Test-AL-Go.ps1
+++ b/e2eTests/Test-AL-Go.ps1
@@ -104,13 +104,11 @@ $runs = 0
 SetTokenAndRepository -github:$github -githubOwner $githubOwner -token $token -repository $repository
 
 # Create repo
-CreateAlGoRepository -github:$github -template $template -branch $branch -private:$private -linux:$linux
-$repoPath = (Get-Location).Path
-
 # Set DoNotPublishApps to true until we have test apps and set useCompilerFolder
 # This causes the CI/CD workflow to use FilesOnly containers or CompilerFolder (if UseCompilerFolder is true)
-Pull -branch $branch
-Add-PropertiesToJsonFile -commit -path ".github\AL-Go-Settings.json" -properties @{ "doNotPublishApps" = $true; "useCompilerFolder" = $useCompilerFolder.IsPresent }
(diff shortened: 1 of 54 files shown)


e2eTests/Test-AL-Go.ps1
```diff
@@ -121,15 +125,23 @@ if ($appSourceApp) {
...
+# Merge and run CI/CD + Tests
+MergePRandPull -branch $branch | Out-Null
 $runs++
 
-# Run CI/CD and wait
+# Wait for CI/CD to finish
 $run = Run-CICD -wait -branch $branch
 $runs++
 
+if ($useCompilerFolder) {
```
aholstrup1, line 137: This is because navcontainerhelper sets _doNotRunTests = $true_ when _doNotPublishApps = $true_? Could you add a short comment on that?

freddydk: done

aholstrup1 (approved):

jonaswre:
Awesome work! I really wanted this feature to exist, thanks. So many new possibilities!

freddydk:
> Awesome work! I really wanted this feature to exist, thanks. So many new possibilities!

There's more to come - this is just the first step on a performance enhancement journey:-)

PR #482, opened 2023-04-20
Update Dependencies1.md

buzzwick:
It is not clear whether "we will fix that" means "we are going to fix that together, later in this particular workshop," or "we, the maintainers of AL-Go, are going to fix that in a later release of AL-Go."


freddydk (approved):

freddydk:
Thanks David

Issue #446, opened 2023-03-20
Wrong newline character in release notes


cegekaJG:
When creating a new release, the description seems to use `%0A` instead of a line return.
I am using the latest version of the AL-Go actions, <https://github.com/microsoft/AL-Go/commit/7455e37005e9430e8e8cb5281e4ba757ccddff3f>.

```
## What's Changed%0A* 'New Test App (CI/CD Test App.Test)' by @github-actions in https://github.com/cegekaJG/cicd-test/pull/1%0A* Update workflows by @cegekaJG in https://github.com/cegekaJG/cicd-test/pull/3%0A* 'New Performance Test App (CI/CD Test App.PerformanceTest)' by @github-actions in https://github.com/cegekaJG/cicd-test/pull/4%0A* Reorganize project by @cegekaJG in https://github.com/cegekaJG/cicd-test/pull/6%0A* Update workspace by @cegekaJG in https://github.com/cegekaJG/cicd-test/pull/5%0A%0A## New Contributors%0A* @github-actions made their first contribution in https://github.com/cegekaJG/cicd-test/pull/1%0A* @cegekaJG made their first contribution in https://github.com/cegekaJG/cicd-test/pull/3%0A%0A**Full Changelog**: https://github.com/cegekaJG/cicd-test/commits/1.0.0
```

CarlOhlsson:
Been having issues with the release notes for some time now. See below how they look for me, might be related to this new line character?

![image](https://user-images.githubusercontent.com/11760980/231991532-b5417d87-2962-42d3-aa9c-f2defa05dd50.png)

freddydk:
This was fixed in v3.1

PR #484, opened 2023-04-24
#446: Wrong new line in Create Release

pzentner:
Updated creating release notes with wrong new line characters. From what I can see that /n and /r should not be replaced as they work to create line breaks via API.

When entering the release notes manually using markdown, two space characters will create the line break, but that seems to be limited to only the actual markdown editor and not the API or "creating from files".


freddydk:
I think this might fail in this line:

Add-Content -Path $env:GITHUB_OUTPUT -Value "releaseNotes=$releaseNotes"

Not sure the github_output supports line breaks.

freddydk:
I have created this fix instead: https://github.com/microsoft/AL-Go/pull/485

PR #495, opened 2023-04-28
Integrate Sign action into CICD

aholstrup1:
**This PR integrates the Sign action in AL-GO into the CICD workflow.** 

To uptake this new feature in an AL-GO project you need to supply the following setting: 
* keyVaultCodesignCertificateName

Additionally, you need to make sure to have AZURE_CREDENTIALS set up


Templates/AppSource App/.github/workflows/CICD.yaml
```diff
@@ -244,6 +244,20 @@ jobs:
...
+        id: sign
+        uses: microsoft/AL-Go-Actions/Sign@main
+        with:
+          shell: ${{ needs.Initialization.outputs.githubRunnerShell }}
+          azureKeyVaultURI: ${{ secrets[format('{0}', env.codesignKeyvaultUriSecretName)] }}
+          azureKeyVaultClientID: ${{ secrets[format('{0}', env.codeSignClientIdSecretName)] }}
+          azureKeyVaultClientSecret: ${{ secrets[format('{0}', env.codeSignClientSecretSecretName)] }}
+          azureKeyVaultTenantID: ${{ secrets[format('{0}', env.codeSignTenantIdSecretName)] }}
+          azureKeyVaultCertificateName: ${{ secrets[format('{0}', env.codeSignCertificateSecretName)] }}
+          pathToFiles: '${{ matrix.project }}/.buildartifacts/Apps/*.app'
```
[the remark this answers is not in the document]
freddydk: Test apps are not signed, no

Actions/AL-Go-Helper.ps1
```diff
@@ -455,6 +455,11 @@ function ReadSettings {
         "keyVaultClientIdSecretName"             = ""
         "codeSignCertificateUrlSecretName"       = "codeSignCertificateUrl"
         "codeSignCertificatePasswordSecretName"  = "codeSignCertificatePassword"
+        "codeSignKeyvaultUriSecretName"          = ""
```
freddydk, line 458: I think we should have one secret called azureCodeSignContext, which should be a compressed json structure with these fields (without the codeSign part)

freddydk (changes requested):

Actions/AL-Go-Helper.ps1
aholstrup1: I was trying to avoid doing that as GitHub specifically discourages the use of [structured data in secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets). What is the downside of having the values as separate secrets?

freddydk: I know, and the primary reason for that is that only the entire secret is safe - meaning that if you add a space to the structure or like, the secret might be exposed.
But... - we actually add every property to the masked value and is not hurt by that.
Reason for the structure is that everything else in AL-Go is using this and it is much easier to maintain one secret than 6 - even though the one secret is a structure.
Let's discuss pros and cons on Wednesday if needed.

Actions/AL-Go-Helper.ps1
```diff
@@ -464,6 +464,7 @@ function ReadSettings {
         "keyVaultCertificateUrlSecretName"       = ""
         "keyVaultCertificatePasswordSecretName"  = ""
         "keyVaultClientIdSecretName"             = ""
+        "keyVaultCodesignCertificateName"        = ""
```
mazhelez, line 467: Doc on the new setting is needed.

Actions/Sign/Sign.ps1
```diff
@@ -31,33 +31,32 @@ $bcContainerHelperPath = $null
...
-        --azure-key-vault-url $AzureKeyVaultURI `
-        --azure-key-vault-client-id $AzureKeyVaultClientID `
-        --azure-key-vault-tenant-id $AzureKeyVaultTenantID `
-        --azure-key-vault-client-secret $AzureKeyVaultClientSecret `
-        --azure-key-vault-certificate $AzureKeyVaultCertificateName `
-        --timestamp-rfc3161 "$TimestampService" `
-        --timestamp-digest $TimestampDigest `
-        $Files
+    Retry-Command -Command { 
+        Register-NavSip
```
mazhelez, line 49: Can you not keep the grouping for better logs?

Templates/AppSource App/.github/workflows/CICD.yaml
```diff
@@ -244,6 +244,20 @@ jobs:
...
+        if: env.doNotSignApps == 'False' && env.keyVaultCodesignCertificateName != ''
+        id: sign
+        uses: microsoft/AL-Go-Actions/Sign@main
+        with:
+          shell: ${{ needs.Initialization.outputs.githubRunnerShell }}
+          azureKeyVaultURI: ${{ secrets.AZURE_KEYVAULT_URI }}
+          azureKeyVaultClientID: ${{ secrets.AZURE_CLIENT_ID }}
+          azureKeyVaultClientSecret: ${{ secrets.AZURE_CLIENT_SECRET }}
+          azureKeyVaultTenantID: ${{ secrets.AZURE_TENANT_ID }}
+          azureKeyVaultCertificateName: ${{ env.keyVaultCodesignCertificateName }}
```
mazhelez, line 258: Why not fetch it from `Read settings` step's output?

At some point we'll need to clean up the env vars and keep only what makes sense to be there. `keyVaultCodesignCertificateName` isn't a good candidate for an env variable.

mazhelez (changes requested):
Documentation needs to be clear on uptaking the new action.

How to origanize the secrets, example set-ups, maybe a mini-migration workflow (is it even possible to add a secret with a workflow?), anything that would help uptaking the new action.

Actions/Sign/Sign.ps1
```diff
@@ -31,33 +31,33 @@ $bcContainerHelperPath = $null
...
+            --azure-key-vault-url $AzureKeyVaultURI `
+            --azure-key-vault-client-id $AzureKeyVaultClientID `
+            --azure-key-vault-tenant-id $AzureKeyVaultTenantID `
+            --azure-key-vault-client-secret $AzureKeyVaultClientSecret `
+            --azure-key-vault-certificate $AzureKeyVaultCertificateName `
+            --timestamp-rfc3161 "$TimestampService" `
+            --timestamp-digest $TimestampDigest `
+            $Files
+    } -MaxRetries 3
+    $file = $_
```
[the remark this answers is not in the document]
mazhelez: Outdated.

Templates/AppSource App/.github/workflows/CICD.yaml
aholstrup1: You mean the [SettingsJson](https://github.com/microsoft/AL-Go/blob/a83fa796d47309a8a8c586f9e18b42fe66cd2e6b/Actions/ReadSettings/action.yaml#LL41C3-L41C3)? Or add an output to the ReadSettings Action called keyVaultCodesignCertificateName? 

I was trying to avoid the first one, as I think it's kinda unnecessary to pass a json string with all settings just to get one setting

mazhelez: Technically, you can do something like `azureKeyVaultCertificateName: ${{ fromJSON(steps.readsettings.SettingsJson).keyVaultCodesignCertificateName }}`, but I agree it isn't pretty.

Separate output for every setting might make the _ReadSettings_ action a bit messy, but isn't it better than having all these environment variables floating around? I fear it'll bite us when we least expect it 😄
Anywho, not a target for your PR. Up to you how you want to pass the cert name.

aholstrup1: Let's address this in a separate PR :) I suspect we'll have a lot of cleanup to do if we decide to remove settings as env variables

Scenarios/CodesigningInAlGo.md
```diff
@@ -0,0 +1,73 @@
...
+    Key: Sign
+    Certificate: Get
+```
+
+### 3 - Generating / Importing your certificate 
+To create or import a certificate into Azure Keyvault, you can follow the provided guides:
+
+ Create a certificate in Azure Keyvault:
+
+* [Create a certificate in Azure Key Vault](https://learn.microsoft.com/en-us/azure/key-vault/certificates/create-certificate): This guide explains how to create a certificate directly within Azure Keyvault. It covers the process of generating a self-signed certificate or using a certificate signing request (CSR) to create a certificate signed by a certificate authority (CA).
```
mazhelez, line 42: Omit the locale-language combo. It's figured out on the go. Technically, this should be a forward link as well, but as we don't have it set up, let it be as it is.

Also, I think the product is called "Key Vault", not "Keyvault". If you don't use the product name, then it should be simply "key vault".

```suggestion
* [Create a certificate in Azure Key Vault](https://learn.microsoft.com/azure/key-vault/certificates/create-certificate): This guide explains how to create a certificate directly within Azure Keyvault. It covers the process of generating a self-signed certificate or using a certificate signing request (CSR) to create a certificate signed by a certificate authority (CA).
```

mazhelez (approved):
Left a comment on the documentation.

The rest looks OK.

Actions/ReadSecrets/ReadSecretsHelper.psm1
```diff
@@ -1,13 +1,27 @@
...
-$script:isKeyvaultSet = $script:gitHubSecrets.PSObject.Properties.Name -eq "AZURE_CREDENTIALS"
+$script:isKeyvaultSet = ($script:gitHubSecrets.PSObject.Properties.Name -eq "AZURE_CREDENTIALS") -or ($script:gitHubSecrets.PSObject.Properties.Name -eq "AZURE_KEYVAULT_URI")
 $script:escchars = @(' ','!','\"','#','$','%','\u0026','\u0027','(',')','*','+',',','-','.','/','0','1','2','3','4','5','6','7','8','9',':',';','\u003c','=','\u003e','?','@','A','B','C','D','E','F','G','H','I','J','K','L','M','N','O','P','Q','R','S','T','U','V','W','X','Y','Z','[','\\',']','^','_',[char]96,'a','b','c','d','e','f','g','h','i','j','k','l','m','n','o','p','q','r','s','t','u','v','w','x','y','z','{','|','}','~')
 
 function IsKeyVaultSet {
     return $script:isKeyvaultSet
 }
 
+function Get-KeyVaultName {
+    if ($script:gitHubSecrets.PSObject.Properties.Name -eq "AZURE_KEYVAULT_URI") {
```
freddydk, line 12: If people are using a KeyVault for their secrets already - and uptake this feature - then the existing keyvault will stop working as you will always return the name of the keyvault containing the certificate - right?

Actions/ReadSecrets/ReadSecretsHelper.psm1
```diff
@@ -170,9 +184,13 @@ function GetKeyVaultSecret {
     if (-not $script:keyvaultConnectionExists) {
             
         InstallKeyVaultModuleIfNeeded
-            
-        $credentialsJson = Get-KeyVaultCredentials
-        ConnectAzureKeyVaultIfNeeded -subscriptionId $credentialsJson.subscriptionId -tenantId $credentialsJson.tenantId -clientId $credentialsJson.clientId -clientSecret $credentialsJson.clientSecret
+
+        if ($script:gitHubSecrets.PSObject.Properties.Name -eq "AZURE_KEYVAULT_URI") {
```
freddydk, line 188: Same here - the AZURE_CREDENTIALS will stop working if people create a secret called AZURE_KEYVAULT_URI - or do I misread the code?

Actions/AL-Go-Helper.ps1
```diff
@@ -2026,3 +2027,35 @@ function Determine-ArtifactUrl {
...
+        [int]$MaxRetries,
+        [Parameter(Mandatory = $false)]
+        [int]$RetryDelaySeconds = 5
+    )
+
+    $retryCount = 0
+    while ($retryCount -lt $MaxRetries) {
+        try {
+            & $Command
+            if ($LASTEXITCODE -ne 0) {
```
freddydk, line 2045: isn't LASTEXITCODE only for shell commands - is that the right error check?

Actions/AL-Go-Helper.ps1
```diff
@@ -2026,3 +2027,35 @@ function Determine-ArtifactUrl {
...
+        [Parameter(Mandatory = $true)]
+        [int]$MaxRetries,
+        [Parameter(Mandatory = $false)]
+        [int]$RetryDelaySeconds = 5
+    )
+
+    $retryCount = 0
+    while ($retryCount -lt $MaxRetries) {
+        try {
+            & $Command
```
freddydk, line 2044: I typically use Invoke-command to invoke scriptblocks - are you sure that & is the right mechanism? (I thought that was for commands only, but looking at the code, it looke like it works)

freddydk (changes requested):

Actions/ReadSecrets/ReadSecretsHelper.psm1
aholstrup1: Yes - In this setup we only support one KeyVault. So, people would have to keep their certificates in the same KeyVaults as their other secrets.

Actions/ReadSecrets/ReadSecretsHelper.psm1
aholstrup1: That is correct - If they've created the other secrets as well (AZURE_CLIENT_ID, AZURE_CLIENT_SECRET, AZURE_KEYVAULT_SUBSCRIPTION_ID, AZURE_TENANT_ID) they shouldn't need AZURE_CREDENTIALS anymore. 

If we're worried about people having a AZURE_KEYVAULT_URI secret in their repo already (for another purpose) we can consider adding "ALGO_" as a prefix to the secrets?

Actions/AL-Go-Helper.ps1
aholstrup1: I think both works. Taking your suggestion and using Invoke-Command

Actions/ReadSecrets/ReadSecretsHelper.psm1
freddydk: I am worried that if people will setup signing - then everything breaks until they have moved everything over.
If you have 100 repositories and create the signing secrets as org. secrets - then all repos might break depending on setup.

freddydk: It seems like we are adding 5 new secrets to replace 1 existing secret - and once you create them - the old one doesn't work any more. If you only support one keyvault - there should only be one way to setup the connection IMO.

aholstrup1: In that case of someone owning 100 Al-GO repositories I would assume the AZURE_CREDENTIALS is already set up as an organizational secret. In that case, they'd only need to remove that one secret and add the 5 new ones. Nothing should break in that scenario. 

It's not that the once you add the 5 secrets, the old mechanism doesn't work anymore. It shouldn't be needed anymore and should be removed. There would be no reason to keep both authentication mechanisms since they accomplish the same goal.

freddydk: If they are using the same keyvault for all repos - yes - but that isn't a requirement.
As I understand it - once you add the 5 secrets - AZURE_CREDENTIALS stops working.
Lets talk tomorrow at the sync meeting once more and land this.

Actions/AL-Go-Helper.ps1
aholstrup1: It's both for [powershell scripts and native programs](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_automatic_variables?view=powershell-5.1). In this case, AzureSignTool is an executable that exits with an exitcode. That's how I capture errors in signing

Templates/AppSource App/.github/workflows/CICD.yaml
```diff
@@ -245,6 +245,17 @@ jobs:
...
           secretsJson: ${{ env.RepoSecrets }}
           buildMode: ${{ matrix.buildMode }}
+      
+      - name: Sign
+        if: env.doNotSignApps == 'False' && env.keyVaultCodesignCertificateName != ''
+        id: sign
+        uses: microsoft/AL-Go-Actions/Sign@main
+        with:
+          shell: ${{ needs.Initialization.outputs.githubRunnerShell }}
+          azureCredentialsJson: ${{ secrets.AZURE_CREDENTIALS }}
```
mazhelez, line 255: Isn't it better to pass AZURE_CREDENTIALS as a secret that action can read, than pass it as a parameter?

freddydk (approved):
LGTM

Templates/AppSource App/.github/workflows/CICD.yaml
mazhelez: Yeah, I guess it's more or less the same.

Issue #497, opened 2023-04-29
UpdateSystemFiles fails


jonaswre:
I have a schedule that updates system files once a week. 

The latest updates failed with the following error.

Error: CheckForUpdates action failed. Error: The property 'CurrentSchedule' cannot be found on this object. Verify that the property exists. Stacktrace: at <ScriptBlock>, D:\a\_actions\microsoft\AL-Go-Actions\preview\CheckForUpdates\CheckForUpdates.ps1: line 181 at <ScriptBlock>, D:\a\_actions\microsoft\AL-Go-Actions\preview\CheckForUpdates\CheckForUpdates.ps1: line 162 at <ScriptBlock>, D:\a\_actions\microsoft\AL-Go-Actions\preview\CheckForUpdates\CheckForUpdates.ps1: line 153 at <ScriptBlock>, D:\a\_temp\42b94f56-91a7-4a74-9704-f17c81a78c4d.ps1: line 2 at <ScriptBlock>, <No file>: line 1

My repo settings contain a currentSchedule entry that used to work. 

Maybe a capitalization error was introduced again?

freddydk:
You are right - another one of these.
I will fix asap and then add a todo to see if I can add a test, which ensures that this doesn't happen again.

freddydk:
Shipped in v3.1

PR #498, opened 2023-04-30
UpdateSystemFiles fails in preview

freddydk:
Fix for issue #497

After the latest change in Update AL-Go System Files to use ReadSettings instead of just reading the config file, the repoSettings now became a OrderedDictionary, which is case sensitive.

This PR changes the ConvertTo-HashTable to also be able to recursively convert a HashTable or an OrderedDictrionary to a new HashTable - and then use that in AL-Go System Files.

I will add a todo to see if we can make a test to test for case sensitivity in settings files.

This change shouldn't have any side effects, but running End-2-End tests now to ensure that I am right.

### DIFF
--- a/Actions/AL-Go-Helper.ps1
+++ b/Actions/AL-Go-Helper.ps1
@@ -88,13 +88,24 @@ function ConvertTo-HashTable() {
     [CmdletBinding()]
     Param(
         [parameter(ValueFromPipeline)]
-        [PSCustomObject] $object,
+        $object,
         [switch] $recurse
     )
+
     $ht = @{}
-    if ($object) {
-        $object.PSObject.Properties | ForEach-Object { 
-            if ($recurse -and ($_.Value -is [PSCustomObject])) {
+    if ($object -is [System.Collections.Specialized.OrderedDictionary] -or $object -is [hashtable]) {
+        $object.Keys | ForEach-Object {
+            if ($recurse -and ($object."$_" -is [System.Collections.Specialized.OrderedDictionary] -or $object."$_" -is [hashtable] -or $object."$_" -is [PSCustomObject])) {
+                $ht[$_] = ConvertTo-HashTable $object."$_"
+            }
+            else {
+                $ht[$_] = $object."$_"
+            }
+        }
+    }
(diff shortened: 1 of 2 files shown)


Actions/AL-Go-Helper.ps1
```diff
@@ -88,13 +88,24 @@ function ConvertTo-HashTable() {
...
     )
+
     $ht = @{}
-    if ($object) {
-        $object.PSObject.Properties | ForEach-Object { 
-            if ($recurse -and ($_.Value -is [PSCustomObject])) {
+    if ($object -is [System.Collections.Specialized.OrderedDictionary] -or $object -is [hashtable]) {
+        $object.Keys | ForEach-Object {
+            if ($recurse -and ($object."$_" -is [System.Collections.Specialized.OrderedDictionary] -or $object."$_" -is [hashtable] -or $object."$_" -is [PSCustomObject])) {
+                $ht[$_] = ConvertTo-HashTable $object."$_"
```
mazhelez, line 99: Probably better to pass the `recurse` switch in the recursion as well.

mazhelez (approved):

Actions/AL-Go-Helper.ps1
```diff
@@ -88,13 +88,24 @@ function ConvertTo-HashTable() {
...
+                $ht[$_] = ConvertTo-HashTable $object."$_" -recurse
+            }
+            else {
+                $ht[$_] = $object."$_"
+            }
+        }
+    }
+    elseif ($object -is [PSCustomObject]) {
+        $object.PSObject.Properties | ForEach-Object {
+            if ($recurse -and ($object."$_" -is [System.Collections.Specialized.OrderedDictionary] -or $object."$_" -is [hashtable] -or $_.Value -is [PSCustomObject])) {
```
mazhelez, line 108: Then a `recurse` is needed in the call to `ConvertTo-HashTable` down here as well.

freddydk: You are absolutely right - it wasn't there before, but any caller of this function would expect it to be there...
I have added it - and am re-running tests.

Issue #491, opened 2023-04-27
Unable to get dependency releases during CI/CD pipeline


cegekaJG:
I've added a dependency to a different repository to my app using `.AL-Go/settings.json`, but whenever the action `Run Pipeline` tries to download the artifacts from that depository, the server responds with a 404 - usually seen when trying to access a repository with no or an invalid token.
Here is my `settings.json`:
```json
{
  "country": "at",
  "appFolders": [],
  "testFolders": [],
  "bcptTestFolders": [],
  "appDependencyProbingPaths": [
    {
      "repo": "https://github.com/CBS-BC-AT/CGK.RESTManager",
      "version": "latest",
      "release_status": "release",
      "authTokenSecret": "GHTOKENWORKFLOW"
    }
  ],
  "repoVersion": "2.3"
}
```

And here is the log from the workflow, starting with the step `Read Settings`, up to the error. From what I can tell, the workflow was able to retrieve `GHTOKENWORKFLOW`, but the request failed anyway. To make sure the token had the necessary permissions, I used `cUrl` with the same token and the same URL, `https://api.github.com/repos/CBS-BC-AT/CGK.RESTManager/releases`, and it was successful - I can only assume between reading the secrets and passing the token, the one I specified gets lost.

```
Run microsoft/AL-Go-Actions/ReadSettings@v[2](https://github.com/CBS-BC-AT/CGK.Mindstorms/actions/runs/4818709766/jobs/8581038091#step:8:2).4
  with:
    shell: powershell
[...]

PR #515, opened 2023-05-08
Issue #491 - Unable to get dependency releases during CI/CD pipeline

freddydk:
This is a fix for issue #491

The reason for this is really that the target repository was renamed, but GitHub handles that in all other cases by redirecting you.
This fix will locate the artifacts needed in the new repository and issue a warning that the repository was renamed or moved.
This PR also relaxes the fields that get masked from StorageContext, as these strings might be reused elsewhere.

### DIFF
--- a/Actions/Github-Helper.psm1
+++ b/Actions/Github-Helper.psm1
@@ -76,7 +76,7 @@ function InvokeWebRequest {
             $response = $_.Exception.Response
             $responseUri = $response.ResponseUri.AbsoluteUri
             if ($response.StatusCode -eq 404 -and $responseUri -ne $uri) {
-                Write-Host "::Warning::Repository was moved, using this URL instead: $responseUri"
+                Write-Host "::Warning::Repository ($uri) was renamed or moved, please update your references with the new name. Trying $responseUri, as suggested by GitHub."
                 $result = Invoke-WebRequest @params -Uri $responseUri
             }
             else {
--- a/Actions/ReadSecrets/ReadSecrets.ps1
+++ b/Actions/ReadSecrets/ReadSecrets.ps1
@@ -79,7 +79,7 @@ try {
                     }
                     $json.Keys | ForEach-Object {
                         if (@("Scopes","TenantId","BlobName","ContainerName","StorageAccountName") -notcontains $_) {
-                            # Mask individual values (but not Scopes and TenantId)
+                            # Mask individual values (but not Scopes, TenantId, BlobName, ContainerName and StorageAccountName)
                             MaskValue -key "$($secret).$($_)" -value $json."$_"
                         }
                     }
--- a/Actions/ReadSecrets/ReadSecretsHelper.psm1
+++ b/Actions/ReadSecrets/ReadSecretsHelper.psm1
@@ -58,9 +58,6 @@ function GetGithubSecret {
 }
 	
 function Get-KeyVaultCredentials {
-    Param(
-        [switch] $dontmask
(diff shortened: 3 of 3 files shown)


Actions/ReadSecrets/ReadSecrets.ps1
```diff
@@ -78,7 +78,7 @@ try {
                         throw "JSON Secret $secret contains line breaks. JSON Secrets should be compressed JSON (i.e. NOT contain any line breaks)."
                     }
                     $json.Keys | ForEach-Object {
-                        if ($_ -ne 'Scopes' -and $_ -ne 'TenantId') {
+                        if (@("Scopes","TenantId","BlobName","ContainerName","StorageAccountName") -notcontains $_) {
                             # Mask individual values (but not Scopes and TenantId)
```
mazhelez, line 82: The comment can be more accurate with the new change

Actions/ReadSecrets/ReadSecretsHelper.psm1
```diff
@@ -62,17 +62,17 @@ function Get-KeyVaultCredentials {
         [switch] $dontmask
     )
     if ($script:isKeyvaultSet) {
+        $jsonStr = $script:gitHuBSecrets.AZURE_CREDENTIALS
+        if ($jsonStr -contains "`n" -or $jsonStr -contains "`r") {
+            throw "Secret AZURE_CREDENTIALS cannot contain line breaks"
```
mazhelez, line 67: Why not simply ```$jsonStr -replace "`n","" -replace "`r",""```?

I suppose you tried something like it already, just it's always worth asking.

Actions/Github-Helper.psm1
```diff
@@ -69,7 +69,21 @@ function InvokeWebRequest {
...
         }
-        Invoke-WebRequest  @params -Uri $uri
+        try {
+            $result = Invoke-WebRequest  @params -Uri $uri
+        }
+        catch [System.Net.WebException] {
+            $response = $_.Exception.Response
+            $responseUri = $response.ResponseUri.AbsoluteUri
+            if ($response.StatusCode -eq 404 -and $responseUri -ne $uri) {
+                Write-Host "::Warning::Repository ($uri) was renamed or moved, please update your references with the new name. Trying $responseUri, as suggested by GitHub."
```
aholstrup1, line 79: I'm fine with the warning but did you consider if we should throw an error with this message instead? I just fear that the warning will be overlooked and the builds will keep targeting the old repo name

Actions/ReadSecrets/ReadSecretsHelper.psm1
freddydk: Having linebreaks in the secrets have caused the output to show parts of the secret.
This construct:

${{ secrets.something }}

Is pre-compiled and with newlines it becomes invalidated yaml and an error with exposed partial secrets.

It might be that we are handling these things better today and are not exposed to that - but I don't see any reason for the structured secrets to include newlines - they should be compressed JSON.

Actions/ReadSecrets/ReadSecrets.ps1
freddydk: fixed

Actions/Github-Helper.psm1
freddydk: I think the warning will allow people to rename a repository without breaking all other repos.
Making an error would cause all other apps to fail, which I think is unnecessary.

PR #522, opened 2023-05-10
Add run-name for Deploy workflow

mazhelez:
Improves troubleshooting.

### DIFF
--- a/.github/workflows/Deploy.yaml
+++ b/.github/workflows/Deploy.yaml
@@ -1,5 +1,7 @@
 name: Deploy
 
+run-name: "Deploy from [${{ github.ref_name }}] to [${{ inputs.branch }}]"
+
 on:
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


.github/workflows/Deploy.yaml
```diff
@@ -1,5 +1,7 @@
 name: Deploy
 
+run-name: "Deploy from [${{ github.ref_name }}] to [${{ inputs.branch }}]"
```
freddydk, line 3: doesn't that need to be below the on: section?

freddydk (approved):

freddydk:
approved with a small comment

aholstrup1 (approved):

.github/workflows/Deploy.yaml
mazhelez: No, it works fine as it is.

PR #544, opened 2023-05-16
Update GetStarted.md

zpeters:
Removed steps that are no longer necessary (updating AL Code).  Since we have just cloned it there is no longer a need to update it since we are using "main" and not "preview"

### DIFF
--- a/Workshop/GetStarted.md
+++ b/Workshop/GetStarted.md
@@ -13,17 +13,6 @@ Click **Use This Template** and select **Create a new repository**. Select your 
 
 After this step, we are done setting up a basic AL-Go for GitHub repository and we only need to add our apps.
 
-For the purpose of this workshop, we will be using the preview version of AL-Go for GitHub. Under **Actions**, select the **Update AL-Go System Files** workflow and click **Run workflow**. Enter **microsoft/AL-Go-PTE@preview** in the **Template Repository URL** field and say **Y** to **Direct COMMIT**.
-
-| ![image](https://user-images.githubusercontent.com/10775043/231537086-4eb4ae5c-6dcb-4e2d-be1b-26df4a67f227.png) |
-|-|
-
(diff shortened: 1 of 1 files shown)


Workshop/GetStarted.md
```diff
@@ -13,17 +13,6 @@ Click **Use This Template** and select **Create a new repository**. Select your
...
-| ![image](https://user-images.githubusercontent.com/10775043/231537086-4eb4ae5c-6dcb-4e2d-be1b-26df4a67f227.png) |
-|-|
-
-The Update AL-Go System Files workflow will start and after a few minutes, your repository have been upgraded to the preview version of AL-Go for GitHub.
-If you click **Actions**, you will also see that a **CI/CD workflow** was kicked off as a result of upgrading your repository:
-
-| ![image](https://user-images.githubusercontent.com/10775043/231539948-48441647-6215-4f5e-abf2-dea1317a8e89.png) |
-|-|
-
 Obviously, the CI/CD workflow didn't compile anything as you didn't add any source code. By clicking the workflow in the list, and scrolling down, you should see:
```
freddydk, line 16: The following lines refer to a build, which was kicked off by the PR with preview and since we didn't update - no build was created.
We should remove the lines you suggest, but replace them with a line stating something like:

```
Click **Actions**, select the **CI/CD** workflow, click **Run workflow** and run the CI/CD workflow on our empty repository. Wait for the CI/CD workflow to complete.
```

Then it naturally flows into the remaining text (I think)

freddydk (changes requested):

Workshop/GetStarted.md
zpeters: just added in some wording and a screenshot.  i did rearrange some of the statements to what made more sense to me.  im still learning, so it might not be correct

zpeters:

freddydk (approved):

PR #563, opened 2023-06-06
Split build job into build and test job

mazhelez:
See https://github.com/orgs/microsoft/projects/521/views/1?pane=issue&itemId=20938099.

- [x] Add new project setting, `splitBuildAndTest`, to indicate that building and testing an AL-Go project will be done in separate jobs
- [x] Generate YAML test job for each build job that contains projects for which `splitBuildAndTest` is set to true. E.g. _Test1_ jobs tests projects, built in _Build1_ job.
- [ ] Create reusable workflow, e.g. _TestALGoProject, to run the tests for an AL-Go project.


lippertmarkus:
@mazhelez is this still planned? :)

mazhelez:
> @mazhelez is this still planned? :)

Yes, in a way.

Splitting the build and test jobs is possible even today by splitting the apps into different AL-Go projects.

Of course, that way apps and test apps will be in different artifacts for delivery/deployment. So, I believe we'll try another approach: having the ability to bundle apps from different AL-Go projects together.

lippertmarkus:
I thought the reason behind this PR is to make the workflows faster by splitting build and test like
1. BC container is started
2. We're not waiting for the container but instead we already download our required symbols via BcNuGet and compile & package our application (build job)
3. As soon as the BC container is ready we publish our app package to the container and run the automated tests (test job)

Or am I mistaken?

mazhelez:
Well, initially, the idea was adding the ability to run build and test for an AL-Go project in _separate_ jobs.

But yes, ultimately, building apps via AL-Go will be done outside a container. Only testing will require creating and setting up a container (or some other environment).

lippertmarkus:
@mazhelez is there already ongoing effort on building apps outside a container with dependencies downloaded from BcNuGet that I can follow?

lippertmarkus:
@mazhelez I think it's not just about running automated tests but also verifying that e.g. app upgrades work by publishing the previous version and then publishing the version a build just compiled. This wouldn't be solved with a separate AL-Go project for the test app.

So at least build and publish should be splitted, while running automated tests could be handled within a separate AL-Go project. Is there an ongoing effort for the former?

mazhelez:
Is that scenario
> app upgrades work by publishing the previous version and then publishing the version a build just compiled

even supported now?
Initially, the idea of the effort was as simple as just being able to build and test the apps in two different jobs, instead of one.

freddydk:
> Is that scenario
> 
> > app upgrades work by publishing the previous version and then publishing the version a build just compiled
> 
> even supported now?

Yes, this is how builds work in AL-Go - the latest released version is used for Cops + upgrade test.

PR #578, opened 2023-06-21
Action for downloading project dependencies

### LINKED ISSUE #450: Build modes with dependent projects do not pick the right artifact
When building dependent projects, the dependencies artifact is always taken from the _Default_ build. See [here](https://github.com/microsoft/AL-Go/blob/56834718a8204caf2b5e44e7456d300468779a1a/Actions/Github-Helper.psm1#L108).

mazhelez:
Fixes #450 

- Define a separate action for fetching project dependencies (`DownloadProjectDependencies`).
- `RunPipeline` only installs the apps that are output from `DownloadProjectDependencies`.
- `Get-Dependencies` works with the provided build mode.

- [x] Documentation
- [x] Release notes
- [x] Propagate changes to workflows


Actions/FetchProjectDependencies/action.yaml
```diff
@@ -0,0 +1,45 @@
...
+      with:
+        path: ${{ inputs.destinationPath}}
+
+    - name: Fetch project dependencies
+      shell: ${{ inputs.shell }}
+      id: FetchDependencies
+      env:
+        gitHubToken: ${{ github.token }}
+      run:
+        ${{ github.action_path }}/FetchProjectDependencies.ps1 -project '${{ inputs.project }}' -buildMode '${{ inputs.buildMode }}' -projectsDependenciesJson '${{ inputs.projectsDependenciesJson }}' -baseBranch '${{ github.base_ref }}' -baseFolder '${{ github.workspace }}'' -destinationPath '${{ inputs.destinationPath }}' -token '${{ github.token }}'
```
freddydk, line 42: inputs should never be used directly in code like this.
inputs should be transferred to env vars and used as env for correct UTF8 conversion, but also for security reasons.

freddydk (changes requested):

Actions/FetchProjectDependencies/action.yaml
mazhelez: Isn't it an issue only when using inline scripts directly in the workflow?

> [Using an action instead of an inline script (recommended)](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-action-instead-of-an-inline-script-recommended)
The recommended approach is to create an action that processes the context value as an argument. This approach is not vulnerable to the injection attack, as the context value is not used to generate a shell script, but is instead passed to the action as an argument

freddydk: That might be, but you will have problems if the input contains UFT8 characters, " or ' - I think we should always use the same pattern as the other actions.

Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -117,44 +119,12 @@ try {
...
-            }
-        }
-        else {
-            Write-Host "No project dependencies"
-        }
-    }
+    $installApps += $installAppsJson | ConvertFrom-Json
+    $installTestApps += $installTestAppsJson | ConvertFrom-Json
 
     if ($repo.appDependencyProbingPaths) {
```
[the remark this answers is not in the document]
mazhelez: Added that to the new action

Actions/FetchProjectDependencies/action.yaml
mazhelez: Switched to using env variables.

Actions/DownloadProjectDependencies/DownloadProjectDependencies.Action.ps1
```diff
@@ -0,0 +1,111 @@
...
+        $downloadedDependencies += $dependency
+    }
+
+    return $downloadedDependencies
+}
+
+$ErrorActionPreference = "Stop"
+Set-StrictMode -Version 2.0
+
+. (Join-Path -Path $PSScriptRoot -ChildPath "..\AL-Go-Helper.ps1" -Resolve)
```
aholstrup1, line 73: Wrap the remaining part of this script in a try catch?

Actions/DownloadProjectDependencies/DownloadProjectDependencies.Action.ps1
```diff
@@ -0,0 +1,111 @@
...
+        }
+    
+        return @{
+            "release_status" = "thisBuild"
+            "version" = "latest"
+            "buildMode" = $dependencyBuildMode
+            "projects" = $dependencyProject
+            "repo" = "$ENV:GITHUB_SERVER_URL/$ENV:GITHUB_REPOSITORY"
+            "branch" = $ENV:GITHUB_REF_NAME
+            "baseBranch" = $ENV:GITHUB_BASE_REF_NAME
```
aholstrup1, line 55: Does the new action run in CI/CD as well? Isn't GITHUB_BASE_REF_NAME only available for builds triggered by pull requests?

Actions/DownloadProjectDependencies/DownloadProjectDependencies.Action.ps1
```diff
@@ -0,0 +1,111 @@
...
+            $dependencyBuildMode = 'Default';
+        }
+    
+        return @{
+            "release_status" = "thisBuild"
+            "version" = "latest"
+            "buildMode" = $dependencyBuildMode
+            "projects" = $dependencyProject
+            "repo" = "$ENV:GITHUB_SERVER_URL/$ENV:GITHUB_REPOSITORY"
+            "branch" = $ENV:GITHUB_REF_NAME
```
aholstrup1, line 54: Isn't GITHUB_REF_NAME always main in our current PRBuild?

mazhelez: Very good observation, I'll test it out.

Actions/DownloadProjectDependencies/DownloadProjectDependencies.Action.ps1
mazhelez: Yes, it runs right before `RunPipeline` step of the build workflow.

I'll double-check the both _CICD_ and _PR Build_ workflows.

Actions/DownloadProjectDependencies/DownloadProjectDependencies.Action.ps1
mazhelez: That or add the try-catch pattern in the action.yaml as the other actions?

```try {...} catch { Write-Host "::Error::Unexpected error when running action ($($_.Exception.Message.Replace("`r",'').Replace("`n",' ')))"; exit 1 }```

Actions/DownloadProjectDependencies/DownloadProjectDependencies.Action.ps1
mazhelez: Defaulted to `GITHUB_REF_NAME`.

Actions/DownloadProjectDependencies/DownloadProjectDependencies.Action.ps1
mazhelez: Defaulted to GITHUB_REF_NAME if GITHUB_BASE_REF_NAME isn't set.

freddydk (approved):
Code looks good, have testsed with a number of my repos and they seem to work.

PR #579, opened 2023-06-23
Change Jobname

freddydk:
This PR makes this:
image
become this:
image

so that project names are first and always visible.
Single project repos will be Build - Default / Build:
image

### DIFF
--- a/Templates/Per Tenant Extension/.github/workflows/_BuildALGoProject.yaml
+++ b/Templates/Per Tenant Extension/.github/workflows/_BuildALGoProject.yaml
@@ -67,8 +67,9 @@ on:
         required: false
         type: string
 jobs:
-  I:
+  BuildALGoProject:
     runs-on: ${{ fromJson(inputs.runsOn) }}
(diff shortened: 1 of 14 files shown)


freddydk:
Modified this after discussions to be:
image

Templates/Per Tenant Extension/.github/workflows/_BuildALGoProject.yaml
```diff
@@ -67,7 +67,7 @@ on:
         required: false
         type: string
 jobs:
-  BuildALGoProject:
+  I:
```
mazhelez, line 70: Can't we set `name` instead?

Also, new idea: what is the nested workflow has the same name as the parent workflow?

freddydk:
@mazhelez - good idea using the same name in the reusable workflow - I have done so, with a slight modification (without the Build) - this seems to be a good mid-way for both left hand side and right hand side UI:
Single project:
image

Multi project:
image

mazhelez (approved):

aholstrup1 (approved):

PR #581, opened 2023-06-26
Fix project dependencies

mazhelez:
_**Problem**_:
Due the way app folders (`appFolders`, `testFolders` etc.) are resolved, when using wild characters for specifying the app folders, the project dependencies are not determined correctly.

Example:
https://github.com/microsoft/ALAppExtensions/blob/builds/22.3.998.0/System%20Application%20Modules/.AL-Go/settings.json
```
{
    "appFolders":  [
        "..\\Modules\\System\\*"
    ],
    "testFolders":  [
        "..\\Modules\\System Test Libraries\\*",
        "..\\Modules\\System Tests\\*"
    ],
    "enableAppSourceCop": false,
    "enablePerTenantExtensionCop":  false,
    "doNotBuildTests": true
}
```

https://github.com/microsoft/ALAppExtensions/actions/runs/5346147623/jobs/9692747522#step:6:200
> - Analyzing project: System Application Modules
Applying settings from D:\a\ALAppExtensions\ALAppExtensions\.github\AL-Go-Settings.json
Applying settings from D:\a\ALAppExtensions\ALAppExtensions\System Application Modules\.AL-Go\settings.json
Warning: Folder ..\Modules\System\* doesn't exist, skipping.
Warning: Folder ..\Modules\System Test Libraries\* doesn't exist, skipping.
Warning: Folder ..\Modules\System Tests\* doesn't exist, skipping.
No apps or tests folders found for project System Application Modules. Scanning for apps in the project folder.

That way there is a missing dependency from "System Application Modules" to "Test Framework":
> ProjectDependenciesJson={"System Application":["Test Framework"],"System Application Modules":[],"Test Framework":[],"Test Stability Tools":[]}

_**Solution**_:
Unify the way app folders are resolved and add tests to that functionality.


Actions/AL-Go-Helper.ps1
```diff
@@ -1779,27 +1815,11 @@ Function AnalyzeProjectDependencies {
...
-            else {
-                Resolve-Path $projectFolder -Relative
-            }
-        }
-        # Default to scanning the project folder if no app folders are specified
-        if (-not $folders) {
-            Write-Host "No apps or tests folders found for project $project. Scanning for apps in the project folder."
-            $folders = @(Get-ChildItem -Path (Join-Path $baseFolder $project) -Recurse | Where-Object { $_.PSIsContainer -and (Test-Path (Join-Path $_.FullName 'app.json')) } | ForEach-Object { $_.FullName.Substring($baseFolder.Length+1) } )
-        }
+        $projectSettings = ReadSettings -path $project -baseFolder $baseFolder
```
freddydk, line 1819: Shouldn't -path be -project ?

mazhelez: Nice catch!

Actions/AL-Go-Helper.ps1
```diff
@@ -645,6 +645,79 @@ function ExcludeUnneededApps {
...
+                $resolvedPaths += Resolve-Path $aLProjectFolder -ErrorAction Ignore | Where-Object { Test-Path (Join-Path $_ 'app.json') }
+            }
+
+            # Get the name of the folder for each resolved path
+            $projectSettings.Value.$_ = $resolvedPaths | ForEach-Object { $(Get-Item $_).Name }
+            
+            if(!$projectSettings.Value.$_) {
+                $projectSettings.Value.$_ = @()
+            }
+        }
```
freddydk, line 675: Easier to read
```suggestion
            # Get the name of the folder for each resolved path
            $projectSettings.Value.$_ = @($resolvedPaths | ForEach-Object { $(Get-Item $_).Name })
        }
```

Actions/AL-Go-Helper.ps1
```diff
@@ -1784,26 +1810,13 @@ Function AnalyzeProjectDependencies {
...
-            }
-            else {
-                Resolve-Path $projectFolder -Relative
-            }
-        }
-        # Default to scanning the project folder if no app folders are specified
-        if (-not $folders) {
-            Write-Host "No apps or tests folders found for project $project. Scanning for apps in the project folder."
-            $folders = @(Get-ChildItem -Path (Join-Path $baseFolder $project) -Recurse | Where-Object { $_.PSIsContainer -and (Test-Path (Join-Path $_.FullName 'app.json')) } | ForEach-Object { $_.FullName.Substring($baseFolder.Length+1) } )
+            return (Resolve-Path (Join-Path $projectPath $_) -Relative)
```
freddydk, line 1819: Using -relative assumes that the current folder is the basefolder right?
We should probably ensure that (like in the resolveprojectfolders)

mazhelez: Added "Push-Location Pop-Location" pattern.

Actions/AL-Go-Helper.ps1
```diff
@@ -645,6 +645,75 @@ function ExcludeUnneededApps {
...
+
+                switch ($true) {
+                    $isTestApp { $projectSettings.Value.testFolders += @($aLProjectFolder.Name) }
+                    $isBcptTestApp { $projectSettings.Value.bcptTestFolders += @($aLProjectFolder.Name) }
+                    Default { $projectSettings.Value.appFolders += @($aLProjectFolder.Name) }
+                }
+            }
+        }
+    }
+    finally {
```
aholstrup1, line 711: Consider adding a catch and writing the error to the terminal? If something goes wrong it might be easier to debug that way

Actions/AL-Go-Helper.ps1
```diff
@@ -645,6 +645,75 @@ function ExcludeUnneededApps {
...
+    Param(
+        [string] $project,
+        [string] $baseFolder,
+        [ref] $projectSettings
+    )
+    $projectPath = Join-Path $baseFolder $project -Resolve
+
+    Push-Location $projectPath
+
+    try {
```
aholstrup1, line 658: What happens if we fail to resolve project folders? I guess this function would still succeed but would the workflow calling this function also succeed?

mazhelez: The `try` is without a `catch` so a potential failure would fail the action.

But the function is generally "error free" (does not throw, given the assumption that the project folder exist, etc.). If an app folder path does not exist or there's no app.json, the path is simply ignored. There's a test to cover that scenario.

Actions/AL-Go-Helper.ps1
mazhelez: If an error is thrown, it is not swallowed, so it'll bubble up in the log.

Tests/ResolveProjectFolders.Test.ps1
```diff
@@ -0,0 +1,199 @@
...
+        
+        $projectSettings.appFolders | Should -HaveCount 1
+        $projectSettings.appFolders | Should -Contain 'app1'
+
+        $projectSettings.testFolders | Should -HaveCount 0
+
+        $projectSettings.bcptTestFolders | Should -HaveCount 0
+    }
+
+    It 'When only non-existing apps are specified, scans all folders and finds all apps' {
```
freddydk, line 163: Isn't this a change in behavior?
Earlier, if you specified one of either appFolders, testFolders or bcptTestFolders, then you were assumed to be "on your own" and no auto-resolve would kick in.

I guess the new behavior won't break anything as it only kicks in if all folders doesn't resolve to anything and as such wouldn't run anything, but in the case where people are switching from Windows to Linux and folders are case sensitive, then you might go from building 1 app to building more apps without understanding why.

mazhelez: The current/existing behavior isn't changed. So, if that causes issues when switching from Windows to Linux, that's another problem.

Actions/AL-Go-Helper.ps1
```diff
@@ -645,6 +645,75 @@ function ExcludeUnneededApps {
...
+        # Filter out app folders that doesn't contain an app.json file
+        'appFolders', 'testFolders', 'bcptTestFolders' | ForEach-Object {
+            $folders = $projectSettings.Value.$_
+            
+            $resolvedPaths = @()
+            foreach($folder in $folders) {
+                $aLProjectFolder = Join-Path $projectPath $folder
+                $resolvedPaths += Resolve-Path $aLProjectFolder -ErrorAction Ignore | Where-Object { Test-Path (Join-Path $_ 'app.json') }
```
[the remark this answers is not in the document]
mazhelez: Add a warning if an app folder isn't resolved to a proper AL project.

Actions/AL-Go-Helper.ps1
```diff
@@ -652,6 +652,82 @@ function ExcludeUnneededApps {
...
+            $resolvedPaths = @()
+            foreach($folder in $folders) {
+                $aLProjectFolder = Join-Path $projectPath $folder
+                $resolvedALProjectsPaths = Resolve-Path $aLProjectFolder -ErrorAction Ignore | Where-Object { Test-Path (Join-Path $_ 'app.json') }
+
+                if($resolvedALProjectsPaths) {
+                    $resolvedPaths += @($resolvedALProjectsPaths)
+                }
+                else {
+                    OutputWarning "The folder '$aLProjectFolder' cannot be resolved or does not contain an app.json file. Skipping."
```
freddydk, line 679: Maybe: The folder '$folder' in project '$project' cannot be.... - in order to not show the full path of the agent

Tests/ResolveProjectFolders.Test.ps1
freddydk: I disagree - it has changed.
Example:
image

Earlier, if you had "appFolders": [ "nonexisting" } - we would NOT scan for appfolders - we would think that you know what you are doing.

Now we will scan for appFolders.

I think the default behavior of scanning for folders is fine if you haven't specified folders yourself - but if you have specified folders - we should not scan

mazhelez: I guess the first attempt to fix the issue with project dependencies was [this commit](https://github.com/microsoft/AL-Go/commit/4134d4aa838aed0733432e40fd541666db99c6b2). Since then, the behavior hasn't changed.

freddydk: In that commit - it only scans for appfolders if settings are not set - and this is how it is in preview today:
image

My point is that the default behavior should only happen if you haven't set the setting yourself.
Default behavior should not kick in if you have misspelled a folder name.

mazhelez: Ok, I think we referred to different process: one is when determining project dependencies and one is when determining the AL projects to built.

Both these processes need to resolve the app folders the same way and with my latest changes it is as you mentioned: only scan for app folders of none are defined.

freddydk (approved):

Tests/ResolveProjectFolders.Test.ps1
mazhelez: Funny thing: the tests failed because they are run on linux in the CI tests. Now they pass both locally (windows) and in the CI (linux).

PR #582, opened 2023-06-27
Fix E2E test failure

freddydk:
E2E Test failure from jobname change


e2eTests/scenarios/ReleaseBranches/runtest.ps1
```diff
@@ -84,7 +84,7 @@ $run = Run-CICD -repository $repository -branch $branch -wait
 Test-ArtifactsFromRun -runid $run.id -folder 'artifacts' -expectedArtifacts @{"Apps"=1} -repoVersion '1.0' -appVersion '1.0'
 
 # Check that no previous release was found
-Test-LogContainsFromRun -runid $run.id -jobName 'Build . - Default  BuildALGoProject' -stepName 'Run pipeline' -expectedText 'No previous release found'
```
mazhelez, line 87: There are more places that reference `BuildALGoProject`. Probably need to be fixed as well.

freddydk: Good catch - I just checked that I could download the log with the new name:-(

mazhelez (approved):

Issue #558, opened 2023-05-31
CI/CD attempts to deploy from feature branch


gntpet:
Hey @freddydk,

I just noticed that CI/CD is trying to deploy to my environment from other branches than **main**
Check the screenshot bellow. As you can see I'm running CICD for a feature branch. To my understanding it should not event try to deploy from this branch. 

![image](https://github.com/microsoft/AL-Go/assets/16342128/ed7afee2-45c2-42b6-b9c8-af04e2e1c3ef)

**My environment setup looks like this:**
![image](https://github.com/microsoft/AL-Go/assets/16342128/8367ab9b-0209-4cc1-aad2-9f9bcc5270c9)

**my settings are:**
**AL-Go-Settings.json**
```
{
  "type": "PTE",
  "templateUrl": "https://github.com/microsoft/AL-Go-PTE@main",
  "treatTestFailuresAsWarnings": true,
  "doNotRunBcptTests": true,
  "enableCodeCop": true,
  "rulesetFile": "../al.ruleset.json",
  "companyName": "TNM Automated Test"
}
```

**settings.json** of my project, which is configured for the environment. 
```
{
    "country": "it",
    "artifact": "bcartifacts/sandbox/22.0/it/latest",
    "appFolders": [],
    "testFolders": [],
    "appDependencyProbingPaths": [
        {
            "repo": ".",
            "release_status": "include",
            "projects": "SRS.core"
        }
    ],
    "installApps": [
        "https://verysecreturl"
    ]
}
```
Is it a bug, or am I'm missing some setup?

Best Regards, 
Gintautas

freddydk:
It should default to only deploy from main, but you should be able to setup from which branches an environment will deploy.
Will investigate this one.

gntpet:
hey @freddydk, do you have any update on this ticket?

rbability:
Did you try to set a
`
"CICDPushBranches": [
    "main",
    "otherbranch"
  ],
`

in your AL-Go-Settings.json? We put in all the branches which should trigger a CI/CD run and it works fine here with the latest AL-Go version.
Do not forget to update AL-Go after changing this setting, so your cicd.yml triggers get updated with your branches.

freddydk:
@rbability - the issue for @gntpet is not that the CI/CD workflow doesn't run - it is that it also triggers the deploy mechanism.

rbability:
[...]

PR #588, opened 2023-07-02
Use GitHub environment branch policies (only if defined)

freddydk:
Fix for issue #558

The problem here was, that if the environment was defined in GitHub, then AL-Go would assume that branch policies was defined on the environment and select all branches itself - waiting for GitHub to reject the branch.

The problem with this approach is that if you have a Teams or Enterprise SKU, then the environment will be auto-created when deploying the first time (even if defined in settings).

The fix for this is to only rely on GitHub branch policies if they actually are defined - else use the default branch policies.

### DIFF
--- a/Actions/ReadSettings/ReadSettings.ps1
+++ b/Actions/ReadSettings/ReadSettings.ps1
@@ -104,8 +104,9 @@ try {
     if ($getenvironments) {
         $environments = @()
         $headers = @{ 
-            "Authorization" = "token $token"
-            "Accept"        = "application/vnd.github.v3+json"
+            "Authorization"        = "token $token"
+            "Accept"               = "application/vnd.github.v3+json"
+            "X-GitHub-Api-Version" = "2022-11-28"
         }
         Write-Host "Requesting environments: $getEnvironments"
         $url = "$($ENV:GITHUB_API_URL)/repos/$($ENV:GITHUB_REPOSITORY)/environments"
@@ -149,7 +150,10 @@ try {
                         Write-Host "GitHub Environment $envName has branch policies, getting branches from GitHub API"
                         $branchesUrl = "$($ENV:GITHUB_API_URL)/repos/$($ENV:GITHUB_REPOSITORY)/environments//$([Uri]::EscapeDataString($envName))/deployment-branch-policies"
                         Write-Host "Getting Branches for $envName from GitHub API"
-                        $branches = @((InvokeWebRequest -Headers $headers -Uri $branchesUrl -ignoreErrors | ConvertFrom-Json).branch_policies | ForEach-Object { $_.name })
+                        Write-Host "BranchesUrl: $branchesUrl"
+                        $result = InvokeWebRequest -Headers $headers -Uri $branchesUrl -ignoreErrors
+                        $result | Out-Host
+                        $branches = @(($result | ConvertFrom-Json).branch_policies | ForEach-Object { $_.name })
                     }
                     else {
                         Write-Host "GitHub Environment $envName dot not have branch policies, using main as default"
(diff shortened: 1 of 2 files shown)


Actions/ReadSettings/ReadSettings.ps1
```diff
@@ -139,13 +141,24 @@ try {
                     $_ -like $getEnvironments -and $_ -notlike '* (PROD)' -and $_ -notlike '* (Production)' -and $_ -notlike '* (FAT)' -and $_ -notlike '* (Final Acceptance Test)'
                 }
             } | Where-Object {
-                Write-Host "Environment: $_"
-                if ($ghEnvironments -contains $_) {
-                    # Environment is GitHub Environment, default branches are controlled on GitHub
-                    $branches = @( '*' )
+                $envName = $_
```
mazhelez, line 144: Consider extracting this check in a separate function to reduce the complexity of the current statement.

The previous `Where-Object`'s complexity could also be reduced by extracting a function like `IsProdName($evvironmentName)`.

Actions/ReadSettings/ReadSettings.ps1
```diff
@@ -139,13 +141,24 @@ try {
...
+                if ($ghEnvironment) {
+                    $branchPolicy = ($ghEnvironment.protection_rules | Where-Object { $_.type -eq "branch_policy" })
+                    if ($branchPolicy) {
+                        Write-Host "GitHub Environment $envName has branch policies, getting branches from GitHub API"
+                        $branchesUrl = "$($ENV:GITHUB_API_URL)/repos/$($ENV:GITHUB_REPOSITORY)/environments/$([Uri]::EscapeDataString($envName))/deployment-branch-policies"
+                        Write-Host "Getting branches for $envName from GitHub API"
+                        $branches = @((InvokeWebRequest -Headers $headers -Uri $branchesUrl -ignoreErrors | ConvertFrom-Json).branch_policies | ForEach-Object { $_.name })
+                    }
+                    else {
+                        Write-Host "GitHub Environment $envName dot not have branch policies, using main as default"
```
mazhelez, line 156:
```suggestion
                        Write-Host "GitHub Environment $envName does not have branch policies, using main as default"
```

mazhelez: Also, some legacy repos might still be using _master_ as default branch.

There's GitHub API to check the default branch: https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#get-a-repository

freddydk: AL-Go currently only supports main as default branch - there is a work-item to change this.
For AL-Go repos - master was never used.

Actions/ReadSettings/ReadSettings.ps1
freddydk: Yeah, the right approach here is to pull this out of ReadSettings (like we have done with projects and more) - but doing this refactoring as part of this bug-fix is slightly more than I have time for:-(

mazhelez: Agreed. Such refactoring can be done at a later stage. However, it's a good practice to avoid increasing the current code complexity (it will help refactoring later on).

freddydk: Agree, but moving the code from the where-object statement to a function becomes a function with many parameters, which doesn't make sense at all (as this function will ever only be called once), and will cause me to have to do an extensive test on this. Furthermore, this function will be deleted when we do the refactoring later.
IMO - Let's fix the bug without introducing new ones.

aholstrup1 (approved):

Issue #608, opened 2023-07-18
When creating a release, throw error if no new artifacts have been added


cegekaJG:
Earlier I found out that if you use the action "Create Release", but haven't had a successful CI/CD since the last release, AL Go will tag the same commit and upload the same artifact from the previous release. However, the changelog in the release description will contain all commits up until the last commit of the target branch.
The workflow should first check if the targeted commit has already been tagged for a release, and throw an error if it has. If someone would want to use the same artifact in multiple releases, it should at least be an option to toggle this check.
Additionally, the changelog should only consist of commits between the last and current release.

freddydk:
The release notes is a duplicate of #559 - and was fixed by https://github.com/microsoft/AL-Go/pull/589
Will check whether the create release notes action could fail and state that no checkins have been made.

freddydk:
Shipped in v3.2

PR #609, opened 2023-07-18
Give error if you are creating a release from the same commit as the latest release (issue608)

freddydk:
Fix for Issue #608

Give error if you are creating a release from the same commit as the latest release


Actions/AL-Go-Helper.ps1
```diff
@@ -627,6 +627,13 @@ function ReadSettings {
     if ($settings.githubRunnerShell -eq "") {
         $settings.githubRunnerShell = $settings.shell
     }
+    # Check that gitHubRunnerShell and Shell is valid
```
aholstrup1, line 630: This was related to another PR wasn't it?

mazhelez (approved):

Issue #555, opened 2023-05-26
protected branch query


rolandpthompson:
This is more a Github thing I think.
But reading the docs its recommended to protect the main branch, which makes sense.
But if we want to protect our branch and only allow merges once specific status checks have passed, how can we do that?

For example, say we want to run the action Increment version no.. It creates a new pull request, but it would never pass any checks (assuming we created them), as it doesnt run anything.

Really speaking, we'd want to make sure any builds (with the tests etc) pass before merging, but this would stop us merging a pull request for just a version increment...

freddydk:
I would assume that a version increment would kick off a build as well.
I think any PR to the main branch should kick off a build. It might not build all projects in the repo, but it should kick off a PR build, isn't that what you are seeing?

freddydk:
Will have a look at why a PR build isn't kicked off by this,

freddydk:
Apparently, this is by design by GitHub. When creating a PR using the GITHUB_TOKEN
https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow

This does have an effect on all workflows, which includes:
- Create new app, testapp, bcpt app
- Add existing app
- Increment version number
- Create development environment
- Create release

And could leave people in problems.
A related problem is that some organizations will disallow GitHub Actions to create PRs (microsoft org being one).
Other policies might say that a PR is required and/or all PR checks needs to succeed.

The only fix suggested by GitHub is to use a personal access token (like the ghTokenWorkflow) to run the action.

A possible fix would be to include a check-box in these workflows for using the GhTokenWorkflow instead of the GITHUB_TOKEN - will investigate this option

freddydk:
Using the GhTokenWorkflow does solve the issue (in this test)
![image](https://github.com/microsoft/AL-Go/assets/10775043/ebedaab2-f62e-4df4-8604-7d9578c03d80)

freddydk:
The PR created by the GITHUB_TOKEN will be created by BOT, the one created by GhTokenWorkflow will be created by the owner of the token:
![image](https://github.com/microsoft/AL-Go/assets/10775043/fcd631b2-f5f6-49dc-828d-3e9931bc3945)

PR #617, opened 2023-07-22
Allow using ghTokenWorkflow in workflows creating pull requests (issue555)

freddydk:
This is a fix for issue #555

AL-Go contains several workflows, which create a Pull Request or pushes code directly.
All (except Update AL-Go System Files) will use the GITHUB_TOKEN to create the PR or commit.
The problem using GITHUB_TOKEN is that is doesn't trigger a pull request build or a commit build.
This is by design: https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow

This PR adds a checkbox to all these actions allowing you to use the GhTokenWorkflow instead of the GITHUB_TOKEN.

### DIFF
--- a/Actions/AL-Go-Helper.ps1
+++ b/Actions/AL-Go-Helper.ps1
@@ -627,6 +627,13 @@ function ReadSettings {
     if ($settings.githubRunnerShell -eq "") {
         $settings.githubRunnerShell = $settings.shell
     }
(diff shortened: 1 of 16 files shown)


mazhelez (approved):
Release notes?

Issue #620, opened 2023-07-24
Error: The environment block used to start a process cannot be longer than 65535 bytes.


mazhelez:
See: https://github.com/microsoft/ALAppExtensions/actions/runs/5643755210/job/15287721005?pr=24198#step:8:201:~:text=Error%3A%20RunPipeline,file%3E%3A%20line%201

> Error: RunPipeline action failed. **Error: The environment block used to start a process cannot be longer than 65535 bytes.**  Your environment block is 73068 bytes long.  Remove some environment variables and try again. Stacktrace: at <ScriptBlock>, C:\Users\runneradmin\AppData\Local\Temp\9597813d-c355-41e8-a483-21987b0ca47b\BcContainerHelper\HelperFunctions.ps1: line 27 at <ScriptBlock>, C:\Users\runneradmin\AppData\Local\Temp\9597813d-c355-41e8-a483-21987b0ca47b\BcContainerHelper\BcContainerHelper.psm1: line 13 at <ScriptBlock>, C:\Users\runneradmin\AppData\Local\Temp\9597813d-c355-41e8-a483-21987b0ca47b\BcContainerHelper\Import-BcContainerHelper.ps1: line 30 at <ScriptBlock>, C:\Users\runneradmin\AppData\Local\Temp\9597813d-c355-41e8-a483-21987b0ca47b\BcContainerHelper\BcContainerHelper.ps1: line 8 at DownloadAndImportBcContainerHelper, D:\a\_actions\microsoft\AL-Go-Actions\preview\AL-Go-Helper.ps1: line 309 at <ScriptBlock>, D:\a\_actions\microsoft\AL-Go-Actions\preview\RunPipeline\RunPipeline.ps1: line 32 at <ScriptBlock>, D:\a\_temp\a2d75a54-af6e-435b-a2de-575451996563.ps1: line 2 at <ScriptBlock>, <No file>: line 1

freddydk:
We probably need to cleanup how we set things.
For a number of the settings, they are both in ENV:SETTINGS and ENV:APPFOLDERS

PR #623, opened 2023-07-25
Remove settings json construct from ENV

freddydk:
Fix for issue #620

Things changed in this PR:
- ReadSettings will now output the settings to GITHUB_OUTPUT instead of GITHUB_ENV
- Actions expecting a settingsJSON will no longer set the input parameter to ENV:_settingsJson, causing the PS script to have an encoding problem (settings might contain special characters) - therefore ReadSettings now needs to output as Base64
- ReadSettings has a parameter on whether to use base64 - set to Y in this version, notifying the action that Base64 response is expected. If this was not done - we might get issues during Update AL-Go System files.
- ReadSecrets and DetermineArtifactUrl did earlier just re-write the settings environment variable. With GITHUB_OUTPUT, they now have to output to GITHUB_OUTPUT as well and the subsequent usages of settings needs to be from there
- ReadSecrets and DetermineArtifactUrl will output base64 if they were given base64 - only ReadSettings will have the useBase64 parameter.

Additional fix:
- removed the string around $ENV:_parameter in the sign action.yaml - no reason for that.

If you want to try this out - you can update AL-Go System Files with freddydk/AL-Go@SettingsEnv - until the branch is deleted (after the merge of this PR)

### DIFF
--- a/Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
+++ b/Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
@@ -13,7 +13,7 @@ Param(
 
 function Set-EnvVariable([string] $name, [string] $value) {
     Write-Host "Assigning $value to $name"
-    Add-Content -Path $env:GITHUB_OUTPUT -Value "$name=$value"
+    Add-Content -Path $env:GITHUB_OUTPUT -Value "$name=$value" -Encoding UTF8
     Add-Content -Path $env:GITHUB_ENV -Value "$name=$value"
 }
 
--- a/Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
+++ b/Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
@@ -13,8 +13,8 @@ Param(
 
 function Set-EnvVariable([string] $name, [string] $value) {
     Write-Host "Assigning $value to $name"
-    Add-Content -Path $env:GITHUB_OUTPUT -Value "$name=$value" -Encoding UTF8
-    Add-Content -Path $env:GITHUB_ENV -Value "$name=$value" -Encoding UTF8
+    Add-Content -Encoding UTF8 -Path $env:GITHUB_OUTPUT -Value "$name=$value"
+    Add-Content -Encoding UTF8 -Path $env:GITHUB_ENV -Value "$name=$value"
 }
 
 $errorActionPreference = "Stop"; $ProgressPreference = "SilentlyContinue"; Set-StrictMode -Version 2.0
--- a/Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
+++ b/Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
@@ -19,8 +19,13 @@ function Set-EnvVariable([string] $name, [string] $value) {
 
 $errorActionPreference = "Stop"; $ProgressPreference = "SilentlyContinue"; Set-StrictMode -Version 2.0
 
-Write-Host $settingsJson
-$settings = $settingsJson | ConvertFrom-Json
+# Support potentially base64 encoded settings
+try {
+    $settings = [System.Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String($settingsJson)) | ConvertFrom-Json | ConvertTo-HashTable
+}
+catch {
+    $settings = $settingsJson | ConvertFrom-Json | ConvertTo-HashTable
+}
 
 if ($project -eq ".") { 
   $project = $settings.repoName 
--- a/Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
+++ b/Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
@@ -21,10 +21,10 @@ $errorActionPreference = "Stop"; $ProgressPreference = "SilentlyContinue"; Set-S
 
 # Support potentially base64 encoded settings
 try {
-    $settings = [System.Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String($settingsJson)) | ConvertFrom-Json | ConvertTo-HashTable
+    $settings = [System.Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String($settingsJson)) | ConvertFrom-Json
 }
 catch {
-    $settings = $settingsJson | ConvertFrom-Json | ConvertTo-HashTable
+    $settings = $settingsJson | ConvertFrom-Json
 }
 
 if ($project -eq ".") { 
--- a/Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
+++ b/Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
@@ -14,7 +14,7 @@ Param(
 function Set-EnvVariable([string] $name, [string] $value) {
     Write-Host "Assigning $value to $name"
     Add-Content -Path $env:GITHUB_OUTPUT -Value "$name=$value" -Encoding UTF8
-    Add-Content -Path $env:GITHUB_ENV -Value "$name=$value"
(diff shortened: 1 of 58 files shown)


Actions/DetermineArtifactUrl/action.yaml
```diff
@@ -22,6 +22,9 @@ inputs:
     required: false
     default: '{"insiderSasToken":""}'
 outputs:
+  SettingsJson:
```
mazhelez, line 25: Why would this action output settings?

Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
```diff
@@ -19,8 +19,13 @@ function Set-EnvVariable([string] $name, [string] $value) {
 
 $errorActionPreference = "Stop"; $ProgressPreference = "SilentlyContinue"; Set-StrictMode -Version 2.0
 
-Write-Host $settingsJson
-$settings = $settingsJson | ConvertFrom-Json
+# Support potentially base64 encoded settings
+try {
```
mazhelez, line 23: This construct seems off. Shouldn't the action's contract specify if the settings are base64-encoded or not?

Actions/ReadSettings/ReadSettings.ps1
```diff
@@ -87,18 +89,23 @@ try {
         }
     }
 
+    Write-Host "OUTPUTS:"
     $outSettingsJson = $outSettings | ConvertTo-Json -Depth 99 -Compress
-    Add-Content -Path $env:GITHUB_OUTPUT -Value "SettingsJson=$outSettingsJson"
-    Add-Content -Path $env:GITHUB_ENV -Value "Settings=$OutSettingsJson"
-    Write-Host "SettingsJson=$outSettingsJson"
+    if ($useBase64) {
+        Add-Content -Path $env:GITHUB_OUTPUT -Value "SettingsJson=$([Convert]::ToBase64String([System.Text.Encoding]::UTF8.GetBytes($outSettingsJson)))"
```
mazhelez, line 95: Readability: Move `Add-Content -Path $env:GITHUB_OUTPUT -Value "SettingsJson=$SettingsJsonValue"` outside of the if-else.

mazhelez, line 95: Isn't the actual problem that we output every setting (line 85 in this file: `Add-Content -Path $env:GITHUB_ENV -Value "$setting=$($settingValue | ConvertTo-Json -Depth 99 -Compress)`).

So, with that change, every setting will be in `$env:GITHUB_ENV` and all the settings (JSON-formatted and maybe base64-encoded) will the in the `$env:GITHUB_OUTPUT`. 

I'm not convinced the issue will be solved.

freddydk: The real problem is that settings where as individual values + as a json construct 2 or 3 times.
This PR removes the json constructs.
It might be a good idea to generally look at what ENV variables are needed - but for now, this fixes the issue (I am convinced:-)) and we can fix the other ones later,

Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
freddydk: Due to the nature of our preview - we will have old versions of AL-Go workflows calling into this function with a json construct which isn't base64 - and every new version will be base64.
I assume that once we have shipped this - we will be able to remove the support for non-base64 in preview - but for now, we have to support both in order to not break partners.

I could also have added the useBase64 parameter - but that one (as this construct) is temporary and can be removed later.

Actions/DetermineArtifactUrl/action.yaml
freddydk: DetermineArtifactUrl updates the settings (in the "old" days it re-wrote the ENV:SETTINGS)
We could instead return the artifactURL, but then we would have to do more changes to CI/CD in order to get this artifact into the settings - or use it from outside the setting.

I didn't want to do major restructuring - instead I was aiming at creating the least disruptive change for removing the ENV problem.

Actions/ReadSettings/ReadSettings.ps1
freddydk: This is a build of ALAppExtensions with the new ENV mechanism:

freddydk: fixed

Tests/TestActionsHelper.psm1
```diff
@@ -83,21 +83,29 @@ function YamlTest {
                 $yaml.AppendLine("  $($name):") | Out-Null
                 $yaml.AppendLine("    description: $description") | Out-Null
                 $yaml.AppendLine("    required: $($required.ToString().ToLowerInvariant())") | Out-Null
-                $envLines.AppendLine("        _$($name): `${{ inputs.$($name) }}")
-                if ($type -eq "System.String" -or $type -eq "System.Int32") {
-                    $parameterString += " -$($name) `$ENV:_$($name)"
+                if ($name -eq 'settingsJson') {
```
mazhelez, line 86: A comment in here would be useful to understand why the special case for `settingsJson`.

Actions/ReadSecrets/ReadSecrets.ps1
```diff
@@ -113,12 +123,17 @@ try {
         }) -join ', ')"
     }
 
+    Write-Host "OUTPUTS:"
+    $outSettingsJson = $outSettings | ConvertTo-Json -Depth 99 -Compress
+    Write-Host "- SettingsJson=$outSettingsJson"
+    if ($useBase64) {
+        $outSettingsJson = [Convert]::ToBase64String([System.Text.Encoding]::UTF8.GetBytes($outSettingsJson))
```
mazhelez, line 130: Since at this point the `$outSettingsJson` may contain secrets, it's quite dangerous to encode it as the encoded string is in no way masked (see [here](https://github.com/freddydk/ALAppExtensions/actions/runs/5657371091/job/15327074618#step:8:100) for example).

mazhelez (changes requested):
Settings may contain secrets so encoding them is a security concern. This needs to be addressed.

Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
```diff
@@ -19,8 +19,13 @@ function Set-EnvVariable([string] $name, [string] $value) {
...
+    $settings = [System.Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String($settingsJson)) | ConvertFrom-Json
```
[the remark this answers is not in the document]
freddydk: Yes, ConvertTo-HashTable is a ContainerHelper function and ContainerHelper is not loaded here.
settings will then be a PSCustomObject

Actions/ReadSecrets/ReadSecrets.ps1
freddydk: Adding the entire settings construct will mask the encoded settings and minimize the clutter in the logs

Actions/ReadSecrets/action.yaml
```diff
@@ -16,16 +16,20 @@ inputs:
     description: Specifies the parent telemetry scope for the telemetry signal
     required: false
     default: '7b7d'
+outputs:
+  SettingsJson:
+    description: Settings in compressed Json format
```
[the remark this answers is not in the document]
freddydk: Or, ReadSecrets and DetermineArtifactUrl should not re-write the entire settings construct.

freddydk:
In order to test that the action.yaml matches the DetermineArtifactUrl.ps1, I have renamed the DetermineArtifactUrl.Action.ps1 to DetermineArtifactUrl and added tests to call YamlTest (which builds the yaml based on the .ps1 and checks that they are identical)

PR #628, opened 2023-07-27
Do not output settings from ReadSettings in GITHUB_ENV

### LINKED ISSUE #620: Error: The environment block used to start a process cannot be longer than 65535 bytes.

mazhelez:
Follow-up on #623 
Fix #620 

- Introduce new action _ReadALGoSettings_ to accommodate backwards compatibility (as _Update AL-Go System Files_ would still run the old action in _preview_). Deprecate _ReadSettings_.
- _ReadALGoSettings_ does not output settings in GITHUB_ENV.
- Remove references to settings through `$env`.

Test run: https://github.com/mazhelez/ALAppExtensions/actions/runs/5677832733/job/15388103495

TODO:
- [ ] Propagate changes to _AppSource App_ template
- [ ] Release notes & doc
- [ ] Tests?
- [ ] Fix CI tests

### DIFF
--- a/Actions/ReadALGoSettings/ReadALGoSettings.ps1
+++ b/Actions/ReadALGoSettings/ReadALGoSettings.ps1
@@ -14,7 +14,9 @@ Param(
     [Parameter(HelpMessage = "Indicates whether this is called from a release pipeline", Mandatory = $false)]
     [bool] $release,
     [Parameter(HelpMessage = "Specifies which properties to get from the settings file, default is all", Mandatory = $false)]
-    [string] $get = ""
+    [string] $get = "",
+    [Parameter(HelpMessage = "Specifies whether to output the settings to the GITHUB_ENV file", Mandatory = $false)]
+    [switch] $outputToEnv
 )
 
 $errorActionPreference = "Stop"; $ProgressPreference = "SilentlyContinue"; Set-StrictMode -Version 2.0
@@ -79,17 +81,24 @@ try {
         $setting = $_.Trim()
         $settingValue = $settings."$setting"
         $outSettings += @{ "$setting" = $settingValue }
-        if ($settingValue -is [System.Collections.Specialized.OrderedDictionary]) {
-            Add-Content -Path $env:GITHUB_ENV -Value "$setting=$($settingValue | ConvertTo-Json -Depth 99 -Compress)"
-        }
-        else {
-            Add-Content -Path $env:GITHUB_ENV -Value "$setting=$settingValue"
+
+        if($outputToEnv) {
+            if ($settingValue -is [System.Collections.Specialized.OrderedDictionary]) {
+                Add-Content -Path $env:GITHUB_ENV -Value "$setting=$($settingValue | ConvertTo-Json -Depth 99 -Compress)"
+            }
+            else {
+                Add-Content -Path $env:GITHUB_ENV -Value "$setting=$settingValue"
+            }
         }
     }
 
     $outSettingsJson = $outSettings | ConvertTo-Json -Depth 99 -Compress
     Add-Content -Path $env:GITHUB_OUTPUT -Value "SettingsJson=$outSettingsJson"
-    Add-Content -Path $env:GITHUB_ENV -Value "Settings=$OutSettingsJson"
+
+    if($outputToEnv) {
+        Add-Content -Path $env:GITHUB_ENV -Value "SettingsJson=$outSettingsJson"
+    }
+    
     Write-Host "SettingsJson=$outSettingsJson"
 
     $gitHubRunner = $settings.githubRunner.Split(',').Trim() | ConvertTo-Json -compress
@@ -188,7 +197,11 @@ try {
         }
         $environmentsJson = $json | ConvertTo-Json -Depth 99 -compress
         Add-Content -Path $env:GITHUB_OUTPUT -Value "EnvironmentsJson=$environmentsJson"
-        Add-Content -Path $env:GITHUB_ENV -Value "environments=$environmentsJson"
+        
+        if($outputToEnv) {
+            Add-Content -Path $env:GITHUB_ENV -Value "EnvironmentsJson=$environmentsJson"
+        }
+        
         Write-Host "EnvironmentsJson=$environmentsJson"
         Add-Content -Path $env:GITHUB_OUTPUT -Value "EnvironmentCount=$($environments.Count)"
         Write-Host "EnvironmentCount=$($environments.Count)"
--- a/Actions/ReadALGoSettings/ReadALGoSettings.ps1
+++ b/Actions/ReadALGoSettings/ReadALGoSettings.ps1
@@ -84,29 +84,29 @@ try {
 
         if($outputToEnv) {
             if ($settingValue -is [System.Collections.Specialized.OrderedDictionary]) {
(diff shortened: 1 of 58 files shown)


Templates/Per Tenant Extension/.github/workflows/CICD.yaml
```diff
@@ -170,7 +175,7 @@ jobs:
         with:
           shell: powershell
           parentTelemetryScopeJson: ${{ needs.Initialization.outputs.telemetryScopeJson }}
-          templateUrl: ${{ env.templateUrl }}
+          templateUrl: ${{ fromJson(steps.ReadSettings.outputs.SettingsJson).templateUrl }}
```
mazhelez, line 178: Instead of doing that, maybe potentially allow certain settings in GITHUB_ENV?

Actions/ReadALGoSettings/ReadALGoSettings.ps1
```diff
@@ -14,7 +14,9 @@ Param(
     [Parameter(HelpMessage = "Indicates whether this is called from a release pipeline", Mandatory = $false)]
     [bool] $release,
     [Parameter(HelpMessage = "Specifies which properties to get from the settings file, default is all", Mandatory = $false)]
-    [string] $get = ""
+    [string] $get = "",
+    [Parameter(HelpMessage = "Specifies whether to output the settings to the GITHUB_ENV file", Mandatory = $false)]
+    [switch] $outputToEnv
```
mazhelez, line 19: This parameter should be deprecated and eventually removed.

freddydk:
Two quick comments:
- A lot of actions which takes settings as parameter will put settings into an ENV called _settings - so with this you have removed one ENV instance and another is still there
- RunPipeline uses Settings.Artifact - which is why determineArtifactUrl rewrote that.

freddydk:
and if you need to transfer settings to actions without ENV - you need base64 encoding to ensure special characters

Templates/Per Tenant Extension/.github/workflows/CICD.yaml
mazhelez: Good enough for now. Potential restructuring might utulize a prettier solutions.

freddydk:
I don't like the idea of adding a new action and deprecating another action. Especially ReadSettings, ReadSecrets and CheckForUpdates should work as before. If they change behavior it should be done in a way that is backwards compatible.

I don't think it is a goal to get rid of all ENV variables. ENV variables are good inside jobs. We should add some more control over which ENV variables are created though.

Actions/ReadSettings/action.yaml
```diff
@@ -71,7 +71,7 @@ runs:
         _includeProduction: ${{ inputs.includeProduction }}
         _release: ${{ inputs.release }}
         _get: ${{ inputs.get }}
-      run: try { ${{ github.action_path }}/ReadSettings.ps1 -actor $ENV:_actor -token $ENV:_token -parentTelemetryScopeJson $ENV:_parentTelemetryScopeJson -project $ENV:_project -getenvironments $ENV:_getenvironments -includeProduction ($ENV:_includeProduction -eq 'Y') -release ($ENV:_release -eq 'Y') -get $ENV:_get } catch { Write-Host "::Error::Unexpected error when running action ($($_.Exception.Message.Replace("`r",'').Replace("`n",' ')))"; exit 1 }
+      run: try { ${{ github.action_path }}/ReadALGOSettings.ps1 -actor $ENV:_actor -token $ENV:_token -parentTelemetryScopeJson $ENV:_parentTelemetryScopeJson -project $ENV:_project -getenvironments $ENV:_getenvironments -includeProduction ($ENV:_includeProduction -eq 'Y') -release ($ENV:_release -eq 'Y') -get $ENV:_get -outputToEnv } catch { Write-Host "::Error::Unexpected error when running action ($($_.Exception.Message.Replace("`r",'').Replace("`n",' ')))"; exit 1 }
```
mazhelez, line 74:
```suggestion
      run: try { ${{ github.action_path }}/../ReadALGoSettings/ReadALGoSettings.ps1 -actor $ENV:_actor -token $ENV:_token -parentTelemetryScopeJson $ENV:_parentTelemetryScopeJson -project $ENV:_project -getenvironments $ENV:_getenvironments -includeProduction ($ENV:_includeProduction -eq 'Y') -release ($ENV:_release -eq 'Y') -get $ENV:_get -outputToEnv } catch { Write-Host "::Error::Unexpected error when running action ($($_.Exception.Message.Replace("`r",'').Replace("`n",' ')))"; exit 1 }
```

freddydk (changes requested):
I don't think this is the right way!

You will also still have env:_settingsJson in all actions, meaning that you could instead just rename all occurrences of the settings json construct env variable to _settingsJson then you would remove all duplicates in a much easier way.

mazhelez:
> I don't think this is the right way!
> 
> You will also still have env:_settingsJson in all actions, meaning that you could instead just rename all occurrences of the settings json construct env variable to _settingsJson then you would remove all duplicates in a much easier way.

That would be an extra optimization that isn't needed at the moment.
In my opinion, settings should be read (`ReadSettings` action) only if they are needed for the yaml itself. Actions can read the settings separately.

freddydk:
> > I don't think this is the right way!
> > You will also still have env:_settingsJson in all actions, meaning that you could instead just rename all occurrences of the settings json construct env variable to _settingsJson then you would remove all duplicates in a much easier way.
> 
> That would be an extra optimization that isn't needed at the moment. In my opinion, settings should be read (`ReadSettings` action) only if they are needed for the yaml itself. Actions can read the settings separately.

Disagree - then you would have to transfer all variables needed to resolve settings to all functions - that doesn't make sense IMO

PR #633, opened 2023-07-29
Remove settings json construct from ENV

freddydk:
Fix for issue https://github.com/microsoft/AL-Go/issues/620

Things changed in this PR:

ReadSettings will now output the settings to GITHUB_OUTPUT instead of GITHUB_ENV
Actions expecting a settingsJSON will no longer set the input parameter to ENV:_settingsJson, causing the PS script to have an encoding problem (settings might contain special characters) - therefore ReadSettings now needs to output as Base64
ReadSettings has a parameter on whether to use base64 - set to Y in this version, notifying the action that Base64 response is expected. If this was not done - we might get issues during Update AL-Go System files.
ReadSecrets and DetermineArtifactUrl did earlier just re-write the settings environment variable. With GITHUB_OUTPUT, they now have to output to GITHUB_OUTPUT as well and the subsequent usages of settings needs to be from there
ReadSecrets and DetermineArtifactUrl will output base64 if they were given base64 - only ReadSettings will have the useBase64 parameter.
Additional fix:

removed the string around $ENV:_parameter in the sign action.yaml - no reason for that.
If you want to try this out - you can update AL-Go System Files with freddydk/AL-Go@SettingsEnv - until the branch is deleted (after the merge of this PR)

### DIFF
--- a/.github/workflows/CleanupTempRepos.yaml
+++ b/.github/workflows/CleanupTempRepos.yaml
@@ -45,7 +45,7 @@ jobs:
           if ($orgmap.PSObject.Properties.Name -eq $githubOwner) {
             $githubOwner = $orgmap."$githubOwner"
           }
-          Add-Content -Path $env:GITHUB_OUTPUT -Value "githubOwner=$githubOwner"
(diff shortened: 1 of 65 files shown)


freddydk:
We will not use this approach

PR #635, opened 2023-07-30
E2E test

freddydk:
- End 2 End test covering special characters (on Windows and Linux)
- Support for running End 2 End tests locally on new single repo AL-Go Development setup

Fails with this:
image

when run on preview today.
Running on main, the test passes and produces output like:
image

### DIFF
--- a/Scenarios/Contribute.md
+++ b/Scenarios/Contribute.md
@@ -67,7 +67,7 @@ You can run the *Cleanup after failed E2E* workflow to cleanup these repositorie
 
 You can also run the end to end tests directly from VS Code, by providing the following global variables:
 
-|Variable|Tyoe|Description|
+|Variable|Type|Description|
 |---|---|---|
 |$global:E2EgitHubOwner| String | The GitHub owner of the test repositories (like `freddydk` or `microsoft`) |
 |$global:SecureE2EPAT| SecureString | A personal access token with workflow permissions |
--- a/Scenarios/Contribute.md
+++ b/Scenarios/Contribute.md
@@ -66,11 +66,16 @@ End to end testing will create a lot of repositories called tmpXYZ, where XYZ is
 You can run the *Cleanup after failed E2E* workflow to cleanup these repositories.
 
 You can also run the end to end tests directly from VS Code, by providing the following global variables:
-- $global:E2EgitHubOwner
-- $global:SecureE2EPAT
-- $global:SecureAdminCenterApiToken
-- $global:SecureLicenseFileUrl
-- $global:SecureInsiderSasToken
+
+|Variable|Tyoe|Description|
+|---|---|---|
+|$global:E2EgitHubOwner| String | The GitHub owner of the test repositories (like `freddydk` or `microsoft`) |
+|$global:SecureE2EPAT| SecureString | A personal access token with workflow permissions |
+|$global:SecureAdminCenterApiToken| SecureString | Admin Center API Credentials |
+|$global:SecureLicenseFileUrl| SecureString | Direct download URL to a license file |
+|$global:SecureInsiderSasToken | SecureString | The Insider SAS Token from https://aka.ms/collaborate  |
+|$global:pteTemplate| String | URL for your PTE template (like `freddyk/AL-Go-PTE@main` or `freddydk/AL-Go@main\|Templates/Per Tenant Extension` for direct AL-Go development) |
+|$global:appSourceTemplate| String | URL for your PTE template (like `freddyk/AL-Go-AppSource@main` or `freddydk/AL-Go@main\|Templates/AppSource App` for direct AL-Go development) |
 
 Where E2EgitHubOwner is of type string and the other global variables is of type SecureString.
 Running end to end tests in VS Code will use your 3 deployed template repositories for running the tests.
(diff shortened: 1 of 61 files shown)


e2eTests/e2eTestHelper.psm1
```diff
@@ -346,13 +346,27 @@ function CreateAlGoRepository {
     if (!$template.Contains('@')) {
         $template += '@main'
     }
+    $waitMinutes = Get-Random -Minimum 0 -Maximum 4
+    $templateFolder = ''
+    if ($template.Contains('|')) {
+        # In order to run tests on the direct AL-Go Development branch, specify the folder in which the template is located after a | character in template
+        # example: "https://github.com/freddydk/AL-Go@branch|Templates/Per Tenant Extension"
+        $templateFolder = $template.Split('|')[1]
```
mazhelez, line 354: What's the reasoning for having the template folder as part for the template parameter?
Isn't true that the template folder can be either `Templates/Per Tenant Extension" or "Templates/AppSource App"?

If yes, then another parameter $templateType (with possible values "PTE" and "AppSource App") sounds like a better approach.

Also, whether you keep that design or not, consider using a regex to match the different types:
E.g. `$template -match "https://github.com/(.*)/AL-Go@(.*)\|(.*)`.

mazhelez (approved):

e2eTests/e2eTestHelper.psm1
```diff
@@ -346,13 +346,27 @@ function CreateAlGoRepository {
     if (!$template.Contains('@')) {
         $template += '@main'
```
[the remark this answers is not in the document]
freddydk: Thanks, will move that

e2eTests/e2eTestHelper.psm1
freddydk: If we are running E2E tests locally, we need to set $global:pteTemplate and $global:appSourceTemplate (which are default values to the parameters in the E2E tests (https://github.com/microsoft/AL-Go/blob/6bc6da7de33a23ed410714f7db7c7e36d62f561d/e2eTests/scenarios/UseProjectDependencies/runtest.ps1#L6C5-L6C5))
For normal E2E test running, there will not be a TemplateFolder and I didn't want to add a parameter more all over the place, just in order to be able to support running locally on direct AL-Go repo.

e2eTests/e2eTestHelper.psm1
```diff
@@ -371,9 +385,31 @@ function CreateAlGoRepository {
...
-    Copy-Item (Join-Path (Get-Item "$tempRepoPath\*").FullName '*') -Destination . -Recurse -Force
+    Copy-Item (Join-Path (Get-Item "$tempRepoPath/*/$templateFolder").FullName '*') -Destination . -Recurse -Force
     Remove-Item -Path $tempRepoPath -Force -Recurse
     Remove-Item -Path $zipFileName -Force
+    if ($templateFolder) {
+        # This is a direct AL-Go development repository
+        # Replace URL's + references to microsoft/AL-Go-Actions with $templateOwner/AL-Go/Actions
+        Get-ChildItem -Path . -File -Recurse | ForEach-Object {
+            $file = $_.FullName
+            $lines = ([string](Get-Content -Encoding UTF8 -Raw -path $file)).Replace("`r","").Split("`n")
```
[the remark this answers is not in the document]
freddydk: The issue here, I think is Linux (LF) vs. Windows (CRLF) line endings.
Files might be written in one format and read on the other - which have caused issues in earlier versions of AL-Go for GitHub.
This construct is used successfully for Update AL-Go System files to avoid detecting changes when run on different OS' in Update AL-Go System Files - haven't tried to use Get-Content without the -raw here

e2eTests/e2eTestHelper.psm1
freddydk: Have added this info to Contribute.md + an explanationation of what the various variables needs to be set to

Scenarios/Contribute.md
```diff
@@ -66,14 +66,16 @@ End to end testing will create a lot of repositories called tmpXYZ, where XYZ is
...
-- $global:E2EgitHubOwner
-- $global:SecureE2EPAT
-- $global:SecureAdminCenterApiToken
-- $global:SecureLicenseFileUrl
-- $global:SecureInsiderSasToken
-
-Where E2EgitHubOwner is of type string and the other global variables is of type SecureString.
-Running end to end tests in VS Code will use your 3 deployed template repositories for running the tests.
+
+|Variable|Tyoe|Description|
```
mazhelez, line 70: Typo

```suggestion
|Variable|Type|Description|
```

e2eTests/e2eTestHelper.psm1
mazhelez: I think I got it now.

But in that case, the template folder can be determined directly, right?
If you set `$global:pteTemplate` to `freddydk/AL-Go@main\|Templates/Per Tenant Extension` then the template folder after `|` is redundant. Because it's easy to determine that it's direct AL-Go and it's a PTE template (_UpdateALGoSystemFiles_ does it).

Also, in the file you mentioned there's this line
[$template = "https://github.com/$($pteTemplate)@main"](https://github.com/microsoft/AL-Go/blob/6bc6da7de33a23ed410714f7db7c7e36d62f561d/e2eTests/scenarios/UseProjectDependencies/runtest.ps1#L52).
Reading the doc, wouldn't `$template` then be `https://github.com/freddydk/AL-Go@main\|Templates/Per Tenant Extension@main`?

freddydk: Yeah, well this is the code that creates the repo - it doesn't know which kind of repo it is until it has downloaded the folder which the templateFolder points out.

freddydk: fixed with the latest checkin - @main gets added later

mazhelez: Ok then, resolving.

mazhelez (approved):

PR #642, opened 2023-08-02
Clean up PreGate check

aholstrup1:


mazhelez (approved):

freddydk (approved):
I guess this will break people on preview until they have updated AL-Go System Files, which is OK

mazhelez (approved):

PR #650, opened 2023-08-07
Secrets refactoring

freddydk:
With this PR, secrets are refactored kind of what Settings were in https://github.com/microsoft/AL-Go/pull/640
Additionally a few bug-fixes were found and included.

ReadSecrets will create an environment variable called Secrets, which contains all requested secrets encoded with base64.
Secrets are no longer created as individual secrets.
All Secrets requested during the call to ReadSecrets will be available in $env:Secrets - empty if the secret doesn't exist.

Part of this refactoring was also to ensure that ReadSecrets and DetermineArtifactUrl didn't have to rewrite the settings environment variable.

ReadSecrets will now determine which secrets are used by AppDependencyProbingPaths and include those in $env:Secrets - and then DownloadProjectDependencies will use the value of the secret to download the dependencies.

ReadSettings in CI/CD will ask for the value of the artifact setting to be read.
DetermineArtifactUrl will now (based on this) determine the artifactUrl and set the environment variable to the "final" value.
RunPipeline will now take artifact as a parameter and use env:artifact as the value.

Additional refactoring, which will be done in separate PRs:
- Split AnalyzeRepo into a series of functions for analyzing specific areas of the repo (to avoid -doNotXXX)
- Use $env:Settings instead of using ReadSettings inside actions
- Add end 2 end tests for delivery and deployment testing the areas, which failed here
- Add end 2 end tests for include in appDependencyProbingPaths
- Add CI test, testing that all actions have a try/catch
- in a few places, secrets are used in yaml files with fromJson(env.Secrets).ghTokenWorkflow - remove this and use $env:secrets inside the action instead

Bugs found during this refactoring:

### Unify try/catch for actions
All Actions now have a code snippet in action.yaml, which sets errorpreference and uses try/catch to give good error messages with stack trace. try/catches inside the action.ps1 is only used for telemetry reasons.

### DownloadProjectDependencies didn't work for appDependencyProbingPaths
The Download function didn't analyze the settings and used appDependencyProbingPaths as a json structure, which it wasn't. It was always failing, but silently due to a missing multiline symbol in run: in the action.yaml. It might still have worked as RunPipeline itself also was also analyzing appDependencyProbingPaths

### Deploy and Deliver didn't work properly after the settings refactoring
After the Settings refactoring, where settings are no longer available as environment variables, Deploy and Deliver was using [System.Environment]::GetEnvironmentVariable to check for settings and secrets - this obviously doesn't work if settings and secrets are no longer individual environment variables.

### RunPipeline was requesting wrong secrets
RunPipeline was asking for the StorageContext secret, which it didn't use - instead it needed applicationInsightsConnectionString, which it didn't get. With the new mechanism, we just use secrets.applicationInsightsConnectionString - and if that property is set, we know that we requested the secret.

### CalculateArtifactNames were emitting both OUTPUT variables and ENV variables
And in the BuildALGoProject yaml file - some were used as output and some as ENV.
Now only OUTPUT variables are emitted and used all over.
BTW buildMode is output, but it seems like it is never used???

### ReadSecrets was called with ghTokenWorkflow=${{ env.GHTOKENWORKFLOWSECRETNAME }}
In multiple places, ReadSecrets was called with ghTokenWorkflow=${{ env.GHTOKENWORKFLOWSECRETNAME }}. The problem is, that the mapping to another secret is done inside ReadSecrets and shouldn't be part of the call. The caller should just ask for GhTokenWorkflow secret and then he should get that secret set from Secrets or Azure KeyVault with whatever name it is mapped to.


Actions/AddExistingApp/AddExistingApp.ps1
```diff
@@ -83,6 +83,7 @@ $telemetryScope = $null
 $bcContainerHelperPath = $null
 
 # IMPORTANT: No code that can fail should be outside the try/catch
+# IMPORTANT: All actions needs a try/catch here and not only in the yaml file, else they can silently fail
```
mazhelez, line 86: That doesn't sound right.

The actions were probably silently failing because of the try/catch in the ps1 files: it catches the exception but doesn't throw is again. So, the outer try/catch (in the YAML files) never catches anything and, thus, never sets the exit code to 1.

Either one of the try/catch-es is redundant.

Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
```diff
@@ -9,49 +9,54 @@ Param(
...
-}
-
-# Set this build artifacts name
-'Apps', 'TestApps' | ForEach-Object {
-  $name = "ThisBuild$($_)ArtifactsName"
-  $value = "thisbuild-$($projectName)-$($buildMode)$($_)"
-  
-  Set-EnvVariable -name $name -value $value
+catch {
+    Write-Host "::ERROR::AnalyzeTests action failed.$([environment]::Newline)Error: $($_.Exception.Message)$([environment]::Newline)Stacktrace: $($_.scriptStackTrace)"
```
mazhelez, line 61: Wrong action name here.

Actions/DownloadProjectDependencies/README.md
```diff
@@ -6,7 +6,9 @@ The action constructs arrays of paths to .app files, that are dependencies of th
 ## INPUT
 
 ### ENV variables
-none
+| Name | Description |
+| :-- | :-- |
+| Settings | env.Settings must be set by a prior call to the ReadSettings Action |
```
mazhelez, line 11: Maybe...

```suggestion
| Settings | env.Settings must be set by a prior call to the DownloadProjectDependencies Action |
```

Actions/ReadSecrets/ReadSecrets.ps1
```diff
@@ -30,11 +29,16 @@ try {
        Write-Host "Other process terminated abnormally"
     }
 
-    Import-Module (Join-Path $PSScriptRoot ".\ReadSecretsHelper.psm1")
+    . (Join-Path -Path $PSScriptRoot -ChildPath "..\AL-Go-Helper.ps1" -Resolve)
+    $BcContainerHelperPath = DownloadAndImportBcContainerHelper -baseFolder $ENV:GITHUB_WORKSPACE
+
+    import-module (Join-Path -path $PSScriptRoot -ChildPath "..\TelemetryHelper.psm1" -Resolve)
```
mazhelez, line 35:
```suggestion
    Import-Module (Join-Path -path $PSScriptRoot -ChildPath "..\TelemetryHelper.psm1" -Resolve)
```

Actions/DetermineArtifactUrl/DetermineArtifactUrl.ps1
```diff
@@ -22,40 +21,31 @@ try {
...
-        $insiderSasToken = ""
-    }
-    $projectSettings = $env:Settings | ConvertFrom-Json | ConvertTo-HashTable
-    $projectSettings = AnalyzeRepo -settings $projectSettings -project $project -doNotCheckArtifactSetting -doNotIssueWarnings
-    $artifactUrl = Determine-ArtifactUrl -projectSettings $projectSettings -insiderSasToken $insiderSasToken
+    $secrets = $env:Secrets | ConvertFrom-Json | ConvertTo-HashTable
+    $insiderSasToken = [System.Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String($Secrets.insiderSasToken))
+    $settings = $env:Settings | ConvertFrom-Json | ConvertTo-HashTable
+    $settings = AnalyzeRepo -settings $settings -project $project -doNotCheckArtifactSetting -doNotCheckAppDependencyProbingPaths -doNotIssueWarnings
+    $artifactUrl = Determine-ArtifactUrl -projectSettings $settings -insiderSasToken $insiderSasToken
```
mazhelez, line 28: `AnalyzeRepo` is already calling `Determine-ArtifactUrl` if the `doNotCheckArtifactSetting` switch isn't set.

Actions/DownloadProjectDependencies/DownloadProjectDependencies.Action.ps1
```diff
@@ -84,45 +80,58 @@ function DownloadDependenciesFromCurrentBuild($baseFolder, $project, $projectsDe
...
+    Write-Host "::group::Downloading project dependencies from current build"
+    $projectsDependencies = $projectsDependenciesJson | ConvertFrom-Json | ConvertTo-HashTable
+    $downloadedDependencies += DownloadDependenciesFromCurrentBuild -baseFolder $baseFolder -project $project -projectsDependencies $projectsDependencies -buildMode $buildMode -destinationPath $destinationPath
+    Write-Host "::endgroup::"
 
-Write-Host "::group::Downloading project dependencies from probing paths"
-$downloadedDependencies += DownloadDependenciesFromProbingPaths -baseFolder $baseFolder -project $project -destinationPath $destinationPath
-Write-Host "::endgroup::"
+    Write-Host "::group::Downloading project dependencies from probing paths"
+    $settings = $downloadedDependencies += DownloadDependenciesFromProbingPaths -baseFolder $baseFolder -project $project -destinationPath $destinationPath
```
mazhelez: What is that? Too convoluted.

Actions/DownloadProjectDependencies/DownloadProjectDependencies.Action.ps1
```diff
@@ -9,16 +9,12 @@ Param(
...
 function DownloadDependenciesFromProbingPaths($baseFolder, $project, $destinationPath) {
-    $projectSettings = ReadSettings -baseFolder $baseFolder -project $project
-    
-    $probingPaths = $projectSettings.appDependencyProbingPaths | ConvertFrom-Json
-
-    $downloadedDependencies = @()
-    if ($probingPaths) {
-        $downloadedDependencies += Get-Dependencies -probingPathsJson $repo.appDependencyProbingPaths -saveToPath $destinationPath  | Where-Object { $_ }
+    $settings = $env:Settings | ConvertFrom-Json | ConvertTo-HashTable -recurse
+    $settings = AnalyzeRepo -settings $settings -token $token -baseFolder $baseFolder -project $project -doNotCheckArtifactSetting
```
mazhelez: If `AnalyzeRepo` is needed here only to determine the `appDependencyProbingPaths`'s value, consider simply creating a dedicated function to do that.

Actions/DownloadProjectDependencies/DownloadProjectDependencies.Action.ps1
```diff
@@ -9,16 +9,12 @@ Param(
...
+    if ($settings.ContainsKey('appDependencyProbingPaths') -and $settings.appDependencyProbingPaths) {
+        return Get-Dependencies -probingPathsJson $settings.appDependencyProbingPaths -saveToPath $destinationPath | Where-Object { $_ }
     }
-
-    return $downloadedDependencies
+    return $settings
```
mazhelez: So, the function returns either the settings or the result from `Get-Dependencies`? I don't get why...

mazhelez (commented):
Posting a half-review as the PRs keeps getting updated.

Actions/AddExistingApp/AddExistingApp.ps1
freddydk: Before this PR - there was no try/catch in the .ps1 - only in the yaml - and look at the run in the link - it gives an error in the .ps1 but the action succeeds.
The try/catch in the yaml will only catch if there is an error during the call (like the script doesn't exist)

Actions/AL-Go-Helper.ps1
```diff
@@ -958,93 +960,101 @@ function AnalyzeRepo {
...
+                        $_
+                    }
+                    else {
+                        New-Object -Type PSObject -Property $_
+                    }
+                })
+            $settings.appDependencyProbingPaths | ForEach-Object {
+                $dependency = $_
+                if (-not ($dependency.PsObject.Properties.name -eq "repo")) {
+                    throw "AppDependencyProbingPaths needs to contain a repo property, pointing to the repository on which you have a dependency"
```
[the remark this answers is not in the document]
freddydk: It's the project that has a dependency

Actions/DownloadProjectDependencies/README.md
freddydk: Nope - the only action setting env.Settings is ReadSettings.

Actions/DetermineArtifactUrl/DetermineArtifactUrl.ps1
freddydk: Yes, but AnalyzeRepo will also download the artifacts - which it shouldn't.
All-in-all - AnalyzeRepo needs to go away and have functions like Determine-ArtifactUrl to take over

Actions/DownloadProjectDependencies/DownloadProjectDependencies.Action.ps1
freddydk: Yep - AnalyzeRepo should be split up - I just thought that this PR is big enough to include this refactoring in that (as stated in the description)

Actions/DownloadProjectDependencies/DownloadProjectDependencies.Action.ps1
freddydk: Bug - I was fixing that while you were reviewing - sorry

Templates/AppSource App/.github/workflows/_BuildALGoProject.yaml
```diff
@@ -128,8 +127,11 @@ jobs:
           with:
             shell: ${{ inputs.shell }}
             parentTelemetryScopeJson: ${{ inputs.parentTelemetryScopeJson }}
+            artifact: ${{ env.artifact }}
+            appFolders: ${{ env.artifact }}
```
freddydk, line 131: should be env.appFolders

Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -116,8 +124,28 @@ try {
         }
     }
 
-    $repo = AnalyzeRepo -settings $settings -token $token -baseFolder $baseFolder -project $project -insiderSasToken $insiderSasToken @analyzeRepoParams
-    if ((-not $repo.appFolders) -and (-not $repo.testFolders)) {
+    $repo = AnalyzeRepo -settings $settings -token $token -baseFolder $baseFolder -project $project -insiderSasToken $insiderSasToken -doNotCheckAppDependencyProbingPaths @analyzeRepoParams
+
+    if ($appFolders) {
```
mazhelez, line 129: Absolutely no idea why all this is needed.

So, there's `$appFolders` parameter , there's `$settings.appFolders` and `$repo.appFolders`. At least one isn't needed. Ideally, all of them should be the same, shouldn't they? If not, why?

mazhelez (commented):
Some of the work around `appFolders` seems too convoluted.

Actions/DownloadProjectDependencies/DownloadProjectDependencies.Action.ps1
```diff
@@ -84,45 +78,59 @@ function DownloadDependenciesFromCurrentBuild($baseFolder, $project, $projectsDe
...
+    Add-Content -Encoding UTF8 -Path $env:GITHUB_OUTPUT -Value "DownloadedApps=$DownloadedAppsJson"
+    Add-Content -Encoding UTF8 -Path $env:GITHUB_OUTPUT -Value "DownloadedTestApps=$DownloadedTestAppsJson"
 
-Add-Content -Encoding UTF8 -Path $env:GITHUB_OUTPUT -Value "DownloadedApps=$DownloadedAppsJson"
-Add-Content -Encoding UTF8 -Path $env:GITHUB_OUTPUT -Value "DownloadedTestApps=$DownloadedTestAppsJson"
+    'appFolders', 'testFolders', 'bcptTestFolders' | ForEach-Object {
+        $propName = $_
+        Write-Host "Setting $($propName) to $($settings."$propName" -join ', ')"
+        $foldersJson = ConvertTo-Json $settings."$propName" -Depth 99 -Compress
+        Add-Content -Encoding UTF8 -Path $env:GITHUB_ENV -Value "$propName=$foldersJson"
```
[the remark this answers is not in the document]
freddydk: Because appDependencyProbingPaths supports 'include' - meaning that you can include apps from other projects in a project.
So, artifact is read in ReadSettings and placed as ENV var - potentially overridden in DetermineArtifactUrl
appFolders, testFolders and bcptTestFolders are read in ReadSettings and potentially overridden in DownloadProjectDependencies.

If we don't do this, then we cannot resolve appDependencyProbingPaths in the action (or only part of it at least)

Actions/AddExistingApp/AddExistingApp.ps1
freddydk: This comment is removed in the next checkin

Actions/RunPipeline/RunPipeline.ps1
freddydk: appFolders is the parameter for the action, which is mandatory false.
it defaults to repo.appFolders.
I will make it mandatory = true to simplify this

Actions/ReadSecrets/ReadSecrets.ps1
```diff
@@ -116,19 +121,16 @@ try {
     #region Action: Output
 
     $outSecretsJson = $outSecrets | ConvertTo-Json -Compress
-    Add-Content -Encoding UTF8 -Path $env:GITHUB_ENV -Value "RepoSecrets=$outSecretsJson"
-
-    $outSettingsJson = $outSettings | ConvertTo-Json -Depth 99 -Compress
-    Add-Content -Encoding UTF8 -Path $env:GITHUB_ENV -Value "Settings=$OutSettingsJson"
+    Add-Content -Encoding UTF8 -Path $env:GITHUB_ENV -Value "Secrets=$outSecretsJson"
```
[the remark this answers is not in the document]
freddydk: Correct, will remove

Actions/DownloadProjectDependencies/DownloadProjectDependencies.Action.ps1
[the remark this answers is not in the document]
freddydk: Let me know if you have other ideas - I don't think that partially solving appDependencyProbingPaths is the right way - then we should move all back to RunPipeline, which also doesn't seem right

Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -116,8 +120,14 @@ try {
         }
     }
 
-    $repo = AnalyzeRepo -settings $settings -token $token -baseFolder $baseFolder -project $project -insiderSasToken $insiderSasToken @analyzeRepoParams
-    if ((-not $repo.appFolders) -and (-not $repo.testFolders)) {
+    $repo = AnalyzeRepo -settings $settings -token $token -baseFolder $baseFolder -project $project -insiderSasToken $insiderSasToken -doNotCheckAppDependencyProbingPaths @analyzeRepoParams
+
+    # Using this construct as there are differences in how PS5 and PS7 interpret simple JSON like [] or ['test']
+    $repo.appFolders = (ConvertFrom-Json "{""folders"":$appFolders}").folders
```
mazhelez, line 126: Doesn't the following work?:

```
$repo.appFolders = @(ConvertFrom-Json $appFolders)
```

Templates/Per Tenant Extension/.github/workflows/_BuildALGoProject.yaml
```diff
@@ -83,17 +83,17 @@ jobs:
             shell: ${{ inputs.shell }}
             parentTelemetryScopeJson: ${{ inputs.parentTelemetryScopeJson }}
             project: ${{ inputs.project }}
-            get: useCompilerFolder,keyVaultCodesignCertificateName,doNotSignApps
+            get: useCompilerFolder,keyVaultCodesignCertificateName,doNotSignApps,artifact,appFolders,testFolders,bcptTestFolders
```
aholstrup1, line 86: Could we make the name 'artifact' a bit more descriptive? From your PR it looks like it has something to do with the artifact url so maybe "artifacturl"?

Actions/DownloadProjectDependencies/DownloadProjectDependencies.Action.ps1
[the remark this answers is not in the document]
freddydk: Then ReadSettings becomes a huge function.

Here's how it works now:
ReadSettings should read the settings and settings.artifact should contain the value of artifact given in settings for this branch, this workflow etc. With the get parameter, we request to have artifact set as environment variable as well.
Then DetermineArtifactUrl will decipher the artifact setting and set env:artifact to the actual URL to use including the insider sas token. (in the old days this action would re-write env:settings)
RunPipeline then gets env:artifact transferred as parameter.

If we were to resolve artifact in ReadSettings we would spend time on querying artifacts store etc. in order to get the right artifactURL - ReadSettings would also need to get secrets and getting secrets needs to get settings.

The same pattern is implemented for appFolders, which are affected by appDependencyProbingPaths, which sometimes needs secrets.

Enumerating appDependencyProbingPaths while reading settings seems wrong.

So now, env:appFolders are the folders that needs to be built and settings.appFolders are the folders you have in settings  - always.

Actions/RunPipeline/RunPipeline.ps1
freddydk: Nope... - you can try to run this:
```
# Works in PS5, not in PS7
$test = ConvertFrom-Json -InputObject '["test"]'
$test.GetType()
$test.Count
$test[0]

# Works in PS5 and PS7
$test = ConvertFrom-Json -InputObject '["test","test2"]'
$test.GetType()
$test.Count
$test[0]

# works in PS7 and in PS5
$test = @(ConvertFrom-Json -InputObject '["test"]')
$test.GetType()
$test.Count
$test[0]

# works in PS7, not in PS5
$test = @(ConvertFrom-Json '["test","test2"]')
$test.GetType()
$test.Count
$test[0]

# works always
$test = (ConvertFrom-Json -InputObject '{"folders":["test","test2"]}').folders
$test.GetType()
$test.Count
$test[0]
```

Actions/ReadSecrets/ReadSecrets.ps1
```diff
@@ -1,24 +1,20 @@
 Param(
+    [Parameter(HelpMessage = "All GitHub Secrets in compressed JSON format", Mandatory = $true)]
+    [string] $gitHubSecrets = "",
     [Parameter(HelpMessage = "Comma separated list of Secrets to get", Mandatory = $true)]
-    [string] $secrets = "",
+    [string] $getSecrets = "",
+    [Parameter(HelpMessage = "Specify whether or not the function should also get AuthToken secrets from AppDependencyProbingPaths", Mandatory = $false)]
+    [bool] $getAppDependencyProbingPathsSecrets,
```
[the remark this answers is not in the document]
freddydk: I like that approach

Actions/DetermineArtifactUrl/DetermineArtifactUrl.ps1
```diff
@@ -22,40 +17,30 @@ try {
...
-    if ($secrets.ContainsKey('insiderSasToken')) {
-        $insiderSasToken = [System.Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String($secrets.insiderSasToken))
-    } else {
-        $insiderSasToken = ""
-    }
-    $projectSettings = $env:Settings | ConvertFrom-Json | ConvertTo-HashTable
-    $projectSettings = AnalyzeRepo -settings $projectSettings -project $project -doNotCheckArtifactSetting -doNotIssueWarnings
-    $artifactUrl = Determine-ArtifactUrl -projectSettings $projectSettings -insiderSasToken $insiderSasToken
+    $secrets = $env:Secrets | ConvertFrom-Json | ConvertTo-HashTable
+    $insiderSasToken = [System.Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String($Secrets.insiderSasToken))
```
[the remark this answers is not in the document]
freddydk: All requested secrets (getSecrets) are available in env:Secrets - empty if not present as secrets.

Actions/AL-Go-Helper.ps1
```diff
@@ -958,93 +960,101 @@ function AnalyzeRepo {
...
-            if (-not ($dependency.PsObject.Properties.name -eq "branch")) {
-                $dependency | Add-Member -name "branch" -MemberType NoteProperty -Value "main"
-            }
-            Write-Host "Dependency to projects '$($dependency.projects)' in $($dependency.Repo)@$($dependency.branch), version $($dependency.version), release status $($dependency.release_status)"
-            if (-not ($dependency.PsObject.Properties.name -eq "AuthTokenSecret")) {
-                if ($token) {
-                    Write-Host "Using token as AuthTokenSecret"
+    if (!$doNotCheckAppDependencyProbingPaths) {
+        Write-Host "Checking appDependencyProbingPaths"
+        if ($settings.appDependencyProbingPaths) {
```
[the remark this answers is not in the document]
mazhelez: The property `appDependencyProbingPaths` is always there because of the default settings.

Actions/DetermineArtifactUrl/DetermineArtifactUrl.ps1
```diff
@@ -22,40 +17,30 @@ try {
...
-    if ($secrets.ContainsKey('insiderSasToken')) {
-        $insiderSasToken = [System.Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String($secrets.insiderSasToken))
-    } else {
-        $insiderSasToken = ""
-    }
-    $projectSettings = $env:Settings | ConvertFrom-Json | ConvertTo-HashTable
-    $projectSettings = AnalyzeRepo -settings $projectSettings -project $project -doNotCheckArtifactSetting -doNotIssueWarnings
-    $artifactUrl = Determine-ArtifactUrl -projectSettings $projectSettings -insiderSasToken $insiderSasToken
+    $secrets = $env:Secrets | ConvertFrom-Json | ConvertTo-HashTable
+    $insiderSasToken = [System.Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String($secrets.insiderSasToken))
```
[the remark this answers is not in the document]
mazhelez: > All requested secrets (getSecrets) are available in env:Secrets - empty if not present as secrets.

(the response for a similar comment)

Actions/AL-Go-Helper.ps1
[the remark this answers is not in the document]
freddydk: If it exists in the AL-Go-Helper settings structure - then it will always exist as a property on the settings variable.

Actions/DetermineArtifactUrl/README.md
```diff
@@ -7,19 +7,19 @@ Determines the artifactUrl to use for a given project
 | Name | Description |
 | :-- | :-- |
 | Settings | env.Settings must be set by a prior call to the ReadSettings Action |
+| Secrets | env.Secrets with insiderSasToken must be read by a prior call to the ReadSecrets Action |
```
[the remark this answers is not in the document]
freddydk: insiderSasToken can be missing, in which case env:Secrets.insiderSasToken is empty - and it will work with bcartifacts.
But the action requires that you have called ReadSecrets with insiderSasToken for Secrets.insiderSasToken to exist.

Actions/DetermineArtifactUrl/DetermineArtifactUrl.ps1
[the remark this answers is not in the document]
freddydk: Nope - the secrets provided in getSecrets are created as properties in env:secrets (possibly empty though)

mazhelez (approved):
There're still some low-hanging issues.

Actions/DownloadProjectDependencies/DownloadProjectDependencies.Action.ps1
```diff
@@ -9,16 +9,11 @@ Param(
...
 function DownloadDependenciesFromProbingPaths($baseFolder, $project, $destinationPath) {
-    $projectSettings = ReadSettings -baseFolder $baseFolder -project $project
-    
-    $probingPaths = $projectSettings.appDependencyProbingPaths | ConvertFrom-Json
-
-    $downloadedDependencies = @()
-    if ($probingPaths) {
-        $downloadedDependencies += Get-Dependencies -probingPathsJson $repo.appDependencyProbingPaths -saveToPath $destinationPath  | Where-Object { $_ }
+    $settings = $env:Settings | ConvertFrom-Json | ConvertTo-HashTable -recurse
+    AnalyzeRepo -settings $settings -token $token -baseFolder $baseFolder -project $project -doNotCheckArtifactSetting -doNotIssueWarnings | Out-Null
```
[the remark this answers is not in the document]
freddydk: You are right (but the | out-null should also go away)

Actions/AL-Go-Helper.ps1
```diff
@@ -958,93 +960,101 @@ function AnalyzeRepo {
...
+                            }
+                            else {
+                                $depProject = $_
+                                Write-Host "Identified dependency to project $depProject in the same repository"
+
+                                $dependencyIds = @( @($settings.appDependencies + $settings.testDependencies) | ForEach-Object { $_.id })
+                                $depSettings = ReadSettings -baseFolder $baseFolder -project $depProject -workflowName "CI/CD"
+                                $depSettings = AnalyzeRepo -settings $depSettings -token $token -baseFolder $baseFolder -project $depProject -includeOnlyAppIds @($dependencyIds + $includeOnlyAppIds + $dependency.alwaysIncludeApps) -doNotIssueWarnings -doNotCheckArtifactSetting -server_url $server_url -repository $repository
+
+                                Set-Location $projectPath
```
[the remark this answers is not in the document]
freddydk: fixed

mazhelez (approved):

Issue #653, opened 2023-08-10
In preview: Using an HSM Code Signing Cert, CI/CD fails


BMI-Mike:
This is due to a missing mandatory settings.json parameter. 

In the Sign step of CI/CD I receive this error:

```
Secrets: {"licenseFileUrl":"***","insiderSasToken":"***","codeSignCertificateUrl":"","codeSignCertificatePassword":"","keyVaultCertificateUrl":"","keyVaultCertificatePassword":"","keyVaultClientId":"","gitHubPackagesContext":"","applicationInsightsConnectionString":""}
    artifactCacheKey: 
    containerName: bc5822555729
    _azureCredentialsJson: ***
    _timestampService: http://timestamp.digicert.com
    _digestAlgorithm: SHA256
    _pathToFiles: ./.buildartifacts/Apps/*.app
    _parentTelemetryScopeJson: 
Error: Unexpected error when running action. Error Message: Cannot process command because of one or more missing mandatory parameters: settingsJson., StackTrace: at <ScriptBlock>, D:\a\_temp\197b8bd8-df1e-[41](https://github.com/BMI-Software/OfficeProducts-AppSource/actions/runs/5822555729/job/15793900820#step:9:42)94-a21f-3f81[44](https://github.com/BMI-Software/OfficeProducts-AppSource/actions/runs/5822555729/job/15793900820#step:9:45)8729df.ps1: line 4 <- at <ScriptBlock>, <No file>: line 1
Error: Process completed with exit code 1.
```

Since I am using an HSM Cert, providing a proper keyVaultCodesignCertificateName value in .AL-Go/settings.json should be all that is required.

This was working just fine for several weeks.  All of our projects began to fail simultaneously, starting roughly 20 hours ago.

I'm not sure if this is a duplicate of #652

freddydk:
What version of AL-Go are you running?
preview?
Did you upgrade AL-Go system Files ~20 hours ago?

freddydk:
For now you have two options really
1. Revert the PR which updated the AL-Go System files ~20 hours ago
2. Wait for the PR I created above to be merged in and deployed to preview - will happen tomorrow morning european timezone.

BMI-Mike:
> What version of AL-Go are you running? **AL-Go for GitHub,d15c25e25c4b0033da9f99970e3a420e231f7fbc**
preview? **Yes**
Did you upgrade AL-Go system Files ~20 hours ago? **Yes**

[...]

PR #654, opened 2023-08-10
Issue #653

freddydk:
Bug in preview, Sign cannot run due to a mandatory parameter not being transferred.


aholstrup1 (approved):

Actions/Sign/Sign.ps1
```diff
@@ -50,14 +46,14 @@ try {
         Register-NavSip 
         Write-Host "::endgroup::"
 
-        AzureSignTool sign --file-digest $FileDigest `
+        AzureSignTool sign --file-digest $digestAlgorithm `
```
mazhelez, line 49: Why using the same value for `--file-digest` and `--timestamp-digest`?

mazhelez: Oh, because we don't expose both parameters in the action.

mazhelez (approved):

PR #668, opened 2023-08-17
Add fullBuildPatterns setting

mazhelez:
- New setting (_fullBuildPatterns_) to specify changes in which files and folders would trigger a full build (building all AL-Go projects).
- _DetermineProjectsToBuild_ action now uses `$env:Settings`
- Refactor DetermineArtifactUrl.Test.ps1 to separate different test cases


Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.ps1
```diff
@@ -114,44 +133,44 @@ function Get-ProjectsToBuild(
...
 {
     Write-Host "Determining projects to build in $baseFolder"
-    
+
     Push-Location $baseFolder
 
     try {
-        $settings = ReadSettings -baseFolder $baseFolder -project '.' # Read AL-Go settings for the repo
-        
+        $settings = $env:Settings | ConvertFrom-Json
```
freddydk, line 143: This makes settings a PSCustomObject (earlier it was a hashtable) - is that a problem?

Tests/DetermineProjectsToBuild.Test.ps1
```diff
@@ -585,14 +687,14 @@ Describe "Get-ProjectsToBuild" {
...
+
         $dependantAppFile = @{ id = '83fb8305-4079-415d-a25d-8132f0436fd2'; name = 'Second App'; publisher = 'Contoso'; version = '1.0.0.0'; dependencies = @(@{id = '83fb8305-4079-415d-a25d-8132f0436fd1'; name = 'First App'; publisher = 'Contoso'; version = '1.0.0.0'} ) }
         New-Item -Path "$baseFolder/Project2/.AL-Go/settings.json" -type File -Force
         New-Item -Path "$baseFolder/Project2/app/app.json" -Value (ConvertTo-Json $dependantAppFile -Depth 10) -type File -Force
 
         #Add settings file
-        $alGoSettings = @{ useProjectDependencies = $true }
-        New-Item -Path "$baseFolder/.github/AL-Go-Settings.json" -Value (ConvertTo-Json $alGoSettings) -type File -Force
+        $alGoSettings = @{ alwaysBuildAllProjects = $false; projects = @(); useProjectDependencies = $true }
+        $env:Settings = ConvertTo-Json $alGoSettings -Depth 99
```
freddydk, line 697: -compress ? (same for all)

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.ps1
mazhelez: I don't think it is. Any specific scenario to look for? Settings always contain the default value, right?

freddydk: HashTable:
.containsKey or .Keys + used as parameter in a number of functions in AL-Go expecting hashtables.
PSCustomObject:
.PSObject.Properties.name -eq 'x'

freddydk: You are not using any of those, but is it a problem that in some actions we work with settings as a hashtable and in other actions we work with it as PSCustomObject? Could maybe lead to misunderstandings down the road?
(we might have the same issue in other places - so OK with me)

freddydk (approved):

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.ps1
```diff
@@ -50,39 +50,58 @@ function Get-FilteredProjectsToBuild($settings, $projects, $baseFolder, $modifie
...
-    
+    Write-Host "$($modifiedFiles.Count) modified file(s): $($modifiedFiles -join ', ')"
+
     if ($modifiedFiles.Count -ge 250) {
         Write-Host "More than 250 files modified, building all projects"
         return $projects
     }
 
-    Write-Host "$($modifiedFiles.Count) modified file(s): $($modifiedFiles -join ', ')"
+    $fullBuildPatterns = @( Join-Path '.github' '*.json')
```
aholstrup1, line 67: Could this be the added as the default value for the fullBuildPatterns setting instead? I think if you decide to use this setting it should use only the build patterns you specify

aholstrup1 (approved):

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.ps1
freddydk: IMO The problem of adding this as a default would be that people would have to remember to include it if they create the setting - I think the current mechanism is correct - we have some settings, which people cannot remove.

aholstrup1: A little documentation on the setting would reduce that risk :) I just don't love how we inject settings like this or have settings that people cannot remove. 

Anyway.. It's not something I feel strong enough about to block the PR. It's up to you Maria. Feel free to dismiss my comment if you disagree.

mazhelez: My thinking in this case was what Freddy described: it's very easy to forget to add `.github/*.json` and this is a _mandatory_ check (as changing the AL-Go settings should re-build all projects).

One other mandatory file check would be _PullRequestHandler.yaml_ but only of the trigger is `pull_request`.

PR #671, opened 2023-08-21
Add required status check job to Pull Request Build

aholstrup1:
Add a job to the Pull Request Build that always fails if there are other failing jobs in the run. This job is intended to be used as a required status check on PRs. See for example: 
https://github.com/aholstrup1/ALAppExtensions/pull/161


Actions/PullRequestStatusCheck/action.yaml
```diff
@@ -0,0 +1,31 @@
+name: PowerShell script
```
mazhelez, line 1:
```suggestion
name: Pull Request Status Check
```

Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
```diff
@@ -103,21 +98,14 @@ jobs:
...
     steps:
-      - name: Checkout
-        uses: actions/checkout@v3
-        with:
-          lfs: true
-          ref: refs/pull/${{ github.event.number }}/merge
-
-      - name: Finalize the workflow
-        id: PostProcess
-        uses: microsoft/AL-Go-Actions/WorkflowPostProcess@main
```
mazhelez, line 119: Is `WorkflowPostProcess` not needed at all?

Actions/PullRequestStatusCheck/PullRequestStatusCheck.ps1
```diff
@@ -0,0 +1,17 @@
...
+    [string] $RunId
+)
+
+Write-Host "Checking workflow status for run $RunId in repository $Repository"
+
+$workflowJobs = gh api /repos/$Repository/actions/runs/$RunId/jobs -H "Accept: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28" | ConvertFrom-Json
+$failedJobs = $workflowJobs.jobs | Where-Object { $_.conclusion -eq "failure" }
+
+if ($failedJobs) {
+    throw "Workflow failed with the following jobs: $($failedJobs.name -join ', ')"
```
mazhelez, line 14: Any AL-Go telemetry around here?

Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
```diff
@@ -103,21 +98,14 @@ jobs:
       secrets: 'licenseFileUrl,insiderSasToken,keyVaultCertificateUrl,keyVaultCertificatePassword,keyVaultClientId,gitHubPackagesContext,applicationInsightsConnectionString'
       publishThisBuildArtifacts: ${{ needs.Initialization.outputs.workflowDepth > 1 }}
 
-  PostProcess:
+  StatusCheck:
     runs-on: [ windows-latest ]
     needs: [ Initialization, Build ]
-    if: (!cancelled())
+    if: always()
```
mazhelez, line 104: What if the workflow run is canceled?

Then _StatusCheck_ would succeed, which isn't quite accurate.

Actions/PullRequestStatusCheck/action.yaml
```diff
@@ -0,0 +1,31 @@
...
+    required: false
+    default: ${{ github.token }}
+
+runs:
+  using: composite
+  steps:
+    - name: run
+      shell: ${{ inputs.shell }}
+      env:
+        GITHUB_TOKEN: ${{ inputs.token }}
```
mazhelez, line 19: Can't we just use `${{ github.token }}` here? What's the point of having it as an input?

```suggestion
        GITHUB_TOKEN: ${{ github.token }}
```

mazhelez (commented):
I imagine this change would affect E2E tests.

If not, we are missing E2E tests that capture the output of a PR Build.

Actions/PullRequestStatusCheck/action.yaml
aholstrup1: I think we can. I was under the impression that GitHub actions could only access the Github_Token if it is explicitly passed. That doesn't seem to be the case https://docs.github.com/en/actions/security-guides/automatic-token-authentication :)

Actions/PullRequestStatusCheck/PullRequestStatusCheck.ps1
mazhelez: Ahh, yes, true.

Never mind, then. I thought it'll be nice to track PRs Builds success rate.

Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
```diff
@@ -103,21 +98,16 @@ jobs:
       secrets: 'licenseFileUrl,insiderSasToken,keyVaultCertificateUrl,keyVaultCertificatePassword,keyVaultClientId,gitHubPackagesContext,applicationInsightsConnectionString'
       publishThisBuildArtifacts: ${{ needs.Initialization.outputs.workflowDepth > 1 }}
 
-  PostProcess:
+  StatusCheck:
     runs-on: [ windows-latest ]
     needs: [ Initialization, Build ]
     if: (!cancelled())
```
mazhelez, line 104: Alternatively, the job can _always_ run and check for canceled task.

aholstrup1: I think it'd be better just to skip it in that case. From what I can see, in order for a GitHub status check to be "green" the job has to run and succeed. If the workflow is cancelled and the "Pull Request Status Check" task is skipped, then status check won't be green. 

Doesn't this give us the behavior we want?

Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
mazhelez: You can leave it for now. The new job (_StatusCheck_) doesn't need to depend on it.
We can have a separate effort to determine how to deal with it.

mazhelez (dismissed):

Actions/PullRequestStatusCheck/action.yaml
```diff
@@ -0,0 +1,35 @@
+name: Pull Request Status Check
+author: Microsoft Corporation
+inputs:
+  shell:
+    description: Shell in which you want to run the action (powershell or pwsh)
+    required: false
+    default: powershell
+  repository:
+    description: Repository name
```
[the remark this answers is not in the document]
aholstrup1: Yeah, that was why I added it. I also kinda liked that you could run the script locally (without the ENV variables defined by GitHub). I see your point though, and I have refactored accordingly :)

mazhelez (approved):

mazhelez (approved):

freddydk (approved):

PR #676, opened 2023-08-25
[Draft][Suggestion] Throw warning or error about obsoleted settings and secrets

aholstrup1:
Suggestion:

When we obsolete an Al-Go setting, we should make sure to write a warning / error into the GitHub workflow, so the repository owner is aware that they are using an obsolete setting. Right now, we're doing this within the different Al-Go actions but for the future it could be beneficial to keep it all in one place by validating the settings. 

See [Example run](https://github.com/aholstrup1/ALAppExtensions/actions/runs/5976065375)


Templates/AppSource App/.github/workflows/CICD.yaml
```diff
@@ -59,6 +59,7 @@ jobs:
           parentTelemetryScopeJson: ${{ steps.init.outputs.telemetryScopeJson }}
           getEnvironments: '*'
           get: type
+          validateSettings: 'Y'
```
freddydk, line 62: Should we do this in CheckForUpdates instead of inside CI/CD?

Actions/ReadSecrets/ReadSecretsHelper.psm1
```diff
@@ -220,3 +220,43 @@ function GetSecret {
...
+            PrintObsoleteMessage -Setting $_ -Message $obsoletedSetting.message -Type $obsoletedSetting.Type
+            if ($obsoletedSetting.Type -eq "Error") {
+                $foundError = $true
+            }
+        }
+    }
+
+    # Throw an error if any obsolete settings were found
+    if ($foundError) {
+        throw "Obsolete settings found. Please update your Al-Go settings."
```
freddydk, line 252: Obsolete secrets?

Actions/ReadSecrets/ReadSecretsHelper.psm1
```diff
@@ -220,3 +220,43 @@ function GetSecret {
...
+    }
+}
+
+function PrintObsoleteMessage($SettingName, $Message, $Type) {
+    if ($Type -eq "Warning") {
+        Write-Host "::Warning::The Al-Go setting $SettingName has been obsoleted. $Message This warning will become an error in a future release."
```
freddydk, line 258: secret?

PR #687, opened 2023-09-04
[SUGGESTION] Temporary Access Token

freddydk:
This PR removes the need for having a GhTokenWorkflow secret created in your repository.
In all workflows, where the GhTokenWorkflow is needed, you can specify a temporary access token in the Workflow Dispatch window:
image

In ReadSecrets, the temporarily specified token is used as GhTokenWorkflow (instead of reading the secret):
image

The value for the specified token is masked in the workflow before being used, in order to avoid leaking the secret.

You can try out the functionality using this template: https://github.com/freddydk/AL-Go@temptoken


freddydk:
Not doing this - will solve the issue another way

PR #692, opened 2023-09-06
Deliver Action - "can not find the container" fix

akoniecki:
_New-AzStorageContainer_ implemented into Deliver action.

- Added logic to verify the existence of the specified Azure Storage container.
- If the container doesn't exist, it will automatically be created. 
- This ensures smooth artifact delivery without interruption due to missing storage containers.

Currently, if storage container does not exists, the Deliver job will fail with an error:
Unexpected error when running action. Can not find the container '<containerName>'.,
D:\a\_actions\microsoft\AL-Go-Actions\v3.2\Deliver\Deliver.ps1: line 363
D:\a\_actions\microsoft\AL-Go-Actions\v3.2\Deliver\Deliver.ps1: line 95 
Error: Process completed with exit code 1.


freddydk:
There are really two reasons why we haven't implemented this.
1. Creating a new container usually happens very seldomly - not with every build. When you setup the connection, you also create the container.
2. What security rules should apply to the container? (public/private/permissions)
3. A simple settings bug can cause multiple blob containers to be created 

Will discuss your proposal with the other team members here, but I lean towards not doing this.

akoniecki:
Sure, I can understand risks you mention. Just want to comment that from my practice, I found it time consuming and little annoying to create storage container for every repository setup. Especially during migration when there are many repositories to handle.

mazhelez:
It is a valid request, but it is not right to change the default behavior.

"Create container if it doesn't exist" is a potential setting. Maybe, similar to `DeployTo<environmentname>` there could be a `DeliverTo<deliveryTarget>` setting.

Another way this custom behavior can be achieved is to add your own `.github/DeliverToStorage.ps1`.

freddydk:
There actually is a `DeliverTo<deliveryTarget>` setting - which today only includes Branches (not documented though - will do that).
The setting could be added inside that.

mazhelez:
> There actually is a `DeliverTo<deliveryTarget>` setting - which today only includes Branches (not documented though - will do that). The setting could be added inside that.

I knew I've seen it somewhere!

@akoniecki how about adding support for such setting?

akoniecki:
Support was added for the following setting:
`"DeployToStorage": {
    "createContainerIfNotExist":  true
}`

akoniecki:
Added support for the following setting:
`"DeliverToStorage": {
    "createContainerIfNotExist": true
}`

freddydk:
I have updated the PR with release notes and added `DeliverTo<deliveryTarget>` to settings.md in the PR as well.

freddydk (dismissed):

mazhelez (approved):

mazhelez (approved):

PR #722, opened 2023-09-22
use companyName setting on UploadImportAndApplyConfigPackageInBc

tyurm:
CompanyName parameter defines the company where tests should be run, but it does not affect test data import.
I suggest adding this modification so companyName setup works for both steps.


freddydk:
This is a good fix, thanks

freddydk (approved):

PR #728, opened 2023-09-26
Removal of Insider SAS Token

freddydk:
This PR will remove all references to the Insider SAS Token.
Currently, this uses a local version of BcContainerHelper - needs to be replaced with preview or latest when this ships.


Actions/DetermineArtifactUrl/DetermineArtifactUrl.ps1
```diff
@@ -17,16 +17,9 @@ try {
     #region Action: Determine artifacts to use
     $telemetryScope = CreateScope -eventId 'DO0084' -parentTelemetryScopeJson $parentTelemetryScopeJson
 
-    $insiderSasToken = ""
```
mazhelez, line 20: In _BuildALGoProject.yaml there's
```
env:
```
when calling _DetermineArtifactUrl_ action. 
It should be removed.

mazhelez (changes requested):

Actions/DetermineArtifactUrl/DetermineArtifactUrl.ps1
freddydk: Good catch

mazhelez (approved):

aholstrup1 (approved):

PR #751, opened 2023-10-02
Environment URL

freddydk:
This PR will add the URL of the environment being deployed to underneath the Environment name in the Build summary, like:
image

It is implemented in the Publish To Environment workflow and for Continuous Deployment in the CI/CD workflow.
image

and also supports custom deployment.

### DIFF
--- a/Actions/Deploy/action.yaml
+++ b/Actions/Deploy/action.yaml
@@ -34,8 +34,8 @@ runs:
   using: composite
   steps:
     - name: run
-      id: Deploy
       shell: ${{ inputs.shell }}
+      id: Deploy
       env:
         _token: ${{ inputs.token }}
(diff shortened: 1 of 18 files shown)


Actions/Deploy/action.yaml
```diff
@@ -26,11 +26,16 @@ inputs:
   deploymentEnvironmentsJson:
     description: The settings for all Deployment Environments
     required: true
+outputs:
```
mazhelez, line 29: Maybe also some doc in the action's README?

mazhelez (approved):

Actions/Deploy/action.yaml
freddydk: done

PR #752, opened 2023-10-03
Only display test result summary when there is something to display 

aholstrup1:
GitHub will only show the first three build summaries and showing "Test Results Not Found" three times doesn't provide much value IMO. 

Only display test result summary when there is something to display

image


freddydk (approved):
LGTM

PR #756, opened 2023-10-04
Fix artifacts names for CalculateArtifactNames and DownloadProjectDependencies

mazhelez:
_**Issue 1**_:
_CalculateArtifactNames_ uses only `github.ref_name` as branch name when constructing artifacts names. However, in the case of PR build, the value is the merge branch, e.g., "123/merge".
Solution:
_**Solution**_: Remove the _branch_ input from _CalculateArtifactNames_ action and let it determine the proper branch on the fly, based on `$env:GITHUB_HEAD_REF` ([doc](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)).

_**Issue 2**_:
_DownloadProjectDependencies_ uses the wrong branches as head and target branches when running from a PR Build.
Solution:
_**Solution**_: Utilize `$env:GITHUB_HEAD_REF` and `GITHUB_BASE_REF` ([doc](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)).


Actions/DownloadProjectDependencies/DownloadProjectDependencies.Action.ps1
```diff
@@ -40,12 +40,16 @@ function DownloadDependenciesFromCurrentBuild($baseFolder, $project, $projectsDe
...
+            $headBranch = $ENV:GITHUB_REF_NAME
+        }
 
-        $baseBranch = $ENV:GITHUB_BASE_REF_NAME
-        # $ENV:GITHUB_BASE_REF_NAME is specified only for pull requests, so if it is not specified, use the current branch
+        $baseBranch = $ENV:GITHUB_BASE_REF
+        # $ENV:GITHUB_BASE_REF is specified only for pull requests, so if it is not specified, use GITHUB_REF_NAME
         if (!$baseBranch) {
-            $baseBranch = $currentBranch
+            $baseBranch = $GITHUB_REF_NAME
```
freddydk, line 52: Shouldn't this be $ENV:GITHUB_REF_NAME ?   (also in the comment)

freddydk (changes requested):

Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
```diff
@@ -20,6 +18,12 @@ if ($project -eq ".") {
     $project = $settings.repoName
 }
 
+$branchName = $ENV:GITHUB_HEAD_REF
```
aholstrup1, line 21: Wouldn't `git branch --show-current` also do the trick? We're just trying to get the current branch we're on?

Actions/DownloadProjectDependencies/DownloadProjectDependencies.Action.ps1
mazhelez: Oh, yes, true. My bad.

Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
mazhelez: Yes, but the current branch is something like `refs/pull/123/merge`.

Actions/DownloadProjectDependencies/DownloadProjectDependencies.Action.ps1
mazhelez:
```suggestion
            $baseBranch = $ENV:GITHUB_REF_NAME
```

Actions/CalculateArtifactNames/CalculateArtifactNames.ps1
mazhelez: For the _head_ branch it might not matter that much, but the PR branch seems more appropriate choice.

aholstrup1 (approved):

PR #761, opened 2023-10-09
fix excessive http requests

freddydk:


Actions/Github-Helper.psm1
```diff
@@ -746,7 +751,8 @@ function GetArtifacts {
                 $result += @($projectArtifact)
             }
             else {
-                $allArtifactsFound = $false
+                # If the last artifact read is not from the same workflow run, we have read all artifacts
+                $allArtifactsFound = $allArtifacts[$allArtifacts.Count-1].workflow_run.id -ne $workflowrunid
```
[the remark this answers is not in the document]
freddydk: Yeah - that fix wasn't good:-(
Was a bit more complex.
I re-wrote the logic (split the loop) and added some comments.

Actions/Github-Helper.psm1
```diff
@@ -720,38 +720,74 @@ function GetArtifacts {
...
+            # The total count changed, restart the loop
+            $total_count = 0
+            continue
+        }
+        $matchingArtifacts += @($artifacts.artifacts | Where-Object { !$_.expired -and $_.name -like $artifactPattern })
+        $buildOutputArtifacts += @($artifacts.artifacts | Where-Object { !$_.expired -and $_.name -like $buildOutputPattern })
+        if ($buildOutputArtifacts.Count -gt 0) {
+            # We have matching artifacts.
+            # If the last artifact in the list of artifacts read is not from the same workflow run, there are no more matching artifacts
+            if ($artifacts.artifacts[$artifacts.artifacts.Count-1].workflow_run.id -ne $buildOutputArtifacts[0].workflow_run.id) {
```
mazhelez, line 762: What if `$artifacts.artifacts.Count` is 1, so here we compare an artifact with itself?

Meanwhile `$matchingArtifacts` can still be empty.

Actions/Github-Helper.psm1
```diff
@@ -720,38 +720,74 @@ function GetArtifacts {
...
+                break
+            }
+        }
+        if ($total_count -le $page*$per_page) {
+            # no more pages
```
mazhelez, line 767: That check would work before `$page` is incremented.

Actions/Github-Helper.psm1
```diff
@@ -720,38 +720,74 @@ function GetArtifacts {
...
+            break
+        }
+    }
+    if ($buildOutputArtifacts.Count -eq 0) {
+        Write-Host "No matching buildOutput artifacts found"
```
mazhelez, line 772: They could be on another page...

freddydk: Correct, but the matchingartifacts are added together into one collection - they will be there.
I tested this with a pagesize of 1, 2, 3, 30 and 100 - I get the same apps every time.

Actions/Github-Helper.psm1
freddydk: That is true, good catch.

Actions/Github-Helper.psm1
```diff
@@ -720,38 +720,74 @@ function GetArtifacts {
...
+        if ($artifacts.artifacts.Count -eq 0) {
+            break
+        }
+        if ($total_count -eq 0) {
+            $total_count = $artifacts.total_count
+        }
+        elseif ($total_count -ne $artifacts.total_count) {
+            # The total count changed, restart the loop
+            $total_count = 0
+            continue
```
mazhelez, line 754: Does this mean more artifacts were added since the previous read?

This would probably be fine for a single build job, but isn't it problematic for multiple build job as some come be 15+ minutes apart?
That would mean some build job use different "baseline" that other from the same workflow.

freddydk: No, this means that the count changed in the middle of this loop - another workflow might be pushing artifacts while we are scanning.

PR #802, opened 2023-11-03
Get artifacts for version 'latest' from the last successful CICD run on the corresponding branch 

mazhelez:
_**Issues**_: Currently, the artifacts are fetched through the _artifacts_ API, which means that for version 'latest' some artifacts may be picked from the latest CICD run that succeeded partially, while others — from a previous CICD run that succeeded. This could lead to a lot of issues for both incremental builds (PR builds) and releasing 'latest' version.

_**Solution**_:
When getting artifacts via `GetArtifacts`, use only artifacts from the latest successful CICD run when the version is 'latest'.
CICD is considered successful if the build jobs succeeded.

Note: GitHub API returns the workflow runs ordered by `created_date` in descending order.


Actions/Github-Helper.psm1
```diff
@@ -708,87 +708,234 @@ function Set-JsonContentLF {
...
+            # No more workflow runs, breaking out of the loop
+            break
+        }
+
+        $CICDRuns = @($workflowRuns.workflow_runs | Where-Object { $_.name -eq ' CI/CD' })
+
+        if ($CICDRuns.Count -gt 0) {
+            $lastSuccessfulCICDRun = $CICDRuns[0].id
+            Write-Host "Found last successful CICD run: $($LastSuccessfulCICDRun)"
+            break;
```
aholstrup1, line 750:
```suggestion
            break
```

Actions/Github-Helper.psm1
```diff
@@ -708,87 +708,234 @@ function Set-JsonContentLF {
...
+
+        if($workflowRuns.workflow_runs.Count -eq 0) {
+            # No more workflow runs, breaking out of the loop
+            break
+        }
+
+        $CICDRuns = @($workflowRuns.workflow_runs | Where-Object { $_.name -eq ' CI/CD' })
+
+        if ($CICDRuns.Count -gt 0) {
+            $lastSuccessfulCICDRun = $CICDRuns[0].id
```
aholstrup1, line 748: I guess this assumes that CICDRuns is sorted with the latest CICD run first? I don't see any mention of it being sorted in the documentation though (https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-repository). Shouldn't we be a bit cautious assuming this if it's not actually documented?

aholstrup1 (commented):
Looks good overall I think 👍 
In the future.. I wonder if we should investigate publishing CI/CD artifacts to GitHub packages instead. I'm guessing finding artifacts for 'latest' or for 'specific version' would be significantly easier if the artifacts are in a nuget feed.

Actions/Github-Helper.psm1
mazhelez: Such things are never really documented. In all my testing, the runs were always sorted.

Ensuring sorted result will be too expensive as the function will have to query for _all_ CICD runs every time.

Actions/Github-Helper.psm1
```diff
@@ -708,87 +708,234 @@ function Set-JsonContentLF {
...
+    If the version is 'latest', the artifacts from the last successful CICD run are returned.
+    Otherwise, the artifacts from the CICD run that built the specified project, mask and version are returned.
+#>
+function GetArtifacts {
+    Param(
+        [string] $token,
+        [string] $api_url = $ENV:GITHUB_API_URL,
+        [string] $repository = $ENV:GITHUB_REPOSITORY,
+        [string] $mask = "Apps",
+        [string] $branch = "main",
```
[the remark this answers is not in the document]
mazhelez: Made the parameters mandatory as in the other functions.

Actions/Github-Helper.psm1
```diff
@@ -708,91 +708,240 @@ function Set-JsonContentLF {
...
-        if ($total_count -eq 0) {
-            # First iteration - initialize variables
-            $matchingArtifacts = @()
-            $buildOutputArtifacts = @()
-            $per_page = 100
-            $page_no = 1
-        }
-        $uri = "$api_url/repos/$repository/actions/artifacts?per_page=$($per_page)&page=$($page_no)"
-        Write-Host $uri
+        $uri = "$api_url/repos/$repository/actions/artifacts?per_page=$($per_page)&page=$($page)&name=$escapedArtifactPattern"
```
freddydk, line 791: I don't think the name property support wildcards?
Is this the reason for specifying project[0]?`
Problem is here that the current version of the repo might have different projects than the workflow run you are looking for.
If you have added a project called ABC, you might be looking for project ABC version 1.2.3.4 - which doesn't exist, but project DEF and GHI does

mazhelez: Yes, the name query parameter doesn't threat wild characters as such.

Using _BuildOutput_ would also work but build output artifacts are produced even on failed CICDs, so that's why I went with the first project in the list.

Are you saying that getting version _latest_ for project _ABC_ shouldn't look for the latest CICD run, but for the latest CICD that actually built project _ABC_? I don't think that should be supported now, should it? 🤔

Actions/Github-Helper.psm1
```diff
@@ -708,91 +708,240 @@ function Set-JsonContentLF {
...
+    $projects = @(@($projects.Split(',')) | ForEach-Object { $_.Replace('\','_').Replace('/','_') })
+
+    $CICDrun = 0
+
+    if($version -eq '*') {
+        # For latest version, use the artifacts from the last successful CICD run
+        $CICDrun = FindLatestSuccessfulCICDRun -token $token -api_url $api_url -repository $repository -branch $branch
+    } else {
+        # For specific version, find the corresponding CICD run based on the first project
+        $CICDrun = FindCICDRunForVersion -token $token -api_url $api_url -repository $repository -branch $branch -project $project[0] -mask $mask -version $version
```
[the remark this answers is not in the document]
mazhelez: Using _BuildOutput_ would also work but build output artifacts are produced even on failed CICDs, so that's why I went with the first project in the list.

Actions/Github-Helper.psm1
freddydk: No, this comment is in FindCICDforVersion.
You call this function with projects[0] - which might be a * or a project name which doesn't exist.
This is the reason for not using the name parameter in the existing implementation.

mazhelez: Oh, `$projects` can actually be a list of patterns itself? That I didn't even think of. 😄 

The project name is in every artifact name, so that approach won't work like this.
Back to introducing a <VERSION> artifact? Any other ideas?

mazhelez: Reverted the usage of `FindLatestSuccessfulCICDRunForVersion`.

For now, latest successful CICD run is used only when version is "latest".

freddydk (changes requested):

Actions/Github-Helper.psm1
```diff
@@ -708,20 +708,170 @@ function Set-JsonContentLF {
...
+    $projects = @(@($projects.Split(',')) | ForEach-Object { $_.Replace('\','_').Replace('/','_') })
+
+    # Get the artifacts from the the CICD run
+    while($true) {
+        $artifactsURI = "$api_url/repos/$repository/actions/runs/$CICDrun/artifacts?per_page=$per_page&page=$page"
+
+        $artifacts = InvokeWebRequest -Headers $headers -Uri $artifactsURI | ConvertFrom-Json
+
+        if($artifacts.artifacts.Count -eq 0) {
+            Write-Verbose "No more artifacts found for CICD run $CICDrun"
```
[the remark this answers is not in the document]
mazhelez: Created: https://github.com/orgs/microsoft/projects/521/views/1?pane=issue&itemId=44410487

Actions/Github-Helper.psm1
```diff
@@ -708,20 +708,170 @@ function Set-JsonContentLF {
...
+
+    $headers = GetHeader -token $token
+    $lastSuccessfulCICDRun = 0
+    $per_page = 100
+    $page = 1
+
+    # Get the latest CICD workflow run
+    while($true) {
+        $runsURI = "$api_url/repos/$repository/actions/runs?per_page=$per_page&page=$page&exclude_pull_requests=true&status=success&branch=$branch"
+        Write-Host "- $runsURI"
```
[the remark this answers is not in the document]
mazhelez: Added

Actions/Github-Helper.psm1
```diff
@@ -708,20 +708,234 @@ function Set-JsonContentLF {
...
+        $workflowRuns = InvokeWebRequest -Headers $headers -Uri $runsURI | ConvertFrom-Json
+
+        if($workflowRuns.workflow_runs.Count -eq 0) {
+            # No more workflow runs, breaking out of the loop
+            break
+        }
+
+        $CICDRuns = @($workflowRuns.workflow_runs | Where-Object { $_.name -eq ' CI/CD' })
+
+        foreach($CICDRun in $CICDRuns) {
```
freddydk, line 798: Could we check whether the workflow is marked successful before checking the individual jobs? (in that case, there is no need for checking whether the build jobs are successful)

freddydk (changes requested):

Actions/Github-Helper.psm1
mazhelez: Oh, nice catch. I thought I had the check for successful workflow run.

As for the `in_progress` status, the query filters on `completed`, so in progress runs will be omitted.

freddydk (approved):

elbardel (approved):

PR #837, opened 2023-11-28
Fix partial build

mazhelez:
_**Issue**_:
Currently, when building partially (e.g. PR Build), the baseline artifacts are fetched from a CICD workflow run on the fly.
This could lead to discrepancies: one build job downloading artifacts from one CICD run, while the next one: from another CICD run (which completed in-between the two build jobs in the PR Build).

_**Solution**_: Determine the baseline CICD workflow once per PR Build and use artifacts from it in all build jobs.

Test run partial build: https://github.com/mazhelez/BCApps/actions/runs/7165434696?pr=2
Test run full build: https://github.com/mazhelez/BCApps/actions/runs/7165432643?pr=5
TODO:
- [x] Tests
- [x] Propagate workflow changes to all templates
- [x] Smarter check to determine the baseline workflow ID only when needed (non-full build)


Actions/DetermineProjectsToBuild/action.yaml
```diff
@@ -35,8 +38,7 @@ runs:
       run: |
         $errorActionPreference = "Stop"; $ProgressPreference = "SilentlyContinue"; Set-StrictMode -Version 2.0
         try {
-          $changedFiles = ${{ github.action_path }}/GetModifiedFiles.ps1 -token ${{ github.token }}
```
freddydk, line 38: If we could follow the original naming strategy and structure of the action yaml file, then you could add a test like this one: https://github.com/microsoft/AL-Go/pull/839/files#diff-d899fcc107a4ee4a13082c5f9daa7c865a2295160b6dba20ea8d45af5cdf1df9 - and ensure that parameters and everything matches the powershell script behind the action.
Unless there is a very good reason not to do that, I think we should

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.psm1
```diff
@@ -76,32 +87,91 @@ function Get-FilteredProjectsToBuild($settings, $projects, $baseFolder, $modifie
...
+            $buildDimensions += @{
+                project = $project
+                projectName = $projectSettings.projectName
+                buildMode = $buildMode
             }
         }
     }
 
-    return $filteredProjects
+    return @(, $buildDimensions) # force array
```
freddydk, line 174: We don't really use this hack anywhere else, and I don't like it - it adds an empty item to the array and you cannot use count etc. typically we cast the function as an array.

Actions/DetermineProjectsToBuild/action.yaml
mazhelez: My reasoning (back then when the action was created) was that half of the action parameters aren't actually required and the actions shouldn't be called with different values than the default (referring to `github.workspace` and `github.token`).

The script requires them because, I suppose, there is no good way to fetch these values directly in the script.

Other than that, I also believe we should take another look at the "YAML matches PS" approach. The more I read about GitHub composite actions, the more I think there're YAML goodies we don't take advantage of because of this approach.

Another example where I went rogue some time ago is the `Download Project Dependencies` action: it has two steps, not one. 😨

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.Action.ps1
```diff
@@ -15,31 +14,47 @@ $telemetryScope = $null
...
+    $allProjects, $projectsToBuild, $projectDependencies, $buildOrder = Get-ProjectsToBuild -baseFolder $baseFolder -isPartialBuild $isPartialBuild -modifiedFiles $modifiedFiles -maxBuildDepth $maxBuildDepth
     AddTelemetryProperty -telemetryScope $telemetryScope -key "projects" -value "$($allProjects -join ', ')"
+    Write-Host "::endgroup::"
     #endregion
 
     #region Action: Output
     $projectsJson = ConvertTo-Json $projectsToBuild -Depth 99 -Compress
     $projectDependenciesJson = ConvertTo-Json $projectDependencies -Depth 99 -Compress
     $buildOrderJson = ConvertTo-Json $buildOrder -Depth 99 -Compress
+    $IsPartialBuildJson = ConvertTo-Json $IsPartialBuild -Depth 99 -Compress
```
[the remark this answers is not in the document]
mazhelez: Well, talk to Copilot 😅

Yes, it's JSON value (still a JSON) 😉

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.psm1
```diff
@@ -1,68 +1,79 @@
...
+    - More than 250 files have been modified
+    - The modified files contain a file that matches one of the fullBuildPatterns
 #>
-function Get-FilteredProjectsToBuild($settings, $projects, $baseFolder, $modifiedFiles) {
+function Get-IsPartialBuild {
+    param(
+        [Parameter(HelpMessage = "The base folder", Mandatory = $true)]
+        [string] $baseFolder,
+        [Parameter(HelpMessage = "The modified files", Mandatory = $false)]
+        [array] $modifiedFiles = @()
```
aholstrup1, line 56:
```suggestion
        [string[]] $modifiedFiles = @()
```

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.psm1
```diff
@@ -124,19 +194,24 @@ function Get-FilteredProjectsToBuild($settings, $projects, $baseFolder, $modifie
...
-{
+function Get-ProjectsToBuild {
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSReviewUnusedParameter', 'modifiedFiles', Justification = 'False-positive. Used in a Where-Object filter')]
+    param (
+        [Parameter(HelpMessage = "The folder to scan for projects to build", Mandatory = $true)]
+        $baseFolder,
+        [Parameter(HelpMessage = "Whether a full build is required", Mandatory = $false)]
+        [bool] $isPartialBuild = $false,
+        [Parameter(HelpMessage = "An array of changed files paths, used to filter the projects to build", Mandatory = $false)]
+        [array] $modifiedFiles = @(),
```
aholstrup1, line 205:
```suggestion
        [string[]] $modifiedFiles = @()
```

aholstrup1 (approved):

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.psm1
```diff
@@ -1,68 +1,79 @@
...
-.Synopsis
-    Creates buils dimensions for a list of projects.
-
-.Outputs
-    An array of build dimensions for the projects and their corresponding build modes.
-    Each build dimension is a hashtable with the following keys:
-    - project: The name of the AL-Go project
-    - buildMode: The build mode to use for the project
+    .Synopsis
+        Gets the modified files in a GitHub pull request.
```
freddydk, line 3: IMO - these are just individual helperfunctions and doesn't belong in a module,
They are only used from within the action .ps1 script and only to make reading code easier.
Module implies variable scope and if you want to use functions from AL-Go-Helper or GitHub-Helper you need to re-import other modules and scripts.
These modules then live in memory multiple times.
Importing modules and calling functions in modules is also much slower than just using helperfunctions.

Now there can be many reasons why we should structure things in modules - but a list of helperfunctions isn't one IMO

Beside that, using dashes in function names makes PowerShell apply rules to the naming (which then are excluded globally because of this) - why not just use function names like GetModifiedFiles in a helperfunction script?

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.psm1
```diff
@@ -124,19 +194,24 @@ function Get-FilteredProjectsToBuild($settings, $projects, $baseFolder, $modifie
             - project: The project to build
             - buildMode: The build mode to use
 #>
```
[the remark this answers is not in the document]
mazhelez: The _PSUseSingularNouns_ rule doesn't bring much value anyway.

The dash here is used to differentiate between public and private function of the module. `Export-ModuleMember *-*` would export only functions that match the *-* pattern.

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.psm1
mazhelez: I see we haven't really settled on how to organize helper functions. 
Why is it AL-Go-Helper.**ps1** and Github-Helper.**psm1**?

PS modules provide encapsulation and that's more important than the performance gain (that is not much and it's not like these modules are imported so many times anyway).
Using scripts for helper function creates a mess and it's matter of time to have unwanted function overwrites. Not that this cannot happen with modules, but it's way less likely, especially if there are structured well.

My idea of having a PS modules (not necessarily related to an action) is to provide good interface of functions to be used both in and outside actions. 

In this case `Get-ModifiedFiles` could be a one liner (`gh pr diff $prNumber--repo $repo --name-only`) and possibly better suited for the Github-Helper module, but I didn't want to change it in this PR.

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.psm1
```diff
@@ -124,19 +194,24 @@ function Get-FilteredProjectsToBuild($settings, $projects, $baseFolder, $modifie
...
-    [Parameter(HelpMessage = "The folder to scan for projects to build", Mandatory = $true)]
-    $baseFolder,
-    [Parameter(HelpMessage = "An array of changed files paths, used to filter the projects to build", Mandatory = $false)]
-    $modifiedFiles = @(),
-    [Parameter(HelpMessage = "The maximum depth to build the dependency tree", Mandatory = $false)]
-    $maxBuildDepth = 0
-)
-{
+function Get-ProjectsToBuild {
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSReviewUnusedParameter', 'modifiedFiles', Justification = 'False-positive. Used in a Where-Object filter')]
```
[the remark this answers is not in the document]
mazhelez: Yeah, but then the convenient PS syntax is lost :/

How about this hack:
Instead of this:
```
... | Where-Object {... $param }
```
Use this:
```
$var = $param
... | Where-Object {... $var}
```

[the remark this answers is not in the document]
freddydk: The buildDimensions array in the function above is built without piping, I think that is nice code.

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.psm1
freddydk: I get that modules should be used for stuff that are used in multiple places, but putting 5 functions that are used only by one script in a module doesn't provide any value IMO. The action in itself is already encapsulated.
Moving reusable code to global modules makes a ton of sense. Having a .ps1 and a .psm1 side by side in all actions doesn't IMO.

The action script imports github-helper module and you now have to import github-helper in the function in the module as well - that is the part I don't like.

BTW, I agree that AL-Go-Helper should have been a module, and GitHub-Helper should have been part of this module, bad choices where made, we can change those.

freddydk: In the ReadSecrets, I think there is a reason for the encapsulation (module variables etc)
TBH - I didn't notice the sign module, but I just comment on this because this is a change, which I don't think is necessary and I don't think is the right thing to do.
If you feel strongly about it - then resolve the conversation and discussion is over.

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.Action.ps1
```diff
@@ -36,10 +47,12 @@ try {
     Add-Content -Encoding UTF8 -Path $env:GITHUB_OUTPUT -Value "ProjectsJson=$projectsJson"
     Add-Content -Encoding UTF8 -Path $env:GITHUB_OUTPUT -Value "ProjectDependenciesJson=$projectDependenciesJson"
     Add-Content -Encoding UTF8 -Path $env:GITHUB_OUTPUT -Value "BuildOrderJson=$buildOrderJson"
+    Add-Content -Encoding UTF8 -Path $env:GITHUB_OUTPUT -Value "IsPartialBuild=$isPartialBuild"
```
freddydk, line 51: I think this is a confusing name - is should be something like buildAllProjects IMO (reversed meaning) - partial builds could be within projects and might be something we would do in the future.

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.psm1
aholstrup1: > I believe it's a good approach to have some function separation and encapsulation.

I tend to agree with this. Unless using Powershell Modules are significantly slower, I don't see any issues with using them in an action like this one.

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.Action.ps1
mazhelez: Renamed to `BuildAllProjects`.

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.psm1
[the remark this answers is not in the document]
mazhelez: Fixed.

freddydk (approved):

aholstrup1 (approved):

PR #853, opened 2023-12-12
Add support for custom templates + customizations

freddydk:
Changes included in this PR:
- Support custom template repository = Point to an existing AL-Go repository as a template, like:

image

In this case, org/myptetemplate is a standard AL-Go repository, which gets updated as any other AL-Go repository using Update AL-Go System Files. This is a way of standardizing your repository based on type specific AL-Go templates (which can be private or public).

Normally, Update AL-Go System Files follows this mechanism:

| File Type | Repository file(s) | AL-Go Template |
| :-- | :-- | :-: |
| Workflows | .github/workflows/*.yaml | <- |
| Repo Scripts | .github/*.ps1 | <- |
| Project Scripts | .AL-Go/*.ps1 | <- |

Workflows and scripts are read from the template repository and modified based on settings and saved to the repository.

Using a custom template repository, the picture looks like this:

| File Type | Repository file(s) | Custom Template | AL-Go Template |
| :-- | :-- | :-: | :-: |
| Workflows | .github/workflows/*.yaml | <- Customizations | <- |
| Workflows | .github/workflows/*.yaml | <- New files | |
| Repo Scripts | .github/*.ps1 | | <- |
| Repo Scripts | .github/*.ps1 | <- New files | |
| Repo Settings | .github/templateRepoSettings.doNotEdit.json | <- .github/AL-Go-Settings.json | |
| Project Scripts | .AL-Go/*.ps1 | | <- |
| Project Scripts | .AL-Go/*.ps1 | <- New files | |
| Project Settings | .github/templateProjectSettings.doNotEdit.json | <- .AL-Go/settings.json | |

Workflows and scripts are read from the "real" template repository and modified based on settings. Customizations for workflows in the custom template repository are applied.

New workflows and scripts from the custom template repository are copied over.

Repo and Project settings from the custom template repository are copied over to new settings files, which are applied before the final repos repo and project settings in order to allow the indirect template to add new defaults to settings to be applied to all repositories.

The final repository can also contain customizations to workflows and the winner here is the indirect template repository (if a custom job exists in both places).

This PR: https://github.com/freddydk/customized/pull/17 - is an example of a repository with customizations of CI/CD and _BuildALGoProject, which runs Update AL-Go System Files and gets customizations from the indirect template repository (both settings, scripts and workflow customizations)

Documentation on customization capabilities is here: https://github.com/freddydk/AL-Go/blob/customize/Scenarios/CustomizingALGoForGitHub.md


mortenseifert:
@freddydk, do you have an estimated time of completion for this PR? We've been testing it for a while and are considering deploying it live. However, we're a bit cautious about when the PR will be finalized and whether it will disrupt any modifications we've implemented.

freddydk:
Check https://github.com/microsoft/AL-Go/pull/1097 - to ensure the scenario is supported

freddydk:
@jonaswre, @mortenseifert and @lippertmarkus - feel free to add your code review as well

jonaswre:
Hi @freddydk, i'm currently on vacation until September 2, but I'll test it in my demo org once im back. Im excited for this change. 

But trusted nuget feeds would be my first priority. We've recently got more requests to use our apps as dependencies so will need to distribute runtimes to customers. We've implemented the GenerateBCNuget Repo which works great but distributing them is the next step for us. If I can help with that let me know.

freddydk:
> Hi @freddydk, i'm currently on vacation until September 2, but I'll test it in my demo org once im back. Im excited for this change.
> 
> But trusted nuget feeds would be my first priority. We've recently got more requests to use our apps as dependencies so will need to distribute runtimes to customers. We've implemented the GenerateBCNuget Repo which works great but distributing them is the next step for us. If I can help with that let me know.

I will resume the work on NuGet while this PR is in review.

lippertmarkus (approved):

Actions/CheckForUpdates/yamlclass.ps1
```diff
@@ -185,4 +224,255 @@ class Yaml {
...
+                $permission = $permissions."$_"
+                $permission2 = $permissions2."$_"
+                if ($permission -eq 'write' -or $permission2 -eq 'write') {
+                    $permissions."$_" = 'write'
+                }
+                elseif ($permission -eq 'read' -or $permission2 -eq 'read') {
+                    $permissions."$_" = 'read'
+                }
+                else {
+                    $permissions."$_" = 'none'
```
aholstrup1, line 426: Would we set the permission to None if one workflow has `actions: read` and the other has `actions: write`? 🤔

Scenarios/CustomizingALGoForGitHub.md
```diff
@@ -0,0 +1,301 @@
...
+    if: (!cancelled())
+    runs-on: [ windows-latest ]
+    steps:
+      ...
+```
+
+Adding a custom job like this, will cause this job to run simultaneously with the deploy and the deliver jobs.
+
+> \[!NOTE\]
+> All custom jobs will be moved to the tail of the yaml file when running Update AL-Go System Files, but dependencies to/from the custom jobs will be maintained.
```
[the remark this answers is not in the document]
freddydk: This would be the responsibility of the person adding the custom job to fix this problem.

Scenarios/Contribute.md
```diff
@@ -5,40 +5,36 @@ This section describes how to contribute to AL-Go. How to set up your own enviro
...
-- Use 3 public repositories in your own **personal GitHub account** (AL-Go-PTE, AL-Go-AppSource and AL-Go-Actions, much like in production)
+- Use 2 public repositories in your own **personal GitHub account** (AL-Go-PTE, AL-Go-AppSource and AL-Go-Actions, much like in production)
 
 ## Use a fork of AL-Go for GitHub in "development mode"
 
 1. Fork the [https://github.com/microsoft/AL-Go](https://github.com/microsoft/AL-Go) repository to your **personal GitHub account**.
 1. You can optionally also create a branch in the AL-Go fork for the feature you are working on.
 
-**https://github.com/<yourGitHubUserName>/AL-Go@<yourBranch>** can now be used as a template in your AL-Go project when running _Update AL-Go System Files_ to use the actions/workflows from this fork.
+**<yourGitHubUserName>/AL-Go@<yourBranch>** can now be used as a template in your AL-Go project when running _Update AL-Go System Files_ to use the actions/workflows from this fork.
```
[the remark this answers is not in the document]
freddydk: Yes, they can do this - but this requires different skills then setting up some indirect templates.
Plus, it would still only give them a PTE and a AppSource template and it is much harder to maintain a fork of AL-Go than to maintain a set of indirect templates for their various project types.

Actions/CheckForUpdates/yamlclass.ps1
freddydk: No, if either we (or a customization) has write permissions - we set write.
Then, if either we (or a customization has read permissions - we set read.
Else we set none.

Actions/CheckForUpdates/yamlclass.ps1
```diff
@@ -185,4 +224,255 @@ class Yaml {
...
+
+    static [string[]] GetPermissionsArray([PSCustomObject] $permissions) {
+        $permissionsArray = @()
+        $permissions.PSObject.Properties.Name | ForEach-Object {
+            $permissionsArray += "$($_): $($permissions."$_")"
+        }
+        return $permissionsArray
+    }
+
+    static [PSCustomObject] MergePermissions([PSCustomObject] $permissions, [PSCustomObject] $permissions2) {
```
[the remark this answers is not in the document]
freddydk: We do not use read-all or write-all and people cannot customize our workflows to do this.
If they customize permissions, it is because they have steps or jobs requiring additional permissions.

Actions/CheckForUpdates/yamlclass.ps1
```diff
@@ -185,4 +224,255 @@ class Yaml {
...
+        $srcYaml = [Yaml]::new($srcContent.Value.Split("`n"))
+        try {
+            $yaml = [Yaml]::Load($yamlFile)
+        }
+        catch {
+            return
+        }
+        # Merge permissions
+        Write-host "Merge permissions"
+        $srcPermissionsObj = $srcYaml.Get('permissions:/')
```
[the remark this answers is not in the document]
freddydk: I actually thought that the upper permissions setting was the permissions allowed for all jobs and no job could set additional permissions.
When I read the docs now - it looks like the upper permissions settings is indeed the permissions allowed for all jobs, but jobs can add additional permissions if they like (except for the reusable workflows).

Will check whether this changes anything,

Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -186,6 +186,19 @@ try {
     }
     $authContext = $null
     $environmentName = ""
+    if ($cicdAuthContext) {
```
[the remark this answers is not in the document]
freddydk: This is the setting for allowing us to run tests in online sandboxes - and the setting Cosmo needs in order to run pipelines with their Alpaca containes.

Could be moved to a seperate PR - but is needed together with this for Cosmo.

Will add it to the docs.

Scenarios/Contribute.md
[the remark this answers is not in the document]
aholstrup1: Why is it much harder to maintain a fork of AL-Go than indirect templates? Don't you just need to occasionally run: 
```powershell
git remote add upstream "https://github.com/microsoft/AL-Go"
git pull upstream
```
And then fix the merge conflicts?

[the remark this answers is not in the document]
freddydk: The requirements are:
1. Be able to create x template repositories for various types of projects. This could be different types of on-premises solutions, PTEs, ISV solutions.
2. Template repositories must be able to be private
3. Template repositories must be able to use AL-Go main, preview or fixed version if necessary.
4.  It should not require deep knowledge of yaml, git, powershell or like to create a new template repository.

There are two ways of using a fork.
You can fork AL-Go or you can fork AL-Go-PTE|AL-Go-AppSource.
If you fork AL-Go-PTE|AL-Go-AppSource, then (1) would have to be in different organizations (you can only have 1 fork in an org). 2 would require a hack and git gymnastics, 3. would be impossible (or at least I don't know how to) and my guess is that you will quickly get into a situation where our changes require you to have more yaml knowledge.

If you fork AL-Go and use direct AL-Go development (using org/AL-Go@branch) then a change to that branch might break your pipelines and even Update AL-Go System Files. 1 would require you to maintain multiple forks. 2 would be hard, 3 would require you to have a preview branch and a main branch with your customizations and then run update in all repositories (potentially 1000s) or do some git gymnastics to update your branch with a different branch.  You would definitely need git skills and yaml skills.

If you fork AL-Go and use Deploy to deploy to your own AL-Go-PTE and your own AL-Go-AppSource (like a complete parallel setup with our AL-Go version), then you are basically developing your own AL-Go version. Having multiple template repositories is not really supported, they cannot be private, it is very hard to maintain a main and a private version and it would definitely require skills that we don't normally require.

Basically this is similar to back in the days, where Navision used source code customizations. Everybody would modify the hell out of the source and would end up not staying up-to-date because it was too hard to follow along.

Today we have an app model with events, which is kind of what this PR implements

You are more than welcome to try and prove me wrong, but by limiting the number of places we support changes, we will keep more people on standard - many things might just be to add some workflows and/or jobs to enhance their experience.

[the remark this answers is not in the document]
PatrickSchiefer: > The requirements are:
> 
> 1. Be able to create x template repositories for various types of projects. This could be different types of on-premises solutions, PTEs, ISV solutions.
> 2. Template repositories must be able to be private
> 3. Template repositories must be able to use AL-Go main, preview or fixed version if necessary.
> 4. It should not require deep knowledge of yaml, git, powershell or like to create a new template repository.
> 

@freddydk this is exactly what we would need to make COSMO Alpaca using AL:GO for GitHub
Since we're having a few hundred repos managed by COSMO Alpaca we would also need different Templates. Some of the templates also might be private because they will contain some closed source. 

When managing so many repos, it is absolutely necessary for us to have 100% control over the templates and the version used. With the custom templates we can provide this for our projects.

mazhelez (commented):
Posting an mid-way review as I got carried away. I'll circle back to this PR.

RELEASENOTES.md
```diff
@@ -1,3 +1,32 @@
+### Issues
+
+### New Settings
+
+- `customALGoSystemFiles` is an array of JSON objects, which holds information about custom AL-Go System Files, which will be applied during Update AL-Go System Files. Every object can hold these 4 properties:
```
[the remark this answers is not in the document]
freddydk: I kind of agree, but I also don't think we should have other settings files.
IMO we should make the same change to settings as I did to secrets a while back, where we specify which secrets we want to read when calling ReadSecrets.

So, the workflows should know (and specify) which settings to look for and the only settings returned would be the once specified. This way the settings carried around in outputs would be limited.

Actions/CheckForUpdates/CheckForUpdates.HelperFunctions.ps1
```diff
@@ -420,32 +434,153 @@ function UpdateSettingsFile {
...
+            if ($content -ne $existingContent) {
+                Write-Output @{ "DstFile" = $_; "content" = $content }
+            }
+        }
+    }
+
+    if ($settings.customALGoSystemFiles -isnot [Array]) {
+        throw "customALGoSystemFiles setting is wrongly formatted, must be an array of objects. See https://aka.ms/algosettings#customalgosystemfiles."
+    }
+    foreach($customspec in $settings.customALGoSystemFiles) {
```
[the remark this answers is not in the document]
freddydk: Haven't worked with json schemas, but sounds like a good idea.
Will have a look - but maybe this shouldn't be part of this PR?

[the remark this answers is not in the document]
mazhelez: Why not? This `customALGoSystemFiles` is new JSON construct.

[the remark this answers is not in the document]
freddydk: Json schema should cover the entire settings file I guess - I don't want to delay this PR for that.

Scenarios/CustomizingALGoForGitHub.md
```diff
@@ -0,0 +1,301 @@
...
+> Script overrides will almost certainly be broken in the future. The current script overrides is very much tied to the current implementation of the `Run-AlPipeline` function in BcContainerHelper. In the future, we will move this functionality to GitHub actions and no longer depend on BcContainerHelper and Run-AlPipeline. At that time, these script overrides will have to be changed to follow the new implementation.
+
+<a id="customALGoSystemFiles"></a>
+
+### Adding custom workflows and/or scripts using a URL
+
+By adding a setting called [`customALGoSystemFiles`](https://aka.ms/algosettings#customalgosystemfiles) in your [repo settings](https://aka.ms/algosettings#settings), you can tell AL-Go for GitHub that these files should be included in the update. Example:
+
+```
+  "customALGoSystemFiles": [
```
mazhelez, line 151: Is `customALGoSystemFiles` supposed to be defined on the indirect template or on the end repositories (the ones that consume the indirect template)?

If it's on the end repos, then it defeats part of the purpose of the indirect template. E.g. in the case of a new common script, the setting needs to be added manually on X repos.

If it's on the indirect template template, then shouldn't relative paths suffice? Does it need to be so complex?

freddydk: Not sure what you mean by relative paths?
The Source is an absolute URL to a file to download - rarely something in the same repo.

The setting can be on both.
If it is defined on the indirect template - it is handled when updating the indirect template.
If it is defined on the end repos - it is handled when updating the end repo (not everybody will have an indirect template)

aholstrup1: Why not require an indirect template if you want to sync custom scripts? Is there a requirement for two different mechanisms of syncing scripts?

Actions/CheckForUpdates/CheckForUpdates.HelperFunctions.ps1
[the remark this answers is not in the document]
freddydk: Not in this PR - checking the schema belongs where the settings file is read - this is where the settings are used and the only reason for this check is that the error down the road if this is wrong was very hard to decipher.

Scenarios/CustomizingALGoForGitHub.md
freddydk: Those are two different things.
An indirect template is something people create in their organization for types of repos.
custom files could be workflows or scripts created by other people (like the onprem publisher created by a partner)

Scenarios/CustomizingALGoForGitHub.md
```diff
@@ -0,0 +1,301 @@
...
+    needs: [ Initialization, Build ]
+    if: (!cancelled()) && (needs.Build.result == 'success')
+    runs-on: [ ubuntu-latest ]
+    steps:
+      - name: Create Tag
+        run: |
+          Write-Host "Create Tag"
+
+  PostProcess:
+    needs: [ Initialization, Build2, Build1, Build, Deploy, Deliver, DeployALDoc, CustomJob-CreateBuildTag ]
```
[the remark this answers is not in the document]
freddydk: Strange - when you add that to the indirect template - it shouldn't do anything in the end repo.
Will have a look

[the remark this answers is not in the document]
aholstrup1: Shouldn't it add CustomJob-CreateBuildTag to the needs block in PostProcess? 🤔 Maybe I misunderstood it

RELEASENOTES.md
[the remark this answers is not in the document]
PatrickSchiefer: As a user I can say having more setting files would make it more complex. Most of the users I know prefer one big setting file where they have everything in one file.

jwikman:
What's needed to get this one completed?

freddydk:
> What's needed to get this one completed?

Time in my calendar (and the reviewers)

Very much hoping to get this done in March or April.

jwikman:
> Time in my calendar (and the reviewers)

I see, sorry to nag about this then. 😅

But since I cannot add time to your calendar, is it something I can do to help you out on this PR?

This is quite a massive PR, and I'm not (yet) enough into this repo to do a code review worth the name. But maybe I can help with something else?

Is it ready to fully test out, or should I wait some more?
Any known issues that still exists that are blocking us? (I.e. app auth with internal repos?)
Any other issues that is non-blocking, but are good to know about?

freddydk:
> > Time in my calendar (and the reviewers)
> 
> I see, sorry to nag about this then. 😅
> 
> But since I cannot add time to your calendar, is it something I can do to help you out on this PR?
> 
> This is quite a massive PR, and I'm not (yet) enough into this repo to do a code review worth the name. But maybe I can help with something else?
> 
> Is it ready to fully test out, or should I wait some more? Any known issues that still exists that are blocking us? (I.e. app auth with internal repos?) Any other issues that is non-blocking, but are good to know about?

I have a few things that I need to change (primarily in the settings area) - but it is fairly ready
Currently, the functionality in this PR tries to merge settings from the template to the final repo - instead, it should copy the template settings file to a AL-Go-Settings.template.json file - and read that as part of the settings chain.
This allows for default changes in the template, which today is hard.

jwikman:
I take that as I could try this out as a preview in our templates, and I'll keep an eye on this PR for updates to the settings part.

lippertmarkus:
We already use it since a few weeks for several repos with our own template and it works fine so far :)

jwikman:
> We already use it since a few weeks for several repos with our own template and it works fine so far :)

Perfect! 👍

freddydk:
> We already use it since a few weeks for several repos with our own template and it works fine so far :)

And you also found a bug and submitted a fix :-)

Actions/CheckForUpdates/yamlclass.ps1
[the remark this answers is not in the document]
freddydk: Have fixed this so that permissions on any job will be merged or added

Actions/CheckForUpdates/CheckForUpdates.ps1
```diff
@@ -86,11 +87,38 @@ $templateInfo = "$templateOwner/$($templateUrl.Split('/')[4])"
...
-    $ALGoSettingsFile = Join-Path $templateFolder "*/$repoSettingsFile"
-    if (Test-Path -Path $ALGoSettingsFile -PathType Leaf) {
-        $templateRepoSettings = Get-Content $ALGoSettingsFile -Encoding UTF8 | ConvertFrom-Json | ConvertTo-HashTable -Recurse
+    $templateRepoSettingsFile = Join-Path $templateFolder "*/$RepoSettingsFile"
+    if (Test-Path -Path $templateRepoSettingsFile -PathType Leaf) {
+        $templateRepoSettings = Get-Content $templateRepoSettingsFile -Encoding UTF8 | ConvertFrom-Json | ConvertTo-HashTable -Recurse
         if ($templateRepoSettings.Keys -contains "templateUrl" -and $templateRepoSettings.templateUrl -ne $templateUrl) {
-            throw "The specified template repository is not a template repository, but instead another AL-Go repository. This is not supported."
+            # The template repository is a url to another AL-Go repository (an indirect template repository)
+            # TemplateUrl and TemplateSha from .github/AL-Go-Settings.json in the indirect template reposotiry points to the "real" template repository
```
jwikman, line 95: Typo
```suggestion
            # TemplateUrl and TemplateSha from .github/AL-Go-Settings.json in the indirect template repository points to the "real" template repository
```

jwikman (changes requested):
While trying to wrap my head around this functionality, I found some small typos and questions popped, so I thought it was better to put it out here. 

I have not gone through all changes (yet) and haven't tried it out yet, so take it for what it is :)

Scenarios/CustomizingALGoForGitHub.md
```diff
@@ -0,0 +1,301 @@
...
+    }
+  ]
+```
+
+`customALGoSystemFiles` is an array of objects, which currently can have 4 properties:
+
+| Property | Description | Mandatory | Default |
+| :-- | :-- | :-: | :-- |
+| Destination | Path in which the file should be placed. Can include the filename if the source doesn't point to a .zip file, must include a terminating / or \\ if a filename is not included. | Yes | |
+| Source | URL to a either a single file or a .zip file containing custom AL-Go System Files. Must be https. | Yes | |
```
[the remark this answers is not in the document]
freddydk: Working on that piece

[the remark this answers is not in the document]
jwikman: Great!

Actions/AL-Go-Helper.ps1
```diff
@@ -535,8 +543,10 @@ function MergeCustomObjectIntoOrderedDictionary {
 # Read settings from the settings files
 # Settings are read from the following files:
 # - ALGoOrgSettings (github Variable)                 = Organization settings variable
+# - .github/templateRepoSettings.json                 = Repository settings from indirect template
```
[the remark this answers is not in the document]
freddydk: Yes, even though they are files, that people shouldn't modify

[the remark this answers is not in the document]
jwikman: I realized that today... 😅

But maybe that should be documented in settings.md anyway, because since they will exist - someone will probably change them and wonder what is happening when they are overwritten...

Actions/CheckForUpdates/CheckForUpdates.ps1
```diff
@@ -86,11 +87,38 @@ $templateInfo = "$templateOwner/$($templateUrl.Split('/')[4])"
...
+            $realTemplateUrl = $templateRepoSettings.templateUrl
+            if ($templateRepoSettings.Keys -contains "templateSha") {
+                $realTemplateSha = $templateRepoSettings.templateSha
+            }
+            else {
+                $realTemplateSha = ""
+            }
+
+            # Download the "real" template repository - use downloadLatest if no TemplateSha is specified in the indirect template repository
+            $realTemplateFolder = DownloadTemplateRepository -headers $headers -templateUrl $realTemplateUrl -templateSha ([ref]$realTemplateSha) -downloadLatest ($realTemplateSha -eq '')
```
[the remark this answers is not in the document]
jwikman: This seems to work for me:

```suggestion
            $repositories = @($realTemplateUrl.Split('@')[0])
            $realTemplateRepoReadToken = GetAccessToken -token $token -permissions @{"actions" = "read"; "contents" = "read"; "metadata" = "read" } -repositories $repositories
            $realTemplateRepoHeaders = GetHeaders -token $realTemplateRepoReadToken
            $realTemplateFolder = DownloadTemplateRepository -headers $realTemplateRepoHeaders -templateUrl $realTemplateUrl -templateSha ([ref]$realTemplateSha) -downloadLatest ($realTemplateSha -eq '')
```

Scenarios/CustomizingALGoForGitHub.md
```diff
@@ -0,0 +1,301 @@
...
+
+> [!CAUTION]
+> This workflow gets triggered when the CI/CD workflow has completed. Note that the name of the CI/CD workflow currently is prefixed with a space, this space will very likely be removed in the future, which is why we specify both names in this example. Obviously this workflow would break if we decide to rename the CI/CD workflow to something different.
+
+### Adding custom scripts
+
+You can add custom powershell scripts under the .github folder for repository scoped scripts or in the .AL-Go folder for project scoped scripts. Specially named scripts in the .AL-Go folder can override standard functionality in AL-Go for GitHub workflows. A list of these script overrides can be found [here](https://aka.ms/algosettings#customdeployment). Scripts under the .github folder can be used in custom workflows instead of using inline scripts inside the workflow.
```
jwikman, line 128: I believe that this link should point to the "Run-AlPipeline script override" section, right?
```suggestion
You can add custom powershell scripts under the .github folder for repository scoped scripts or in the .AL-Go folder for project scoped scripts. Specially named scripts in the .AL-Go folder can override standard functionality in AL-Go for GitHub workflows. A list of these script overrides can be found [here](https://aka.ms/algosettings#scriptoverrides). Scripts under the .github folder can be used in custom workflows instead of using inline scripts inside the workflow.
```

jwikman:
I've added a custom step in `_BuildALGoProject.yaml` in the indirect template. In this step I want to invoke a .ps1 file. But the .ps1 files added to the .github folder is not transfered to the repo that is using this indirect template.
What is needed to get those .ps1 files from the indirect template to the repo using this template? 
I rather not go with the "**Adding custom** workflows and/or **scripts using a URL**" feature, since that feels a bit too complicated to maintain for this scenario.

freddydk:
> I've added a custom step in `_BuildALGoProject.yaml` in the indirect template. In this step I want to invoke a .ps1 file. But the .ps1 files added to the .github folder is not transfered to the repo that is using this indirect template. What is needed to get those .ps1 files from the indirect template to the repo using this template? I rather not go with the "**Adding custom** workflows and/or **scripts using a URL**" feature, since that feels a bit too complicated to maintain for this scenario.

.ps1 files added to .github should be transferred to the final repo - did you add it to a subfolder or directly in .github?

jwikman:
> .ps1 files added to .github should be transferred to the final repo - did you add it to a subfolder or directly in .github?
> 
> 

Tested both, will try again if you say that  it _should_ work. 👍

jwikman:
> .ps1 files added to .github should be transferred to the final repo - did you add it to a subfolder or directly in .github?

I tried again, with no luck. So, I have dived into the code again. :)

According to [the comments and code](https://github.com/freddydk/AL-Go/blob/7fa0ca6cccabab47358f98476adc05b8bc1a687e/Actions/CheckForUpdates/CheckForUpdates.ps1#L131-L154) it seems as only *.ps1 files are copied from the .AL-Go folder, not the .github folder, if I'm not missing something...? 🤔

freddydk:
> > .ps1 files added to .github should be transferred to the final repo - did you add it to a subfolder or directly in .github?
> 
> I tried again, with no luck. So, I have dived into the code again. :)
> 
> According to [the comments and code](https://github.com/freddydk/AL-Go/blob/7fa0ca6cccabab47358f98476adc05b8bc1a687e/Actions/CheckForUpdates/CheckForUpdates.ps1#L131-L154) it seems as only *.ps1 files are copied from the .AL-Go folder, not the .github folder, if I'm not missing something...? 🤔

You might be right - I do think the .github folder needs to be included though

jwikman:
> You might be right - I do think the .github folder needs to be included though

Yes, I believe so. If the .ps1 is going to be invoked from a pipeline, I think the .github (or a subfolder) is the most appropriate place.

Adding the .github folder was pretty straight forward, just added the following to the $realTemplateFolder condition block
```powershell
@{ 'dstPath' = '.github'; 'newname' = ''; 'srcPath' = '.github'; 'pattern' = '*.ps1'; 'type' = 'script' }
```

Are there any (simple) way to do this recursive, to also include subfolders with .ps1 files?

Actions/CheckForUpdates/CheckForUpdates.ps1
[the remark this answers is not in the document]
freddydk: I merged the updated code to the customize branch, @jwikman, please let me know whether this works for you.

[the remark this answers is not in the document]
jwikman: > I merged the updated code to the customize branch, @jwikman, please let me know whether this works for you.

Thanks @freddydk, I will test this out!

Actions/CheckForUpdates/CheckForUpdates.ps1
```diff
@@ -109,17 +122,24 @@ if (-not $isDirectALGo) {
...
-    @{ 'dstPath' = '.github'; 'srcPath' = '.github'; 'pattern' = '*.copy.md'; 'type' = 'releasenotes' }
+    @{ 'dstPath' = (Join-Path '.github' 'workflows'); 'newname' = ''; 'srcPath' = Join-Path '.github' 'workflows'; 'pattern' = '*'; 'type' = 'workflow' },
+    @{ 'dstPath' = '.github'; 'newname' = ''; 'srcPath' = '.github'; 'pattern' = '*.copy.md'; 'type' = 'releasenotes' }
 )
 
+if ($realTemplateFolder) {
+    $checkfiles += @(
+        @{ 'dstPath' = ([system.IO.Path]::GetDirectoryName($IndirectTemplateRepoSettingsFile)); 'newname' = ([system.IO.Path]::GetFileName($IndirectTemplateRepoSettingsFile)); 'SrcPath' = ([system.IO.Path]::GetDirectoryName($RepoSettingsFile)); 'pattern' = ([system.IO.Path]::GetFileName($RepoSettingsFile)); 'type' = 'template repo settings' }
+        @{ 'dstPath' = ([system.IO.Path]::GetDirectoryName($IndirectTemplateProjectSettingsFile)); 'newname' = ([system.IO.Path]::GetFileName($IndirectTemplateProjectSettingsFile)); 'SrcPath' = ([system.IO.Path]::GetDirectoryName($ALGoSettingsFile)); 'pattern' = ([system.IO.Path]::GetFileName($ALGoSettingsFile)); ; 'type' = 'template project settings' }
+    )
```
jwikman, line 133: I believe that .ps1 files from the .github folder should be added here as well (as mentioned before somewhere).
Don't know how to check files recursively here, but at least include *.ps1 from the folder:
```suggestion
        @{ 'dstPath' = '.github'; 'newname' = ''; 'srcPath' = '.github'; 'pattern' = '*.ps1'; 'type' = 'script' }
    )
```

Actions/CheckForUpdates/CheckForUpdates.HelperFunctions.ps1
```diff
@@ -12,12 +26,28 @@ If true, the latest SHA of the template repository will be downloaded
...
 
+    $templateRepositoryUrl = $templateUrl.Split('@')[0]
+    $templateRepository = $templateRepositoryUrl.Split('/')[-2..-1] -join '/'
+
+    if ([string]::IsNullOrEmpty($ENV:GITHUB_TOKEN)) {
+        Write-Host "::WARNING::GITHUB_TOKEN is not set."
+    }
+    # Use Authenticated API request if possible to avoid the 60 API calls per hour limit
+    $headers = GetHeaders -token $env:GITHUB_TOKEN -repository $templateRepository
+    $response = Invoke-WebRequest -Headers $headers -Method Head -Uri $templateRepositoryUrl -ErrorAction SilentlyContinue
```
jwikman, line 43:
```suggestion
    $response = $null
    try {
        $response = Invoke-WebRequest -Headers $headers -Method Head -Uri $templateRepositoryUrl -ErrorAction SilentlyContinue
    }
    catch {
        # Ignore error, $response will be null
    }
```

freddydk: Isn't this the same??? (with -erroraction silentlyContinue - would this ever throw?)

jwikman: Yeah, that was I first thought as well... 
The error handling is not 100% straight forward. 😂

Some errors are terminating and some are not. The $ErrorActionPreference works on both, but the -ErrorAction parameter only works on non-terminating... Clear as mud.

Actions/CheckForUpdates/CheckForUpdates.ps1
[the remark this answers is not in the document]
jwikman: Oh, btw, I like the refactoring to move the token and headers handling inside `DownloadTemplateRepository()` 👍

Actions/CheckForUpdates/CheckForUpdates.HelperFunctions.ps1
jwikman: Instead of the, IMO, ugly try/catch approach, I just discovered the `SkipHttpErrorCheck` parameter of the `Invoke-WebRequest` command.
With that, a response is returned even on failure http status.
Ref for SkipHttpErrorCheck: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-webrequest?view=powershell-7.5#-skiphttperrorcheck
Ref for 4XX and 5XX statuses treated as terminating errors: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-webrequest?view=powershell-7.5#example-7-catch-non-success-messages-from-invoke-webrequest

I haven't tested yet, but this change would probably work as well:
```suggestion
    $response = Invoke-WebRequest -Headers $headers -Method Head -Uri $templateRepositoryUrl  -SkipHttpErrorCheck

```
Infrastructure errors such as "No such host is known" will still be treated as errors, but that should probably be OK in our scenario, don't you think? We will probably only receive 40X statuses here (except the expected 200, of course)

Will get back after a test.

jwikman: Oh, `SkipHttpErrorCheck` was added in PowerShell 7, and we are running on default shell...
We should probably stick with the try/catch approach, to support all PS versions.

Actions/AL-Go-Helper.ps1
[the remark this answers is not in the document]
freddydk: Have renamed the files to templateRepoSettings.doNotEdit.json

Actions/CheckForUpdates/CheckForUpdates.HelperFunctions.ps1
```diff
@@ -488,46 +507,182 @@ function GetModifiedSettingsContent {
         $schemaKey = '$schema'
         $schemaValue = $srcSettings."$schemaKey"
 
+    $schemaKey = '$schema'
+    if ($srcSettings.PSObject.Properties.Name -eq $schemaKey) {
```
ps610, line 511: @freddydk small merge issue: missing closing '}' for this if statement. When trying to import the script, an exception is thrown (probably also the reason why the CI/CD workflow fails now: https://github.com/microsoft/AL-Go/actions/runs/15161710242/job/42629159545?pr=853#step:5:2123)

ps610: Also `$schemaKey = '$schema'` seems to be duplicated

Templates/Per Tenant Extension/.github/workflows/_BuildALGoProject.yaml
```diff
@@ -167,6 +171,7 @@ jobs:
         env:
           Secrets: '${{ steps.ReadSecrets.outputs.Secrets }}'
           BuildMode: ${{ inputs.buildMode }}
+          NeedsContext: ${{ inputs.needsContext }}
```
aholstrup1, line 174: Is this referenced anywhere? I can't seem to find how it's used

Scenarios/CustomizingALGoForGitHub.md
```diff
@@ -0,0 +1,282 @@
+# Customizing AL-Go for GitHub
+
+AL-Go for GitHub is a plug-and-play DevOps solution, intended to support 100% of the functionality needed by 90% of the people developing applications for Microsoft Dynamics 365 Business Central out-of-the-box.
+
+If AL-Go functionality out-of-the-box doesn't match your needs, you really have three options:
+
+1. Customize AL-Go for GitHub to fit your needs
```
aholstrup1, line 7: Could "Creating a feature suggestion in the AL-Go repo" be the first option? I guess we don't want too many people going down the path of customizing their AL-Go workflows

Scenarios/settings.md
```diff
@@ -12,10 +14,14 @@ When running a workflow or a local script, the settings are applied by reading s
 
 1. `ALGoOrgSettings` is a **GitHub variable**, which can be defined on an **organizational level** and will apply to **all AL-Go repositories** in this organization.
 
+1. `.github/templateRepoSettings.doNotEdit.json` is the organizational settings from an indirect template repository (if applicable)
```
aholstrup1, line 17:
```suggestion
1. `.github/templateRepoSettings.doNotEdit.json` is the repository settings from an indirect template repository (if applicable)
```
Repo settings I'm guessing?

e2eTests/scenarios/IndirectTemplate/runtest.ps1
```diff
@@ -0,0 +1,244 @@
...
+Pull
+
+# Make modifications to the template repository
+
+# Add Custom Jobs to CICD.yaml
+$cicdWorkflow = Join-Path $templateRepoPath '.github/workflows/CICD.yaml'
+$cicdYaml = [yaml]::Load($cicdWorkflow)
+$cicdYaml | Should -Not -BeNullOrEmpty
+
+# Modify the permissions
```
aholstrup1, line 99: Doesn't look like the E2E test modifies the permissions. I think that's a scenario we should test if we support it

Scenarios/CustomizingALGoForGitHub.md
```diff
@@ -0,0 +1,282 @@
...
+> All custom jobs will be moved to the tail of the yaml file when running Update AL-Go System Files, but dependencies to/from the custom jobs will be maintained.
+
+> [!CAUTION]
+> Custom jobs might be broken if the customized AL-Go for GitHub workflow has been refactored and the referenced jobs have been renamed.
+
+### Modifying workflow permissions
+
+If any of your custom jobs require permissions, which exceeds the permissions already assigned in the workflow, then these permissions can be specified directly on the custom job.
+
+If any of your custom steps require permissions, which exceeds the permissions already assigned in the workflow, you can modify the permissions of the workflow and assign additional permissions. AL-Go for GitHub will not allow you to remove permissions, which might be needed in other steps/jobs, but additional permissions will be included when running Update AL-Go System Files.
```
aholstrup1, line 236: Is this supported? In my repo it's just removed again 🤷‍♂️ 
image

Templates/Per Tenant Extension/.github/workflows/_BuildALGoProject.yaml
freddydk: The needsContext is needed by Cosmo to allow them to access outputs from their custom jobs in overrides inside the build

Scenarios/CustomizingALGoForGitHub.md
freddydk: very good idea

Actions/CheckForUpdates/CheckForUpdates.ps1
```diff
@@ -81,7 +91,9 @@ if ($repoSettings.templateUrl -ne $templateUrl -or $templateSha -eq '') {
     $downloadLatest = $true
 }
 
-$templateFolder = DownloadTemplateRepository -headers $headers -templateUrl $templateUrl -templateSha ([ref]$templateSha) -downloadLatest $downloadLatest
+$realTemplateFolder = $null
+$templateHeaders = GetHeadersForTemplateUrl -templateUrl $templateUrl -token $token
+$templateFolder = DownloadTemplateRepository -headers $templateHeaders -templateUrl $templateUrl -templateSha ([ref]$templateSha) -downloadLatest $downloadLatest
```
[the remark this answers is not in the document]
freddydk: Good catch

Actions/CheckForUpdates/yamlclass.ps1
```diff
@@ -194,4 +233,95 @@ class Yaml {
...
+            foreach($customJob in $customJobs) {
+                Write-Host "- $customJob"
+                $jobsWithDependency = $nativeJobs | Where-Object { $this.GetPropertyArray("jobs:/$($_):/needs:") | Where-Object { $_ -eq $customJob } }
+                # If any Build Job has a dependency on this CustomJob, add will be added to all build jobs later
+                if ($jobsWithDependency | Where-Object { $_ -like 'Build*' }) {
+                    $jobsWithDependency = @($jobsWithDependency | Where-Object { $_ -notlike 'Build*' }) + @('Build')
+                }
+                if ($jobsWithDependency) {
+                    Write-Host "  - Jobs with dependency: $($jobsWithDependency -join ', ')"
+                    $result += @(@{ "Name" = $customJob; "Content" = @($this.Get("jobs:/$($customJob):").content); "NeedsThis" = @($jobsWithDependency) })
```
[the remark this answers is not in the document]
freddydk: It is the other way around - if no job needs the customJob, it will be removed.
I have fixed this.

e2eTests/scenarios/IndirectTemplate/runtest.ps1
```diff
@@ -0,0 +1,243 @@
...
+#                                                             | |
+#                                                             |_|
+# This test tests the following scenario:
+#
+#  - Create a new repository based on the PTE template with no apps (this will be the "indirect" template repository)
+#  - Create a new repository based on the PTE template with 1 app, using compilerfolder and donotpublishapps (this will be the "final" template repository)
+#  - Run Update AL-Go System Files in final repo (using indirect repo as template)
+#  - Run Update AL-Go System files in indirect repo
+
+# TODO: describe the scenario
```
mazhelez, line 30: To be removed?

e2eTests/scenarios/IndirectTemplate/runtest.ps1
```diff
@@ -0,0 +1,243 @@
...
+Test-LogContainsFromRun -runid $run.id -jobName 'CustomJob-TemplateInit' -stepName 'Init' -expectedText 'CustomJob-TemplateInit was here!'
+Test-LogContainsFromRun -runid $run.id -jobName 'CustomJob-TemplateDeploy' -stepName 'Deploy' -expectedText 'CustomJob-TemplateDeploy was here!'
+Test-LogContainsFromRun -runid $run.id -jobName 'CustomJob-PreDeploy' -stepName 'PreDeploy' -expectedText 'CustomJob-PreDeploy was here!'
+Test-LogContainsFromRun -runid $run.id -jobName 'CustomJob-PostDeploy' -stepName 'PostDeploy' -expectedText 'CustomJob-PostDeploy was here!'
+Test-LogContainsFromRun -runid $run.id -jobName 'JustSomeJob' -stepName 'JustSomeStep' -expectedText 'JustSomeJob was here!' | Should -Throw
+Test-LogContainsFromRun -runid $run.id -jobName 'JustSomeTemplateJob' -stepName 'JustSomeTemplateStep' -expectedText 'JustSomeTemplateJob was here!' | Should -Throw
+
+Set-Location $prevLocation
+
+# TODO: Modify settings in the template repository and re-run Update AL-Go System Files in the final repository to check that the settings are copied to the final repository
```
mazhelez, line 236: Is there something missing here?

Actions/CheckForUpdates/CheckForUpdates.ps1
```diff
@@ -173,34 +228,44 @@ foreach($checkfile in $checkfiles) {
...
 
+                if ($type -eq 'workflow' -and $realSrcFile -ne $srcFile) {
+                    # Apply customizations from indirect template repository
+                    Write-Host "Apply customizations from indirect template repository: $srcFile"
+                    [Yaml]::ApplyCustomizations([ref] $srcContent, $srcFile)
+                }
+
                 if ($dstFileExists) {
+                    if ($type -eq 'workflow') {
+                        Write-Host "Apply customizations from my repository: $dstFile"
```
mazhelez, line 261:
```suggestion
                        Write-Host "Apply customizations from current repository: $dstFile"
```

Actions/CheckForUpdates/yamlclass.ps1
```diff
@@ -194,4 +233,98 @@ class Yaml {
...
+        }
+    }
+
+    static [void] ApplyCustomizations([ref] $srcContent, [string] $yamlFile) {
+        $srcYaml = [Yaml]::new($srcContent.Value.Split("`n"))
+        try {
+            $yaml = [Yaml]::Load($yamlFile)
+        }
+        catch {
+            return
```
mazhelez, line 318: Consider adding some logging here.

Actions/CheckForUpdates/CheckForUpdates.ps1
```diff
@@ -173,34 +228,44 @@ foreach($checkfile in $checkfiles) {
...
+                        $srcContent = Get-ContentLF -Path $realSrcFile
                     }
                 }
                 # Replace static placeholders
                 $srcContent = $srcContent.Replace('{TEMPLATEURL}', $templateUrl)
 
-                if ($isDirectALGo) {
-                    # If we are using direct AL-Go repo, we need to change the owner to the templateOwner, the repo names to AL-Go and AL-Go/Actions and the branch to templateBranch
+                if ($isFileDirectALGo) {
+                    # If we are using direct AL-Go repo, we need to change the owner to the remplateOwner, the repo names to AL-Go and AL-Go/Actions and the branch to templateBranch
```
mazhelez, line 249:
```suggestion
                    # If we are using direct AL-Go repo, we need to change the owner to the templateOwner, the repo names to AL-Go and AL-Go/Actions and the branch to templateBranch
```

Templates/Per Tenant Extension/.github/workflows/_BuildALGoProject.yaml
freddydk: env's are not transferred from one job to another.
The only way you can transfer stuff from jobs is output.

RELEASENOTES.md
```diff
@@ -15,6 +15,28 @@ As stated in [AL-Go Deprecations](https://aka.ms/algodeprecations#cleanModePrepr
...
+  - **AuthTokenSecret** (optional) - the name of a secret (like ghTokenWorkflow) which contains a Personal Access Token or GitHub App specification with access to the source files.
+
+### Add custom jobs to AL-Go workflows
+
+It is now possible to add custom jobs to AL-Go workflows. The Custom Job needs to be named `CustomJob<something>` and should be placed after all other jobs in the .yaml file. The order of which jobs are executed is determined by the Needs statements. Your custom job will be executed after all jobs specified in the Needs clause in your job and if you need the job to be executed before other jobs, you should add the job name in the Needs clause of that job. See [https://aka.ms/algosettings#customjobs](https://aka.ms/algosettings#customjobs) for details.
+
+Note that custom jobs might break by future changes to AL-Go for GitHub workflows. If you have customizations to AL-Go for GitHub workflows, you should always doublecheck the pull request generated by Update AL-Go System Files.
+
+### Indirect AL-Go template repository
```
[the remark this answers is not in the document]
freddydk: I have renamed everything to Custom Templates

Actions/CheckForUpdates/CheckForUpdates.ps1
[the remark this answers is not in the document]
freddydk: This should be fixed now

Scenarios/CustomizingALGoForGitHub.md
```diff
@@ -0,0 +1,301 @@
...
+
+If AL-Go functionality out-of-the-box doesn't match your needs, you really have three options:
+
+1. Customize AL-Go for GitHub to fit your needs
+1. Select another managed DevOps solution
+1. Create your own DevOps solution from scratch (not recommended)
+
+Creating your own DevOps solution from scratch requires dedicated resources to develop and maintain workflows, processes etc. **This is not a small task**. There are many moving parts in a DevOps solution, which might require you to make changes to workflows and scripts over time and stay secure and having to maintain many repositories is tedious and time consuming, even when using templates and other advanced features.
+
+Microsoft will continuously develop and maintain AL-Go for GitHub and ensure that we always use the latest versions of GitHub actions, which are under our control. Microsoft will never add dependencies to any third party GitHub action, which are not under our control.
```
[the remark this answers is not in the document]
freddydk: added an if microsoft to this

Scenarios/CustomizingALGoForGitHub.md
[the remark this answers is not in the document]
freddydk: should work now

Actions/CheckForUpdates/CheckForUpdates.ps1
```diff
@@ -90,11 +82,38 @@ $templateInfo = "$templateOwner/$($templateUrl.Split('/')[4])"
...
+    $templateRepoSettingsFile = Join-Path $templateFolder "*/$RepoSettingsFile"
+    if (Test-Path -Path $templateRepoSettingsFile -PathType Leaf) {
+        $templateRepoSettings = Get-Content $templateRepoSettingsFile -Encoding UTF8 | ConvertFrom-Json | ConvertTo-HashTable -Recurse
         if ($templateRepoSettings.Keys -contains "templateUrl" -and $templateRepoSettings.templateUrl -ne $templateUrl) {
-            throw "The specified template repository is not a template repository, but instead another AL-Go repository. This is not supported."
+            # The template repository is a url to another AL-Go repository (a custom template repository)
+            # TemplateUrl and TemplateSha from .github/AL-Go-Settings.json in the custom template repository points to the "real" template repository
+            # Copy files and folders from the custom template repository, but grab the unmodified file from the "real" template repository if it exists and apply customizations
+            # Copy .github/AL-Go-Settings.json to .github/templateRepoSettings.json (will be read before .github/AL-Go-Settings.json in the final repo)
```
mazhelez: Minor: the files are called something else now.

Actions/CheckForUpdates/CheckForUpdates.ps1
```diff
@@ -81,7 +72,8 @@ if ($repoSettings.templateUrl -ne $templateUrl -or $templateSha -eq '') {
     $downloadLatest = $true
 }
 
-$templateFolder = DownloadTemplateRepository -headers $headers -templateUrl $templateUrl -templateSha ([ref]$templateSha) -downloadLatest $downloadLatest
+$realTemplateFolder = $null
```
mazhelez: Minor: naming. 

How about `originalTemplateFolder`?

Actions/CheckForUpdates/CheckForUpdates.ps1
```diff
@@ -149,20 +176,48 @@ foreach($checkfile in $checkfiles) {
     $srcPath = $checkfile.srcPath
     $dstPath = $checkfile.dstPath
     $dstFolder = Join-Path $baseFolder $dstPath
-    $srcFolder = GetSrcFolder -repoSettings $repoSettings -templateUrl $templateUrl -templateFolder $templateFolder -srcPath $srcPath
+    $srcFolder = GetSrcFolder -repoType $repoSettings.type -templateUrl $templateUrl -templateFolder $templateFolder -srcPath $srcPath
+    $realSrcFolder = $null
+    if ($realTemplateFolder -and $type -notlike 'template*settings') {
+        # Get Real source folder except for template settings - these are applied from the custom temoplate´repository
```
mazhelez:
```suggestion
        # Get Real source folder except for template settings - these are applied from the custom template repository
```

Actions/CheckForUpdates/CheckForUpdates.HelperFunctions.ps1
```diff
@@ -12,12 +12,33 @@ If true, the latest SHA of the template repository will be downloaded
...
+    }
+    catch {
+        # Ignore error
+        $response = $null
+    }
+    if (-not $response -or $response.StatusCode -ne 200) {
+        # GITHUB_TOKEN doesn't have access to template repository, must be private/internal
+
+``
+        # Get token with read permissions for the template repository
```
aholstrup1:
```suggestion
        # GITHUB_TOKEN doesn't have access to template repository, must be private/internal
        # Get token with read permissions for the template repository
```

Templates/Per Tenant Extension/.github/workflows/_BuildALGoProject.yaml
mazhelez: I meant to set it as env to the whole job, not just the step.

Alternatively, jobs can also share info via artifacts.

If you decide to keep it like this, let's at least come up with a better name. `CustomInput`? Or `ExternalInput`? Or simply `JobInput`?

By the way, if the input is passed like that, the other inputs from `Initialization` are not needed anymore.

RELEASENOTES.md
```diff
@@ -15,6 +15,19 @@ As stated in [AL-Go Deprecations](https://aka.ms/algodeprecations#cleanModePrepr
 
 Some workflow inputs are now validated early in order to avoid workflows to make modifications like creating a release, when we already should know that an error will occur later.
 
+### Add custom jobs to AL-Go workflows
```
mazhelez: Make sure to move this apart above the latest released version.

Templates/Per Tenant Extension/.github/workflows/_BuildALGoProject.yaml
freddydk: GitHub calls it Needs Context itself - https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#needs-context - renaming it might only confuse things IMO

The reason for the location here is that this is the only location where you can have overrides, which can use the outputs from your custom jobs. Putting the NeedsContext on a higher level just means the we will carry around this in a lot of steps, which doesn't need it.

And you are right - we could eliminate other inputs, but I like to have explicit references to things instead of de-reffing a json structure in a lot of places.

The NeedsContext was originally suggested by Cosmo to allow them to make their integration and IMO it is the least intrusive way to transfer values from custom jobs to script overrides in the build step.

Actions/CheckForUpdates/CheckForUpdates.ps1
freddydk: correct

Actions/CheckForUpdates/CheckForUpdates.ps1
freddydk: Agreed, that is a better name

RELEASENOTES.md
freddydk: It is - and I always do that as part of the release process - the last step of releasing a new version is to go through open PR's and modify target version

freddydk: Forgot to update branch - I see what you mean

Actions/CheckForUpdates/CheckForUpdates.HelperFunctions.ps1
```diff
@@ -12,12 +12,31 @@ If true, the latest SHA of the template repository will be downloaded
...
         [string] $templateUrl,
         [ref] $templateSha,
         [bool] $downloadLatest
     )
 
+    $templateRepositoryUrl = $templateUrl.Split('@')[0]
+    $templateRepository = $templateRepositoryUrl.Split('/')[-2..-1] -join '/'
+
+    # Use Authenticated API request if possible to avoid the 60 API calls per hour limit
+    $headers = GetHeaders -token $env:GITHUB_TOKEN -repository $templateRepository
```
mazhelez, line 25: Shouldn't this be `$token`?

```suggestion
    $headers = GetHeaders -token $token -repository $templateRepository

```

mazhelez: Ah, no, it doesn't. But I still don't get why the call with `$env:GITHUB_TOKEN` is needed, when there is `$token` parameter.

Actions/CheckForUpdates/CheckForUpdates.ps1
```diff
@@ -90,11 +82,38 @@ $templateInfo = "$templateOwner/$($templateUrl.Split('/')[4])"
...
 if (-not $isDirectALGo) {
-    $ALGoSettingsFile = Join-Path $templateFolder "*/$repoSettingsFile"
-    if (Test-Path -Path $ALGoSettingsFile -PathType Leaf) {
-        $templateRepoSettings = Get-Content $ALGoSettingsFile -Encoding UTF8 | ConvertFrom-Json | ConvertTo-HashTable -Recurse
+    $templateRepoSettingsFile = Join-Path $templateFolder "*/$RepoSettingsFile"
+    if (Test-Path -Path $templateRepoSettingsFile -PathType Leaf) {
+        $templateRepoSettings = Get-Content $templateRepoSettingsFile -Encoding UTF8 | ConvertFrom-Json | ConvertTo-HashTable -Recurse
         if ($templateRepoSettings.Keys -contains "templateUrl" -and $templateRepoSettings.templateUrl -ne $templateUrl) {
-            throw "The specified template repository is not a template repository, but instead another AL-Go repository. This is not supported."
+            # The template repository is a url to another AL-Go repository (a custom template repository)
```
mazhelez, line 83: Telemetry would be useful when custom template is being used.

Actions/CheckForUpdates/CheckForUpdates.ps1
```diff
@@ -109,21 +128,29 @@ if (-not $isDirectALGo) {
 # - All files in .github that ends with .copy.md
 # - All PowerShell scripts in .AL-Go folders (all projects)
 $checkfiles = @(
-    @{ 'dstPath' = Join-Path '.github' 'workflows'; 'srcPath' = Join-Path '.github' 'workflows'; 'pattern' = '*'; 'type' = 'workflow' },
-    @{ 'dstPath' = '.github'; 'srcPath' = '.github'; 'pattern' = '*.copy.md'; 'type' = 'releasenotes' },
-    @{ 'dstPath' = '.github'; 'srcPath' = '.github'; 'pattern' = 'AL-Go-Settings.json'; 'type' = 'settings' },
-    @{ 'dstPath' = '.github'; 'srcPath' = '.github'; 'pattern' = '*.settings.json'; 'type' = 'settings' }
+    @{ 'dstPath' = (Join-Path '.github' 'workflows'); 'newname' = ''; 'srcPath' = (Join-Path '.github' 'workflows'); 'pattern' = '*'; 'type' = 'workflow' },
```
mazhelez: Minor: naming.

How about `dstName` to match the current patter ("src" for source and "dst" for destination)?

Actions/CheckForUpdates/yamlclass.ps1
```diff
@@ -194,4 +233,99 @@ class Yaml {
...
+            $yaml = [Yaml]::Load($yamlFile)
+        }
+        catch {
+            Write-Host "Unable to read YAML file $yamlFile. Skipping custom jobs."
+            return
+        }
+
+        # Locate custom jobs in destination YAML
+        Write-Host "Apply custom jobs"
+        $customJobs = @($yaml.GetCustomJobsFromYaml('CustomJob*'))
```
mazhelez, line 324: Minor: convention

Why does the name have to match "CustomJob*"? Aren't all jobs that are not in the original template custom jobs?
With the exception of the additional build jobs (if any), "Build1", "Build2", etc.

RELEASENOTES.md
```diff
@@ -1,6 +1,19 @@
...
+
+Note that custom jobs might break by future changes to AL-Go for GitHub workflows. If you have customizations to AL-Go for GitHub workflows, you should always doublecheck the pull request generated by Update AL-Go System Files.
+
+### Support for Custom AL-Go template repositories
+
+Create an AL-Go for GitHub repository based on [https://aka.ms/algopte](https://aka.ms/algopte) or [https://aka.ms/algoappsource](https://aka.ms/algoappsource), add custom workflows, custom jobs, custom steps and/or settings to this repository and then use that repository as the template repository for other repositories. Using custom template repositories allows you to create and use highly customized template repositories and control the uptake of this in all repositories. See [https://aka.ms/algosettings#customtemplate](https://aka.ms/algosettings#customtemplate) for details.
+
+> [!NOTE]
+> Customized repositories might break by future changes to AL-Go for GitHub. If you are customizing AL-Go for GitHub, you should always doublecheck the pull request when updating AL-Go system files in your custom template repositories.
```
mazhelez:
```suggestion
> Customized repositories might break by future changes to AL-Go for GitHub. If you are customizing AL-Go for GitHub, you should always double-check the pull request when updating AL-Go system files in your custom template repositories.
```

Templates/AppSource App/.AL-Go/cloudDevEnv.ps1
```diff
@@ -42,7 +51,7 @@ Write-Host -ForegroundColor Yellow @'
 
 $tmpFolder = Join-Path ([System.IO.Path]::GetTempPath()) "$([Guid]::NewGuid().ToString())"
 New-Item -Path $tmpFolder -ItemType Directory -Force | Out-Null
-$GitHubHelperPath = DownloadHelperFile -url 'https://raw.githubusercontent.com/microsoft/AL-Go-Actions/main/Github-Helper.psm1' -folder $tmpFolder
+$GitHubHelperPath = DownloadHelperFile -url 'https://raw.githubusercontent.com/microsoft/AL-Go-Actions/main/Github-Helper.psm1' -folder $tmpFolder -first
```
mazhelez: What scenario is covered here? Should `-first` be added to the other `DownloadHelperFile` call?

Actions/CheckForUpdates/CheckForUpdates.ps1
```diff
@@ -45,19 +45,10 @@ if ($token) {
...
-    # GITHUB_TOKEN doesn't have access to template repository, must be is private/internal
-    # Get token with read permissions for the template repository
-    # NOTE that the GitHub app needs to be installed in the template repository for this to work
-    $templateRepository = $templateRepositoryUrl.Split('/')[-2..-1] -join '/'
-    $templateReadToken = GetAccessToken -token $token -permissions @{"actions"="read";"contents"="read";"metadata"="read"} -repository $templateRepository
-
-    # Use read token for authenticated API request
-    $headers = GetHeaders -token $templateReadToken
+# Get secrets for Custom AL-Go System Files
+$secrets = @{}
```
mazhelez: Why the secrets needed here?

Templates/AppSource App/.github/workflows/CICD.yaml
```diff
@@ -166,6 +166,7 @@ jobs:
         uses: microsoft/AL-Go-Actions/CheckForUpdates@main
         env:
           GITHUB_TOKEN: ${{ github.token }}
+          Secrets: '${{ steps.ReadSecrets.outputs.Secrets }}'
```
mazhelez: Isn't only `ghTokenWorkflow` read from the `ReadSecrets` step? And it's already passed in the `token` parameter.

Scenarios/settings.md
```diff
@@ -347,12 +379,67 @@ Settings, which might be relevant to set in the settings file includes
...
+
+This would make sure that all repositories would have this script (and updated versions of the script) in the future.
+
+The items, which are currently supported from custom template repositories are:
+
+- Repository script overrides in the .github folder
+- Project script overrides in the .AL-Go folder
+- Custom workflows in the .github/workflows folder
+- Custom jobs in any AL-Go for GitHub workflow
+- Custom steps in the \_BuildALGoProject workflow
```
mazhelez: Not yet.

Scenarios/settings.md
```diff
@@ -347,12 +379,67 @@ Settings, which might be relevant to set in the settings file includes
...
+
+Custom jobs will be preserved when running Update AL-Go System Files.
+
+**Note** that installing [apps from the GitHub marketplace](https://github.com/marketplace?type=apps) might require you to add custom jobs or steps to some of the workflows to get the right integration. In custom jobs and custom steps, you can use any [actions from the GitHub marketplace](https://github.com/marketplace?type=actions).
+
+<a id="customtemplate"></a>
+
+## Custom template repositories
+
+If you are utilizing script overrides, custom jobs, custom steps, custom delivery or like in many repositories, you might want to take advantage of the custom template repository feature.
```
mazhelez: There are no "custom steps", right?

Workshop/Index.md
```diff
@@ -22,6 +22,13 @@ This workshop shows you how to take advantage of the functionality, which is pro
 1. [The Development Process](TheDevelopmentProcess.md) - *FUTURE TOPIC: The recommended way to work with feature branches, pull requests, code reviews and branch protection rules.*
 1. [Keeping your Repository Up-to-date](KeepUpToDate.md) - *FUTURE TOPIC: Updating AL-Go for GitHub to the latest version by running a workflow.*
 
+## Expert level
+
+1. [Custom Delivery](CustomDelivery.md) - *FUTURE TOPIC: Setting up custom delivery to f.ex. a Teams channel.*
+1. [Custom Delivery](CustomDeployment.md) - *FUTURE TOPIC: Setting up custom deployment to f.ex. an on-premises environment.*
```
mazhelez:
```suggestion
1. [Custom Deployment](CustomDeployment.md) - *FUTURE TOPIC: Setting up custom deployment to f.ex. an on-premises environment.*

```

Scenarios/CustomizingALGoForGitHub.md
```diff
@@ -0,0 +1,225 @@
...
+      ...
+```
+
+Adding a custom job like this, will cause this job to run simultaneously with the deploy and the deliver jobs.
+
+> [!NOTE]
+> All custom jobs will be moved to the tail of the yaml file when running Update AL-Go System Files, but dependencies to/from the custom jobs will be maintained.
+
+> [!CAUTION]
+> Custom jobs might be broken if the customized AL-Go for GitHub workflow has been refactored and the referenced jobs have been renamed.
```
aholstrup1: Minor suggestion
```suggestion
> Custom jobs might be broken if the customized AL-Go for GitHub workflow has been refactored and the referenced jobs have been renamed. Therefore, please make sure to review the changes in AL-Go workflows when running Update AL-Go System Files.
```

Scenarios/settings.md
freddydk: Correct

Actions/CheckForUpdates/CheckForUpdates.ps1
freddydk: They are not

Actions/CheckForUpdates/yamlclass.ps1
freddydk: If we don't keep the CustomJob prefix, chances are that people will name their custom jobs like jobs we might add or rename in the future.
Prefixing makes this very clear and has no disadvantages - you can always add a display name to your job.

Templates/AppSource App/.github/workflows/CICD.yaml
freddydk: Correct - secrets was only used for custom files

Templates/AppSource App/.AL-Go/cloudDevEnv.ps1
freddydk: Renamed the parameter to notifyAuthenticatedAttempt, which is what it does.
If the template cannot be reached, a warning is written that authenticated request is attempted
The subsequent calls will follow the same pattern and shouldn't all display that authenticated request is attempted.

Actions/CheckForUpdates/CheckForUpdates.ps1
freddydk: agree

Actions/CheckForUpdates/CheckForUpdates.HelperFunctions.ps1
freddydk: The issue here is that if we use token and token is f.ex. expired, then an expired token cannot read from a public repo - and will fail on something that shouldn't fail.
It should fail when trying to update or read a private repo - not microsofts public repo.

Actions/CheckForUpdates/CheckForUpdates.ps1
freddydk: agree

Actions/CheckForUpdates/CheckForUpdates.ps1
```diff
@@ -87,6 +81,8 @@ if (-not $isDirectALGo) {
         $templateRepoSettings = Get-Content $templateRepoSettingsFile -Encoding UTF8 | ConvertFrom-Json | ConvertTo-HashTable -Recurse
         if ($templateRepoSettings.Keys -contains "templateUrl" -and $templateRepoSettings.templateUrl -ne $templateUrl) {
             # The template repository is a url to another AL-Go repository (a custom template repository)
+            Trace-Information -Message "Using custom AL-Go template repository: $($templateRepoSettings.templateUrl)"
```
aholstrup1:
```suggestion
            Trace-Information -Message "Using custom AL-Go template repository"
```
For privacy reasons I think we'd be better off not logging the name of the template repository.

PR #877, opened 2024-01-18
Support release branches that start with "releases/"

mazhelez:
Support release branches that start with "releases/"

- [x] Propagate workflow changes
- [x] Add release notes

### DIFF
--- a/Actions/Github-Helper.psm1
+++ b/Actions/Github-Helper.psm1
@@ -533,7 +533,9 @@ function GetLatestRelease {
     # Get Latest release
     $latestRelease = $releases | Where-Object { -not ($_.prerelease -or $_.draft) } | Select-Object -First 1
     $releaseBranchesFormats = 'release/*', 'releases/*'
-    if ($releaseBranchesFormats | Where-Object { $ref -like $_ }) {
+    $isReleaseBranch = [boolean] $($releaseBranchesFormats | Where-Object { $ref -like $_ })
+
+    if ($isReleaseBranch) {
         # If release branch, get the latest release from that the release branch
         # This is given by the latest release with the same major.minor as the release branch
         $semVerObj = SemVerStrToSemVerObj -semVerStr $ref.SubString($releaseBranchPrefix.Length) -allowMajorMinorOnly
--- a/Actions/Github-Helper.psm1
+++ b/Actions/Github-Helper.psm1
@@ -532,8 +532,8 @@ function GetLatestRelease {
 
     # Get Latest release
     $latestRelease = $releases | Where-Object { -not ($_.prerelease -or $_.draft) } | Select-Object -First 1
-    $releaseBranchPrefix = 'release/'
-    if ($ref -like "$releaseBranchPrefix*") {
+    $releaseBranchesFormats = 'release/*', 'releases/*'
+    if ($releaseBranchesFormats | Where-Object { $ref -like $_ }) {
         # If release branch, get the latest release from that the release branch
         # This is given by the latest release with the same major.minor as the release branch
         $semVerObj = SemVerStrToSemVerObj -semVerStr $ref.SubString($releaseBranchPrefix.Length) -allowMajorMinorOnly
--- a/Actions/ReadSettings/ReadSettings.ps1
+++ b/Actions/ReadSettings/ReadSettings.ps1
@@ -57,7 +57,7 @@ $settings.Keys | ForEach-Object {
     }
     $outSettings += @{ "$setting" = $settingValue }
     if ($getSettings -contains $setting) {
-        if ($settingValue -is [System.Collections.Specialized.OrderedDictionary] -or $settingValue -is [hashtable]) {
+        if ($settingValue -is [System.Collections.Specialized.OrderedDictionary] -or $settingValue -is [hashtable] -or $settingValue -is [array]) {
             Add-Content -Encoding UTF8 -Path $env:GITHUB_ENV -Value "$setting=$(ConvertTo-Json $settingValue -Depth 99 -Compress)"
         }
         else {
--- a/Actions/ReadSettings/ReadSettings.ps1
+++ b/Actions/ReadSettings/ReadSettings.ps1
@@ -66,6 +66,8 @@ $settings.Keys | ForEach-Object {
     }
 }
 
+# Add default values for settings that are not set
+# CICDPushBranches is not present in the settings object, although it can be set there.
 if('CICDPushBranches' -in $getSettings -and 'CICDPushBranches' -notin $settings.Keys) {
     Add-Content -Encoding UTF8 -Path $env:GITHUB_ENV -Value "CICDPushBranches=$(Convert-ToJson $defaultCICDPushBranches -Depth 99 -Compress))"
 }
--- a/Actions/ReadSettings/ReadSettings.ps1
+++ b/Actions/ReadSettings/ReadSettings.ps1
@@ -57,19 +57,13 @@ $settings.Keys | ForEach-Object {
     }
     $outSettings += @{ "$setting" = $settingValue }
     if ($getSettings -contains $setting) {
-        if ($settingValue -is [System.Collections.Specialized.OrderedDictionary] -or $settingValue -is [hashtable] -or $settingValue -is [array]) {
+        if ($settingValue -is [System.Collections.Specialized.OrderedDictionary] -or $settingValue -is [hashtable]) {
             Add-Content -Encoding UTF8 -Path $env:GITHUB_ENV -Value "$setting=$(ConvertTo-Json $settingValue -Depth 99 -Compress)"
         }
         else {
             Add-Content -Encoding UTF8 -Path $env:GITHUB_ENV -Value "$setting=$settingValue"
         }
     }
-}
-
-# Add default values for settings that are not set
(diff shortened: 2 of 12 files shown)


Actions/Github-Helper.psm1
```diff
@@ -532,8 +532,8 @@ function GetLatestRelease {
 
     # Get Latest release
     $latestRelease = $releases | Where-Object { -not ($_.prerelease -or $_.draft) } | Select-Object -First 1
-    $releaseBranchPrefix = 'release/'
-    if ($ref -like "$releaseBranchPrefix*") {
+    $releaseBranchesFormats = 'release/*', 'releases/*'
+    if ($releaseBranchesFormats | Where-Object { $ref -like $_ }) {
```
mazhelez, line 536: Separate variable `$isReleaseBranch`?

Actions/ReadSettings/ReadSettings.ps1
```diff
@@ -66,6 +66,10 @@ $settings.Keys | ForEach-Object {
     }
 }
 
+if('CICDPushBranches' -in $getSettings -and 'CICDPushBranches' -notin $settings.Keys) {
```
mazhelez, line 69: This is done because `CICDPushBranches` is not in the settings by default. Add a comment to explain it.

Templates/Per Tenant Extension/.github/workflows/CICD.yaml
```diff
@@ -113,6 +114,25 @@ jobs:
           getEnvironments: '*'
           type: 'CD'
 
+      - name: Determine Publish Artifacts
+        id: DeterminePublishArtifacts
+        run: |
+          $publishArtifacts = $false
```
mazhelez, line 120: Avoid scripts directly in the YAML.
Is this worth an action though?

Templates/Per Tenant Extension/.github/workflows/CreateRelease.yaml
```diff
@@ -27,6 +27,10 @@ on:
         description: Create Release Branch?
         type: boolean
         default: false
+      releaseBranchPrefix:
+        description: The preffix for the release branch. Used only if 'Create Release Branch?' is checked.
```
aholstrup1, line 31:
```suggestion
        description: The prefix for the release branch. Used only if 'Create Release Branch?' is checked.
```

Templates/Per Tenant Extension/.github/workflows/CICD.yaml
```diff
@@ -63,7 +64,7 @@ jobs:
         uses: microsoft/AL-Go-Actions/ReadSettings@main
         with:
           shell: powershell
-          get: type
+          get: type, CICDPushBranches
```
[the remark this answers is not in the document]
mazhelez: Yes, `ReadSettings` outputs an env variable with all settings, but `CICDPushBranches` might not be included.

Templates/Per Tenant Extension/.github/workflows/CICD.yaml
```diff
@@ -153,7 +176,7 @@ jobs:
       projectDependenciesJson: ${{ needs.Initialization.outputs.projectDependenciesJson }}
       secrets: 'licenseFileUrl,codeSignCertificateUrl,*codeSignCertificatePassword,keyVaultCertificateUrl,*keyVaultCertificatePassword,keyVaultClientId,gitHubPackagesContext,applicationInsightsConnectionString'
       publishThisBuildArtifacts: ${{ needs.Initialization.outputs.workflowDepth > 1 }}
-      publishArtifacts: ${{ github.ref_name == 'main' || startswith(github.ref_name, 'release/') || needs.Initialization.outputs.deliveryTargetsJson != '[]' || needs.Initialization.outputs.environmentCount > 0 }}
+      publishArtifacts: ${{ needs.Initialization.outputs.PublishArtifacts == 'True' }}
```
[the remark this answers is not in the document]
mazhelez: > What's a scenario where you run CICD where you don't want to publish the artifacts?

When the CICD is run on a branch that's not _main_ or a release branch.

> Could publish artifacts just be true always for CICD?

Changing it to _true_ always might be breaking (haven't tested, though). If CICD is run on a dev branch, then the dev artifacts might potentially get deployed/delivered.

@freddydk any input here?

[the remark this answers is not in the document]
aholstrup1: > Changing it to true always might be breaking (haven't tested, though). If CICD is run on a dev branch, then the dev artifacts might potentially get deployed/delivered.

To be fair the D in CICD does stand for deploy/deliver 😄 If you don't want to deploy / deliver should you really be running the CIC**D** workflow? 

If you just want to test that your branch builds, and the tests pass shouldn't you be running a PR Build?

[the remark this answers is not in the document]
freddydk: Earlier, artifacts was published for main + release branches + if any delivery or deployment was true.
At the same time CICDPushBranches was also main and release - but many partners are changing this to run CI/CD on dev branches and they could even remove release branches if they decide to run CI/CD manually.
If they do this - then they cannot create hotfixes anymore after this change - or?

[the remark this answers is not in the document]
freddydk: Since the branch name is in the artifacts pushed - we could decide to push always - which would then remove the need for the thisbuild artifacts (I guess) - I need to think whether there are any unwanted consequences with this...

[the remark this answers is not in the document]
mazhelez: CICD on dev branches sounds wild. 

I think the Deploy/Deliver part won't succeed at the moment as there will be no artifacts when CICD is run on a dev branch.
But if we change it to _always_ produce artifacts, then wouldn't Deploy/Deliver pick them up, potentially deploying/delivering artifacts from dev branches?

Templates/Per Tenant Extension/.github/workflows/CICD.yaml
mazhelez: Not relevant anymore.

Templates/Per Tenant Extension/.github/workflows/CICD.yaml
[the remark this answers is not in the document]
mazhelez: Resorted to just add the `startswith(github.ref_name, 'releases/')` check. This shouldn't break existing scenarios.

Templates/Per Tenant Extension/.github/workflows/CICD.yaml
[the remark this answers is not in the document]
mazhelez: Resorted to not using `CICDPushBranches` at all. See other conversation for the potential scenarios that could be broken.

Templates/AppSource App/.github/workflows/CreateRelease.yaml
```diff
@@ -304,11 +308,13 @@ jobs:
       - name: Create Release Branch
         run: |
           $errorActionPreference = "Stop"; $ProgressPreference = "SilentlyContinue"; Set-StrictMode -Version 2.0
-          git checkout -b ${{ needs.CreateRelease.outputs.releaseBranch }}
+          $releaseBranch = '${{ github.event.inputs.releaseBranchPrefix }}${{ needs.CreateRelease.outputs.releaseVersion }}'
```
freddydk, line 311: Isn't there a security risk when we use inputs inline in code?
I think this should be converted to env and then used as env

freddydk (changes requested):

Templates/AppSource App/.github/workflows/CreateRelease.yaml
mazhelez: Not sure what the risk is here. The workflow has only the `workfow_dispatch` trigger.

Anyway, I extracted an env variable.

Templates/Per Tenant Extension/.github/workflows/CreateRelease.yaml
```diff
@@ -302,13 +306,17 @@ jobs:
           ref: '${{ needs.createRelease.outputs.commitish }}'
 
       - name: Create Release Branch
+        env:
+          releaseBranchPrefix: ${{ github.event.inputs.releaseBranchPrefix }}
         run: |
           $errorActionPreference = "Stop"; $ProgressPreference = "SilentlyContinue"; Set-StrictMode -Version 2.0
-          git checkout -b ${{ needs.CreateRelease.outputs.releaseBranch }}
+          $releaseBranch = "$($env:releaseBranchPrefix)" + "${{ needs.CreateRelease.outputs.releaseVersion }}"
```
freddydk, line 313: It doesn't need to be in " - could just be $env:xxx

freddydk (approved):

aholstrup1 (approved):

Templates/Per Tenant Extension/.github/workflows/CreateRelease.yaml
mazhelez: Without the `+` it was a bit unreadable. And the quotes `"` cover the case when, for whatever reason, `$($env:releaseBranchPrefix)` is a number.

freddydk: Environment variables are not type strong - they are always strings.

PR #884, opened 2024-01-23
Enable ability to setup custom build modes

aholstrup1:
Enable ability to setup custom build modes by not validating the build mode is either Default, Translated or Clean.

Useful in cases like https://github.com/microsoft/BCApps/pull/502 where we need to add another build mode (Strict / StrictMode)


mazhelez (dismissed):
Maybe add a line in the release notes as well.

mazhelez (approved):

freddydk (approved):

PR #909, opened 2024-01-31
Fix bug with new Power Platform solution model

andersgMSFT:
Power Platform recently rolled out some changes to the solutions model and can bring the solution manifest in a broken state. (The solution manifest includes a reference to a newly added project file, but the packaging tools does not pack it) 

Override the PAC version to mitigate the issue until a fix is rolled out for the GitHub Actions


Actions/BuildPowerPlatform/action.yaml
```diff
@@ -71,7 +74,7 @@ runs:
         ${{ github.action_path }}\updateEnvironmentAndCompany.ps1 -companyId $ENV:_companyId -environmentName $ENV:_environmentName -solutionFolder $ENV:_solutionFolder
     
     - name: Test Pack (Not real - just workaround for https://github.com/microsoft/powerplatform-vscode/issues/412)
-      uses: microsoft/powerplatform-actions/pack-solution@v0
+      uses: microsoft/powerplatform-actions/pack-solution@v1
```
andersgMSFT, line 77: @freddydk Didn't we merger this into the PPPreview branch in the last update?

freddydk (approved):

Actions/BuildPowerPlatform/action.yaml
freddydk: I cannot remember that, sorry

PR #910, opened 2024-01-31
Improve IncrementVersionNumber action

### LINKED ISSUE #889: UpdateVersionNumber in CreateRelease fails if appFolders contain wild characters
See: https://github.com/mazhelez/BCApps/actions/runs/7583563656/job/20656003973#logs

```
Reading settings from build\projects\System Application Modules\.AL-Go\settings.json
projectName       : System Application Modules
appFolders        : {..\..\..\src\System Application\App\*}
testFolders       : {..\..\..\src\System Application\Test\*, ..\..\..\src\System Application\Test Library\*}
useCompilerFolder : True
doNotPublishApps  : True
rulesetFile       : ..\..\..\src\rulesets\internal.module.ruleset.json
repoVersion       : 25.0

{
  "projectName": "System Application Modules",
  "appFolders": [
    "..\\..\\..\\src\\System Application\\App\\*"
  ],
  "testFolders": [
    "..\\..\\..\\src\\System Application\\Test\\*",
    "..\\..\\..\\src\\System Application\\Test Library\\*"
  ],
  "useCompilerFolder": true,
  "doNotPublishApps": true,
  "rulesetFile": "..\\..\\..\\src\\rulesets\\internal.module.ruleset.json",
  "repoVersion": "25.0"
}
Modifying app.json in folder build\projects\System Application Modules\..\..\..\src\System Application\App\*
AL-Go action ran: IncrementVersionNumber Telemetry Correlation Id: a1a1d53d-420a-444a-96c5-5f0621c2293d
Error: Unexpected error when running action. Error Message: Application manifest file(build\projects\System Application Modules\..\..\..\src\System Application\App\*\app.json) is malformed., StackTrace: at <ScriptBlock>, D:\a\_actions\mazhelez\AL-Go\fix-hardcoded-release\Actions\IncrementVersionNumber\IncrementVersionNumber.ps1: line 96 <- at <ScriptBlock>, D:\a\_actions\mazhelez\AL-Go\fix-hardcoded-release\Actions\IncrementVersionNumber\IncrementVersionNumber.ps1: line 76 <- at <ScriptBlock>, D:\a\_temp\08513[331](https://github.com/mazhelez/BCApps/actions/runs/7583563656/job/20656003973#step:4:333)-d23d-4506-864d-d6be71137b10.ps1: line 4 <- at <ScriptBlock>, <No file>: line 1
```

mazhelez:
Fixes #889 : `appFolders` with wild characters are not resolved when reading the app.json files.
Also fixes issue that the `version` property in `dependencies` array in the app.json is not changed.

- [x] Release notes
- [x] More tests


Actions/IncrementVersionNumber/IncrementVersionNumber.psm1
```diff
@@ -0,0 +1,94 @@
...
+    .Parameter settingsFilePath
+        Path to the settings file.
+    .Parameter settingName
+        Name of the setting to change.
+    .Parameter newValue
+        New value of the setting.
+    .Parameter incremental
+        If set, the new value will be added to the old value. The old value must be a version number.
+#>
+function Set-SettingInFile($settingsFilePath, $settingName, $newValue, [switch] $incremental) {
```
mazhelez, line 13: These are not specific for the action. Move to a common helper module? Or AL-Go-Helper.ps1.

Actions/IncrementVersionNumber/IncrementVersionNumber.psm1
```diff
@@ -0,0 +1,94 @@
...
+    Write-Host "Folders: $folders"
+
+    $folders | ForEach-Object {
+        $folder = $_
+        $folder = Join-Path $projectPath $folder
+        Write-Host "Modifying app.json in folder $folder"
+
+        $appJsonFile = Join-Path $folder "app.json"
+        try {
+            $appJson = Get-Content $appJsonFile -Encoding UTF8 | ConvertFrom-Json
```
mazhelez, line 72: `Set-SettingInFile` should be re-used around here.

Actions/IncrementVersionNumber/IncrementVersionNumber.psm1
```diff
@@ -0,0 +1,94 @@
...
+        If set, the new version number will be added to the old version number.
+#>
+function Set-ProjectVersion($projectPath, $projectSettings, $newVersion, [switch] $incremental) {
+    $projectSettingsPath = Join-Path $projectPath $ALGoSettingsFile
+    Set-SettingInFile -settingsFilePath $projectSettingsPath -settingName 'repoVersion' -newValue $newVersion -incremental:$incremental | Out-Null
+
+    # Check if the project uses repoVersion versioning strategy
+    $useRepoVersion = (($projectSettings.PSObject.Properties.Name -eq "versioningStrategy") -and (($projectSettings.versioningStrategy -band 16) -eq 16))
+
+    $folders = @($projectSettings.appFolders) + @($projectSettings.testFolders)
```
mazhelez, line 61: `bcptTestFolders` is missing here.

Actions/IncrementVersionNumber/IncrementVersionNumber.psm1
```diff
@@ -0,0 +1,94 @@
...
+        return
+    }
+
+    $oldValue = $settingFileContent.$settingName
+
+    if ($incremental) {
+        $oldValue = [System.Version]$oldValue
+        $newValue = [System.Version]$newValue
+
+        $newValue = "$($newValue.Major + $oldValue.Major).$($newValue.Minor + $oldValue.Minor)"
```
mazhelez, line 32: This part doesn't make much sense to me. Why would anyone bump their version with  "+1.2", e.g. from 2.1 to 3.3?
Even "+1.0" seems unlikely. 

@freddydk what the reasoning behind this implementation?

Actions/IncrementVersionNumber/IncrementVersionNumber.ps1
```diff
@@ -19,92 +19,50 @@ $telemetryScope = $null
...
     $serverUrl, $branch = CloneIntoNewFolder -actor $actor -token $token -updateBranch $updateBranch -DirectCommit $directCommit -newBranchPrefix 'increment-version-number'
     $baseFolder = (Get-Location).path
     DownloadAndImportBcContainerHelper -baseFolder $baseFolder
 
-    import-module (Join-Path -path $PSScriptRoot -ChildPath "..\TelemetryHelper.psm1" -Resolve)
+    Import-Module (Join-Path -path $PSScriptRoot -ChildPath "..\TelemetryHelper.psm1" -Resolve)
     $telemetryScope = CreateScope -eventId 'DO0076' -parentTelemetryScopeJson $parentTelemetryScopeJson
 
+    # Validate version number. It should match <Major>.<Minor> or +<Major>.<Minor>
+    if(-not ($versionNumber -match '^\+?\d+\.\d+$')) {
```
aholstrup1, line 32: Could we support build and revision number as well? So e.g. 8.0.1 would also be a valid version

Actions/IncrementVersionNumber/IncrementVersionNumber.psm1
aholstrup1: Mmh Set-SettingInFile could be generally useful. The Incremental switch seems fairly specific to this action though 😄  If you could refactor that somehow then I agree

mazhelez: I think the name should be more like `Set-VersionSetting` because it's not as generic.
Then `$incremental` would make sense.

Actions/IncrementVersionNumber/IncrementVersionNumber.ps1
mazhelez: I think AL-Go sets the build number at some point.

The PR doesn't aim to support more scenarios. Just fix the issue and refactor to allow testing.

Actions/IncrementVersionNumber/IncrementVersionNumber.psm1
mazhelez: Changed the function name to `Set-VersionSettingInFile` to handle only version settings.

Also, included the "incremental" logic in the function itself.

Actions/IncrementVersionNumber/IncrementVersionNumber.psm1
mazhelez: My suggestion here is to allow bumping the minor version, e.g. "+0.1" would change "1.2" to "1.3", or bumping to the next major version: "+1" would change "1.2" to "2.0".

What do you think? It's a design change though.

Actions/IncrementVersionNumber/IncrementVersionNumber.psm1
```diff
@@ -0,0 +1,139 @@
...
+        $folder = Join-Path $projectPath $_
+        $appJsonFile = Join-Path $folder "app.json"
+
+        $appJsonContent = Get-Content $appJsonFile -Raw -Encoding UTF8 | ConvertFrom-Json
+        $dependencies = $appJsonContent.dependencies
+        if ($null -ne $dependencies) {
+            $dependencies | ForEach-Object {
+                $dependency = $_
+                # Find the version of the dependency in the appInfos. If it's found, it means the dependency is part of the project and the version should be set.
+                $dependencyAppInfo = $appInfos | Where-Object { $_.name -eq $dependency.name -and $_.publisher -eq $dependency.publisher }
```
mazhelez, line 128: This wouldn't work if the dependency app is from another AL-Go project.
`$appInfos` contains the apps from the current AL-Go project.

Actions/IncrementVersionNumber/IncrementVersionNumber.psm1
freddydk: You are probably right - +1.0 seems strange.
We can make this change to basically only support:
- major.minor
- +1 = increment major and set minor to 0
- +0.1 = increment minor
Meaning that we don't support

Actions/IncrementVersionNumber/IncrementVersionNumber.psm1
freddydk: Yeah - I guess we should resolve all appfolders for all projects and then call into this function with multi-project wide appfolders instead of a single project.

Actions/IncrementVersionNumber/IncrementVersionNumber.psm1
mazhelez: Implemented.

Actions/IncrementVersionNumber/IncrementVersionNumber.psm1
mazhelez: Implemented. Testing...

mazhelez: Test run: https://github.com/mazhelez/BCApps/actions/runs/7833162162

Actions/IncrementVersionNumber/IncrementVersionNumber.psm1
```diff
@@ -0,0 +1,171 @@
...
+
+        $versionNumberFormat = '^\d+\.\d+$' # Major.Minor
+        if (-not ($newValue -match $versionNumberFormat)) {
+            throw "Version number $newValue is not in the correct format. The version number must be in the format Major.Minor (e.g. 1.0 or 1.2)"
+        }
+    }
+    #endregion
+
+    $oldValue = [System.Version] $settingFileContent.$settingName
+    $versions = @() # an array to hold the version numbers: major, minor, build, revision
```
freddydk, line 59: Minor: Confusing name versions - digits might be better

Actions/IncrementVersionNumber/IncrementVersionNumber.psm1
```diff
@@ -0,0 +1,171 @@
...
+    .Synopsis
+        Changes a version setting value in a settings file.
+    .Parameter settingsFilePath
+        Path to a JSON file containing the settings.
+    .Parameter settingName
+        Name of the setting to change. The setting must be a version number.
+    .Parameter newValue
+        New value of the setting. Allowed values are: +1 (increment major version number), +0.1 (increment minor version number), or a version number in the format Major.Minor (e.g. 1.0 or 1.2
+#>
+function Set-VersionInSettingsFile {
```
freddydk, line 11: Minor: Name implies that it always sets the version - but it only sets it if it exists already

freddydk (approved):
Looks good

Actions/IncrementVersionNumber/IncrementVersionNumber.psm1
freddydk: Actually - thinking about this - we need to update the repoversion even if it isn't there.
Else the next version will just be 1.0 as well - the repoversion defaults to 1.0 and is used as filename for the artifacts for a project

freddydk (changes requested):
We need to set RepoVersion in the settings file always.

Actions/IncrementVersionNumber/IncrementVersionNumber.psm1
```diff
@@ -0,0 +1,171 @@
...
+            Id = $app.id
+            Version = $app.version
+        }
+    })
+
+    # Update dependencies in app.json files
+    $appFolders | ForEach-Object {
+        $appJsonPath = Join-Path $_ "app.json"
+
+        $appJson = Get-Content -Path $appJsonPath -Raw | ConvertFrom-Json
```
freddydk, line 155: Get-Content needs -encoding uft8

Actions/IncrementVersionNumber/IncrementVersionNumber.psm1
```diff
@@ -0,0 +1,171 @@
...
+
+        $dependencies | ForEach-Object {
+            $dependency = $_
+            $appInfo = $appsInfos | Where-Object { $_.Id -eq $dependency.id }
+            if ($appInfo) {
+                $dependency.version = $appInfo.Version
+            }
+        }
+
+        $appJson | ConvertTo-Json -Depth 99 | Set-Content -Path $appJsonPath
```
freddydk, line 167: Should use Set-JsonContentLF which ensures a unified json formatting
ConvertTo-Json itself formats differently on PS5 and PS7

Actions/IncrementVersionNumber/IncrementVersionNumber.psm1
```diff
@@ -0,0 +1,171 @@
...
+function Set-DependenciesVersionInAppManifests {
+    param(
+        [Parameter(Mandatory = $true)]
+        [string[]] $appFolders
+    )
+
+    # Get all apps info: app ID and app version
+    $appsInfos = @($appFolders | ForEach-Object {
+        $appJson = Join-Path $_ "app.json"
+        $app = Get-Content -Path $appJson -Raw | ConvertFrom-Json
```
freddydk, line 144: -encoding utf8

freddydk (changes requested):

Actions/IncrementVersionNumber/IncrementVersionNumber.psm1
mazhelez: Changed to `$versionNumbers`. Plus the comment, I hope it's clear enough.

Actions/IncrementVersionNumber/IncrementVersionNumber.psm1
mazhelez: Ok, added functionality to ensure `repoVersion` is set in the repo settings.

For project settings, `repoVersion` will be changed only if it existed before.

Actions/IncrementVersionNumber/IncrementVersionNumber.psm1
mazhelez: `Set-JsonContentLF` outputs a warning and I don't really get why:

> WARNING: pwsh (PowerShell 7) not installed, json will be formatted by PowerShell

Anyway, changed to using `Set-JsonContentLF`. The warning could be handled separately.

freddydk (approved):

elbardel (approved):

PR #941, opened 2024-02-19
Always set repoVersion on project-level settings

mazhelez:
- `repoVersion` should always be set on project-level settings as `InrecrementVersionNumber` can be called on a project.
- Disallow lowering the `repoVersion` value.

### DIFF
--- a/Actions/IncrementVersionNumber/IncrementVersionNumber.ps1
+++ b/Actions/IncrementVersionNumber/IncrementVersionNumber.ps1
@@ -30,20 +30,18 @@ try {
 
     $settings = $env:Settings | ConvertFrom-Json
 
-    # Ensure the repoVersion setting exists in the repository settings. Defaults to 1.0 if it doesn't exist.
-    Set-VersionInSettingsFile -settingsFilePath (Join-Path $baseFolder $RepoSettingsFile) -settingName 'repoVersion' -newValue $settings.repoVersion -Force # $RepoSettingsFile is defined in AL-Go-Helper.ps1
-
-    # Change repoVersion in repository settings
-    Set-VersionInSettingsFile -settingsFilePath (Join-Path $baseFolder $RepoSettingsFile) -settingName 'repoVersion' -newValue $versionNumber
-
     $projectList = @(GetProjectsFromRepository -baseFolder $baseFolder -projectsFromSettings $settings.projects -selectProjects $projects)
 
     $allAppFolders = @()
     foreach($project in $projectList) {
         $projectPath = Join-Path $baseFolder $project
 
-        # Set repoVersion in project settings (if it exists)
         $projectSettingsPath = Join-Path $projectPath $ALGoSettingsFile # $ALGoSettingsFile is defined in AL-Go-Helper.ps1
+
+        # Ensure the repoVersion setting exists in the project settings. Defaults to 1.0 if it doesn't exist.
+        Set-VersionInSettingsFile -settingsFilePath $projectSettingsPath -settingName 'repoVersion' -newValue $settings.repoVersion -Force
+
+        # Set repoVersion in project settings according to the versionNumber parameter
         Set-VersionInSettingsFile -settingsFilePath $projectSettingsPath -settingName 'repoVersion' -newValue $versionNumber
 
         # Resolve project folders to get all app folders that contain an app.json file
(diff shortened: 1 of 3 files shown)


freddydk (approved):

Actions/IncrementVersionNumber/IncrementVersionNumber.ps1
```diff
@@ -30,20 +30,18 @@ try {
...
 
     $allAppFolders = @()
     foreach($project in $projectList) {
         $projectPath = Join-Path $baseFolder $project
 
-        # Set repoVersion in project settings (if it exists)
         $projectSettingsPath = Join-Path $projectPath $ALGoSettingsFile # $ALGoSettingsFile is defined in AL-Go-Helper.ps1
+
+        # Ensure the repoVersion setting exists in the project settings. Defaults to 1.0 if it doesn't exist.
+        Set-VersionInSettingsFile -settingsFilePath $projectSettingsPath -settingName 'repoVersion' -newValue $settings.repoVersion -Force
```
aholstrup1, line 42: Why force add repoVersion to projects? 🤔 repoVersion could also live in the repo settings (like in BCApps)

Actions/IncrementVersionNumber/IncrementVersionNumber.ps1
```diff
@@ -30,20 +30,18 @@ try {
 
     $settings = $env:Settings | ConvertFrom-Json
 
-    # Ensure the repoVersion setting exists in the repository settings. Defaults to 1.0 if it doesn't exist.
-    Set-VersionInSettingsFile -settingsFilePath (Join-Path $baseFolder $RepoSettingsFile) -settingName 'repoVersion' -newValue $settings.repoVersion -Force # $RepoSettingsFile is defined in AL-Go-Helper.ps1
-
-    # Change repoVersion in repository settings
-    Set-VersionInSettingsFile -settingsFilePath (Join-Path $baseFolder $RepoSettingsFile) -settingName 'repoVersion' -newValue $versionNumber
-
     $projectList = @(GetProjectsFromRepository -baseFolder $baseFolder -projectsFromSettings $settings.projects -selectProjects $projects)
```
aholstrup1, line 33: Can we not change the repoVersion in the repo settings if it already exists there?

Actions/IncrementVersionNumber/IncrementVersionNumber.ps1
mazhelez: (see the other comment)

Yes, yes, the action can check if it's changing the version of _all_ projects and `repoVersion` is defined in repo settings and only change that. But then it still has to go through all the projects settings and change it there as well (in case it is defined on project level as well).

aholstrup1: Perhaps.. But wouldn't it also mean that if we run this in BCApps we'd add the repoVersion into all projects? In the case of BCApps we would like the repoVersion to be the same in all projects. Doesn't it then make sense to have repoVersion on repo level rather than on project level?

Actions/IncrementVersionNumber/IncrementVersionNumber.ps1
aholstrup1: How about if (projects = '*') then change repoVersion in repo settings? If it already exists there already

freddydk: What is the problem in having repoVersion in all projects and remove it from repo settings?
We also have the version number in all app.json files, right?

freddydk (approved):

freddydk (approved):

Actions/IncrementVersionNumber/IncrementVersionNumber.ps1
freddydk: I think we can discuss this later - for now - moving the change to modify the project level settings file is what the action did before.

The earlier version also didn't modify repo settings.: https://github.com/BusinessCentralApps/buildorder/pull/166/files

The only reason that it works is that all project settings basically can be set in repo settings.

We should then decide whether we should change this - and have repoversion and projectversion or something else. Lets not change the functionality here.

Actions/IncrementVersionNumber/IncrementVersionNumber.ps1
mazhelez: Yes, it does, but it is more of a cosmetic issue.

aholstrup1 (approved):

PR #954, opened 2024-02-26
Applying conditional settings for PR Builds

aholstrup1:
AL-GO doesn't apply conditional settings on PR Builds. In BCApps we have the following conditional setting in the project settings 

```
"buildModes": [ "Translated" ],
"ConditionalSettings": [
        {
            "branches": [
                "releases/*.[0-5]"
            ],
            "settings": {
                "buildModes": [ "Translated", "Strict"]
            }
        }
    ]
```

Yet when you see the PR builds, it doesn't build in StrictMode: 
https://github.com/microsoft/BCApps/actions/runs/8044365490

In ReadSettings we only apply conditional settings if the branchname matches the pattern in the setting. The problem is that the branchName is set to GITHUB_BASE_REF which for PR builds triggered by the pull_request trigger is the merge branch (e.g. refs/heads/merge/123). The branch pattern for the conditional setting will therefore never match on PR Builds.

### DIFF
--- a/.github/workflows/CleanupTempRepos.yaml
+++ b/.github/workflows/CleanupTempRepos.yaml
@@ -70,5 +70,5 @@ jobs:
                   # Pipe empty string into GH API --METHOD DELETE due to https://github.com/cli/cli/issues/3937
                   '' | invoke-gh api --method DELETE -H "Accept: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28" "/orgs/$repoOwner/packages/nuget/$($_.name)" --input -
               }
-              invoke-gh repo delete "https://github.com/$repo" --confirm | Out-Host
+              invoke-gh repo delete "https://github.com/$repo" --yes | Out-Host
           }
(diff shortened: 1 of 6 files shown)


Actions/AL-Go-Helper.ps1
```diff
@@ -680,6 +680,12 @@ function ReadSettings {
             $settingsObjects += @($projectWorkflowSettingsObject, $userSettingsObject)
         }
     }
+
+    # If the build is triggered by a pull request the refname will be the merge branch. To apply conditional settings we need to use the base branch
```
mazhelez, line 684: Minor: `ReadSettings` has a `$branch` parameter.
So, this check is better fitted in Actions\ReadSettings\ReadSettings.ps1

mazhelez (approved):

Actions/AL-Go-Helper.ps1
aholstrup1: Mmh the issue is actually in DetermineProjectsToBuild because that also calls ReadSettings. I could call it from there though? https://github.com/microsoft/AL-Go/blob/e9f4cc63d149d341172cebfc8112e825123683d8/Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.psm1#L158

mazhelez: Yeah, I see `AnalyzeProjectDependencies` calls it too. Ok, let's keep it in here.

mazhelez (approved):

PR #957, opened 2024-02-27
Uptake DotNet signing tool 

aholstrup1:
Uptake DotNet signing tool as a replacement for AzureSignTool


Actions/Sign/Sign.psm1
```diff
@@ -1,44 +1,98 @@
...
+            Write-Host "Signing tool not found. Installing version $Version."
+            dotnet tool install sign --version $Version --global
+        }
 }
 
 <#
-.SYNOPSIS
-Register the navsip.dll in the system32 folder
+    .SYNOPSIS
+    Signs files in a given path using a certificate from Azure Key Vault.
```
mazhelez, line 25: \+ installs dotnet signing tool if it's not already installed.

Actions/Sign/Sign.psm1
```diff
@@ -1,44 +1,98 @@
...
+    Installs the dotnet signing tool.
+    .DESCRIPTION
+    Installs the dotnet signing tool.
+    .PARAMETER Version
+    The version of the signing tool to install.
+#>
+function Install-SignTool() {
+        param(
+            [Parameter(Mandatory = $false)]
+            [string] $Version = "0.9.1-beta.24123.2"
```
mazhelez, line 12: Is a repo uses self-hosted runners and this pre-release version is installed, the signing tool will never be updated when we change this version.

This version should be in some AL-Go config where we could update it when a newer version is available (or automate it? Dependabot?) and deploy new AL-Go version if needed.

mazhelez (approved):
Looks alright. Except for handling the signing tool version, the other comments are minor.

Actions/Sign/Sign.psm1
aholstrup1: True.. I've updated it now such that we install it locally rather than globally. That way we install the version specified **unless** the machine already has dotnet sign installed. This could be the case on some self-hosted runners if they are configured with the signing tool installed globally. That is not encouraged though.

I'll try to look into your second suggestion as well. Although setting up support for dependabot seems a bit out of scope for this PR. It's a nice idea though

mazhelez: The Dependabot suggestion was more of a future enhancement. My idea was to have the version in some config, not in code, so that it's easier to update it if needed.

mazhelez: I meant a config for AL-Go. So that it's clear, for example, that AL-Go v4.1 runs with v0.9.1 of the signing tool and AL-Go v5 runs with v1.0.1 of the signing tool.

freddydk: Got it - that might be a slice for later though?
I am OK with both.

Actions/Sign/Sign.psm1
```diff
@@ -1,44 +1,99 @@
...
-        Write-Host "Downloaded artifacts to $artifactTempFolder"
-        $navsip = Get-ChildItem -Path $artifactTempFolder -Filter "navsip.dll" -Recurse
-        Write-Host "Found navsip at $($navsip.FullName)"
-        Copy-Item -Path $navsip.FullName -Destination $NavSipDestination -Force | Out-Null
-        Write-Host "Copied navsip to $NavSipDestination"
-    }
-    finally {
-        Remove-Item -Path $artifactTempFolder -Recurse -Force
-    }
+        if (Get-Command -Name "sign" -ErrorAction SilentlyContinue) {
```
freddydk, line 16: Doesn't this return true if there is a function called sign in the scope as well? Is this really the right way to check whether dotnet sign tool is installed?

freddydk: Wouldn't it be safer to just always install the signing tool we need locally and ignore any globally installed version and then set the alias - this way we know what we run.
Installing it locally just says it is already installed if it is:
image

Actions/Sign/Sign.psm1
```diff
@@ -1,44 +1,99 @@
...
+            Write-Host "Signing tool is already installed"
+        } else {
+            Write-Host "Signing tool not found. Installing version $Version locally."
+            dotnet tool install sign --version $Version --tool-path SigningTool
```
freddydk, line 20: This will install the tool in the path SigningTool under the current path - but what is the current path?
I would think the safest method would be to install the signingtool to a temp folder (random name) and return the sign.exe path and then invoke that - avoiding alias' altogether.

Actions/Sign/Sign.ps1
```diff
@@ -42,23 +45,25 @@ try {
...
-            --azure-key-vault-client-id $AzureCredentials.clientId `
-            --azure-key-vault-tenant-id $AzureCredentials.tenantId `
-            --azure-key-vault-client-secret $AzureCredentials.clientSecret `
-            --azure-key-vault-certificate $Certificate `
-            --timestamp-rfc3161 "$TimestampService" `
-            --timestamp-digest $digestAlgorithm `
-            $Files
-    } -MaxRetries 3 -ArgumentList $AzureKeyVaultName, $AzureCredentials, $digestAlgorithm, $TimestampService, $Settings.keyVaultCodesignCertificateName, $Files
+    # Sign files
+    Install-SignTool -Version (Get-PackageVersion -PackageName "sign")
```
freddydk, line 52: Install-SignTool vs. Invoke-SigningTool - use SigningTool both places

freddydk (changes requested):

Actions/Sign/Sign.psm1
```diff
@@ -3,23 +3,24 @@
     Installs the dotnet signing tool.
     .DESCRIPTION
     Installs the dotnet signing tool if isn't already installed.
```
freddydk, line 5: Remove "if isn't already installed" and remove parameter description

freddydk (changes requested):

mazhelez (approved):

freddydk (approved):

PR #959, opened 2024-02-27
Update SelfHostedGitHubRunner.md

JeremyVyska:
A vital note about the CONTEXT of running these installation steps.

I was trying to setup my first GH Runner to do Signing on a private Windows device.  The sign step kept failing over and over.  Turned out, the AzureSignTool was installed in the context of the user account if the instructions were followed as-is, which my GH Runner was running under a local user account.

I lost a day, so I leave this here for folks to not suffer as I suffered.

### DIFF
--- a/Scenarios/SelfHostedGitHubRunner.md
+++ b/Scenarios/SelfHostedGitHubRunner.md
@@ -21,6 +21,10 @@ GitHub runners can be registered for an organization (accessible for all reposit
 1. Go to [Allow your repository access to your runners](#allow-your-repository-access-to-your-runners) to continue the configuration.
 
 ## Create your self-hosted runner manually
+
+> [!NOTE]
+> If the runner is running under a service account, the following should be run using the account, as this will ensure that any of the per-user installation steps are completed.  Failure to do so will result in commands not being recognized when the runner performs steps.
+
 1. To create a self-hosted runner manually, choose Windows under Runner Image and x64 in architecture and follow the description on how to create a self-hosted runner manually
 1. Make sure that the following software is installed on the computer (the suggestion in parentheses explains the mechanism used in https://aka.ms/getbuildagent)
    - Docker (getbuildagent use [this script](https://github.com/microsoft/nav-arm-templates/blob/master/InstallOrUpdateDockerEngine.ps1) to install or update Docker Engine on the Azure VM)


cegekaJG:
I've already wasted a day or two, but I think this is the exact problem I am facing right now. Can you elaborate what you mean by "using the account"?

freddydk:
Really???
When using aka.ms/getbuildagent - I do setup the agent to run under a system account - https://github.com/microsoft/nav-arm-templates/blob/06439fbb2b4fbee4051ac629737e2b659ddbade7/SetupAgents.ps1#L76 - but all the pre-requisites are just installed normally.

cegekaJG:
Now that I read it again, it's not quite the same - I am having trouble with a self-hosted runner used for deployment. I installed it as a Windows service and I kept getting errors because of the execution policy, but couldn't for the life of me figure out how to fix it. I'm currently just running it in a terminal using my admin account, but I'd like to use it as a service after all - much more reliable that way.

JeremyVyska:
For clarity, when setting up agents for Windows services, it can be a good practice to setup agents to run under a dedicated account and assign those permissions necessary for that agent, and I just followed that methodology.

Since I follow that practice, I missed that all the `choco install` commands provided are not global, but specific to the user account.  So, when logged in with an admin account, the admin account would have been all set, but not the account I needed.  :D  This note will just help with that easy-to-overlook issue.  It's not a fault with your instructions or code, but a cautionary note for IT folks trying to help BC folks.

Note:  Part of the reason for this was I could also pre-configure various User Environmental settings I needed (like Path) for that account.


For cegekaJG:   When I installed the GitHub Runner as a service, I then went to the Services and changed the login account to a dedicated user account.  It was that account I needed run all the installs under.

freddydk:
> Since I follow that practice, I missed that all the `choco install` commands provided are not global, but specific to the user account. So, when logged in with an admin account, the admin account would have been all set, but not the account I needed. :D This note will just help with that easy-to-overlook issue. It's not a fault with your instructions or code, but a cautionary note for IT folks trying to help BC folks.

Thanks - my problem is, that I don't really understand the note. How can I install something under a service account? I don't think i can? I need to install things globally, which choco install does - if the package is configured for that.
I know that the signing code using AzureSignTool checked for the existance of azuresigntool - which is probably what failed in the first place - but in the next version we are no longer using AzureSignTool, instead we are switching to dotnet sign (which now has support for BC apps) - and for dotnet sign - we always grab the version we need (we dont use the version installed if any).
So, if AzureSignTool was the primary problem (as mentioned in the description) for this - then that is no longer a problem.

PR #998, opened 2024-03-21
Add run-name to create release

### LINKED ISSUE #996: Create Release action - include version / release name.
Would it be possible to update the display name of the 'Create Release' action / workflow to include the release name and/or version, to make the log of actions a bit more readable?

image

Envisioned overview would be a bit like the releases / tags overview.

image

Expected result

- Create release - v 23.0.3
- Create release - v 23.0.2
- Create release - v 23.0.1

aholstrup1:
Add run-name to Create Release workflow

image

Fixes #996

### DIFF
--- a/Templates/Per Tenant Extension/.github/workflows/CreateRelease.yaml
+++ b/Templates/Per Tenant Extension/.github/workflows/CreateRelease.yaml
@@ -1,4 +1,5 @@
 name: ' Create release'
+run-name: "Create release - Version ${{ inputs.tag }}"
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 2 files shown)


Templates/Per Tenant Extension/.github/workflows/CreateRelease.yaml
```diff
@@ -1,4 +1,5 @@
 name: ' Create release'
+run-name: "Create release - Version ${{ inputs.tag }}"
```
freddydk, line 2: doesn't the run-name section has to be after the workflow-dispatch section?

aholstrup1: This worked just fine for me 🤷‍♂️ 

Here's an example (that failed but for a different reason I think 😆)
https://github.com/aholstrup1/BCApps/actions/runs/8371584108

mazhelez (approved):

freddydk (approved):
I can see we do the same in all other workflows.

PR #1029, opened 2024-04-14
useEarliestArtifact + a few small bugs

### LINKED ISSUE #818: The link https://docs.microsoft.com/en-us/azure/developer/github/github-key-vault is redirected to https://learn.microsoft.com/en-us/azure/developer/github, but I can not find the page where it describe how to setup the key vault for access from GitHub.
The link https://docs.microsoft.com/en-us/azure/developer/github/github-key-vault is redirected to https://learn.microsoft.com/en-us/azure/developer/github, but I can not find the page where it describe how to setup the key vault for access from GitHub.
Can you update this to  the correct link?

_Originally posted by @FrankAuwers in https://github.com/microsoft/AL-Go/issues/182#issuecomment-1809912737_

### LINKED ISSUE #918: Add ValidateSet for auth parameter in localdevenv scripts
Would be great if we could add the following attribute to the parameter ``$auth`` of the ``localDevEnv`` scripts.
Therefore autocomplete for this parameter would work.

``` pwsh
[ValidateSet("UserPassword", "Windows")]
```

```diff
Param(
    [string] $containerName = "",
+   [ValidateSet("UserPassword", "Windows")]
    [string] $auth = "",
    [pscredential] $credential = $null,
    [string] $licenseFileUrl = "",
    [switch] $fromVSCode,
    [switch] $accept_insiderEula
)
```

Affected scripts:
``Templates/Per Tenant Extension/.AL-Go/localDevEnv.ps1``
``Templates/AppSource App/.AL-Go/localDevEnv.ps1``

### LINKED ISSUE #970: LicenseFileUrl
@freddydk I'm currently relaunching our 'move to AL-GO' journey internally and have a question on the 'LicenseFileUrl' secret as referenced in various wiki pages / documentation. I was about to configure the 'LicenseFileUrl' secret as described in the wiki, but wondered if this is still valid / required while reading this post about updated Cronus demo licenses. Do we still need a **developers** license to build apps in CI/CD?

https://www.yammer.com/dynamicsnavdev/#/Threads/show?threadId=2172494757232640

`Developers no longer need partner licenses or powerful customer licenses to do development. `

### LINKED ISSUE #1018: Artifact setting - possiblity to read version from app.json
From what I can tell from the AL-Go PS Code, the artifact settings always needs to contain a version number. Any chance this version can be read from the app.json files in a project? When multiple versions are available, the highest should be used.

freddydk:
Fix regression: DownloadRelease escapes the '*' sign in the project name

Fixes #1018
Fixes #970
Fixes #818
Fixes #918

Adding new setting:
- `useEarliestArtifact`: set this to true in order for AL-Go for GitHub to build a project using the earliest compatible Business Central artifact, determined by the highest application dependency from app.json in all apps in the project and the `applicationDependency` in project settings.

Small bugs:
- Only show a warning about a specific version of BcContainerHelper if this is specified in Settings


RELEASENOTES.md
```diff
@@ -13,9 +13,11 @@ Note that when using the preview version of AL-Go for GitHub, we recommend you U
 - Issue 997 'Deliver to AppSource' action fails for projects containing a space
 - Issue 987 Resource not accessible by integration when creating release from specific version
 - Issue 979 Publish to AppSource Documentation
+- Issue 1018 Artifact setting - possiblity to read version from app.json
 - Issue 1008 Allow PullRequestHandler to use ubuntu or self hosted runners for all jobs except for pregateCheck
 
 ### New Settings
+- `useEarliestArtifact`: set this to true in order for AL-Go for GitHub to build a project using the earliest compatible Business Central artifact, determined by the highest application dependency from app.json in all apps in the project and the `applicationDependency` in project settings.
```
aholstrup1, line 20: Maybe I misunderstand the original issue (or the name of the setting), but aren't they expecting to get the _latest_ artifact of a certain BC version rather than the earliest? 

If their artifact setting currently is `"artifact": "//23.1//latest"` I'm guessing they're expecting to get the latest 23.1 artifact

Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -124,12 +124,6 @@ try {
         exit
     }
 
-    if ($settings.type -eq "AppSource App" ) {
```
mazhelez, line 127: Is this change related?

Actions/AL-Go-Helper.ps1
```diff
@@ -18,7 +18,7 @@ $defaultCICDPushBranches = @( 'main', 'release/*', 'feature/*' )
 [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSUseDeclaredVarsMoreThanAssignments', 'defaultCICDPullRequestBranches', Justification = 'False positive.')]
 $defaultCICDPullRequestBranches = @( 'main' )
 $runningLocal = $local.IsPresent
-$defaultBcContainerHelperVersion = "preview" # Must be double quotes. Will be replaced by BcContainerHelperVersion if necessary in the deploy step - ex. "https://github.com/organization/navcontainerhelper/archive/refs/heads/branch.zip"
+$defaultBcContainerHelperVersion = "https://github.com/freddydk/navcontainerhelper/archive/refs/heads/customcodecops.zip" # Must be double quotes. Will be replaced by BcContainerHelperVersion if necessary in the deploy step - ex. "https://github.com/organization/navcontainerhelper/archive/refs/heads/branch.zip"
```
mazhelez, line 21: Don't forget to revert this part.

RELEASENOTES.md
aholstrup1: That is the one I read 😄 Don't they want to use `"artifact": "//23.1//latest"` but without specifying 23.1

freddydk: "artifact": "//23.1//latest" will indeed give them the latest 23.1 - and it will continue to do so.
"artifact": "////latest" will give them the latest - artifact.
removing the artifact setting and setting useEarlistArtifact (a better name suggestion is welcome) will give them the first Business Central version they claim to be compatible with. If they have 23.1.52400.43123, 23.0.0.0, 20.1.0.0 and 22.5.23567.0 in 4 different App.Json's - they will get the first version greater or equal to 23.1.52400.43123. You could argue that we should take the latest 23.1 version, instead of looking at the full version number, but you could be in a situation where you would install on 23.1.52400.43123 right?

Actions/AL-Go-Helper.ps1
freddydk: Agree - but FYI - if we forget - deploy will always use latest or preview - we will never get a version in prod with this

RELEASENOTES.md
freddydk: By using the earliest we also get caching of the artifact - it doesn't change all the time.
Test Current can then test the current version.
In the end - we should build in a mechanism to clean up sandbox artifacts no longer used by tenants online - then we would build against exactly what we can end up installing this on.

freddydk: These comments gave me an idea though...
Instead of adding another property, we could use

"artifact": "//*//latest" for latest major with the same major.minor as the highest version from app.json
and
"artifact": "//*//first" for the first version -ge than the highest version from app.json

What do you think about that idea?

aholstrup1: > These comments gave me an idea though... Instead of adding another property, we could use
> 
> `"artifact": "//*//latest"` for latest major with the same major.minor as the highest version from app.json and `"artifact": "//*//first"` for the first version -ge than the highest version from app.json
> 
> What do you think about that idea?

I like that more 👍 

As I read the issue, this `"artifact": "//*//latest"` is really what the issue was asking for (maybe I misunderstood it though)

freddydk: That is exactly what he was asking for and `//*//first` was what -useEarliestArtifact would give you
I have changed the implementation and remove the useEarliestArtifact setting
Will run end 2 end to see if this has any impact

freddydk: The log will look like this:
image

Scenarios/UseAzureKeyVault.md
```diff
@@ -1,5 +1,5 @@
...
+*Prerequisites: A completed [scenario 6](UpdateAlGoSystemFiles.md), an Azure KeyVault and you will need to follow the guidelines on how to connect to an Azure KeyVault as specified [here](https://learn.microsoft.com/en-us/azure/developer/github/connect-from-azure?tabs=azure-portal%2Cwindows#use-the-azure-login-action-with-a-service-principal-secret). Add your KeyVault name to the the JSON construct from this walkthrough (using **“keyVaultName” : “{your keyvault name}”**) and add this JSON construct as a repository secret called AZURE_CREDENTIALS. You can also specify the KeyVault name in the AL-Go settings file if you do not wait to mess with the JSON construct.*
```
aholstrup1, line 2:
```suggestion
*Prerequisites: A completed [scenario 6](UpdateAlGoSystemFiles.md), an Azure KeyVault and you will need to follow the guidelines on how to connect to an Azure KeyVault as specified [here](https://learn.microsoft.com/azure/developer/github/connect-from-azure?tabs=azure-portal%2Cwindows#use-the-azure-login-action-with-a-service-principal-secret). Add your KeyVault name to the the JSON construct from this walkthrough (using **“keyVaultName” : “{your keyvault name}”**) and add this JSON construct as a repository secret called AZURE_CREDENTIALS. You can also specify the KeyVault name in the AL-Go settings file if you do not wait to mess with the JSON construct.*
```

aholstrup1 (dismissed):

mazhelez (approved):

PR #1031, opened 2024-04-14
issue #962

### LINKED ISSUE #792: powershell: command not found error when using the ubuntu settings
I have the configuration 

"DeployToSandbox": {
        "EnvironmentType": "SaaS",
        "EnvironmentName": "Sandbox",
        "Branches": [
          "main"
        ],
        "Projects": "*",
        "SyncMode": "ForceSync",
        "ContinuousDeployment": true,
        "runs-on": "ubuntu-latest"
    },

I also have a self-hosted agent.

the "runs-on": "ubuntu-latest" causes a deployment error in my setup.

` $errorActionPreference = "Stop"; $ProgressPreference = "SilentlyContinue"; Set-StrictMode -Version 2.0
  try {
    /home/runner/work/_actions/microsoft/AL-Go-Actions/v4.0/ReadSettings/ReadSettings.ps1 -project $ENV:_project -get $ENV:_get
  }
  catch {
    Write-Host "::ERROR::Unexpected error when running action. Error Message: $($_.Exception.Message.Replace("`r",'').Replace("`n",' ')), StackTrace: $($_.ScriptStackTrace.Replace("`r",'').Replace("`n",' <- '))";
    exit 1
  }
  Error: powershell: command not found`

### LINKED ISSUE #962: Finer control of "shell"-property
During troubleshooting a custom OnPrem deployment solution, I realized (unfortunately rather late) that AL-Go now uses `pwsh` as the PowerShell command by default. The documentation says it's still possible to use the old `powershell` using the property "shell", but that affects all jobs in all runners. I checked "DetermineDeploymentEnvironments", and it didn't look like the property exists in this scope.
Since a runner or environment is far more indicative as to which command of powershell to use, one of these two changes should be made:

1) Add the property "shell" to the properties of the "DeployTo..." object in `.github/AL-Go-settings.json`
2) Add the condition "runner tags" to the conditional settings

Either of these should make it possible to define the PowerShell command per environment.

freddydk:
Fixes #962
Fixes #792 

Add "shell" as a property under DeployTo, allowing you to control which shell to use on a self hosted runner f.eks. - not forcing you to use the same shell as for other jobs.


mazhelez (dismissed):

mazhelez (approved):

Actions/DetermineDeploymentEnvironments/DetermineDeploymentEnvironments.ps1
```diff
@@ -139,18 +141,22 @@ else {
...
         if ($settings.ContainsKey($settingsName)) {
             # If a DeployTo<environmentName> setting exists - use values from this (over the defaults)
             $deployTo = $settings."$settingsName"
-            foreach($key in 'EnvironmentType','EnvironmentName','Branches','Projects','SyncMode','ContinuousDeployment','runs-on') {
+            foreach($key in 'EnvironmentType','EnvironmentName','Branches','Projects','SyncMode','ContinuousDeployment','runs-on','shell') {
                 if ($deployTo.ContainsKey($key)) {
                     $deploymentSettings."$key" = $deployTo."$key"
                 }
             }
+            if ($deploymentSettings."shell" -ne 'pwsh' -and $deploymentSettings."shell" -ne 'powershell') {
```
aholstrup1, line 157: Shouldn't this validation happen in ReadSettings rather? I guess shell should always be either pwsh or powershell regardless of whether you deploy to and environment or not

aholstrup1 (approved):

mazhelez (approved):

Actions/DetermineDeploymentEnvironments/DetermineDeploymentEnvironments.ps1
freddydk: Problem is really that these shell settings are under DeployToXXX - and we don't enumerate DeployTo's in ReadSettings.
Testing it here ensures that we don't use a different shell for these.

PR #1053, opened 2024-04-24
Update docs around CICDSchedule 

### LINKED ISSUE #1048: Question - CICDSchedule in AL-Go-Settings.json overwrites workflow_dispatch push in CICD.yaml
Hi there,
when adding the CICDSchedule in AL-Go-Settings.json it overwrites the default push dispatch.
If this is done intentionally the docs (https://github.com/microsoft/AL-Go/blob/main/Scenarios/settings.md#cicdschedule) should be clear about this.

Otherwise, it seems to be a bug 🙆🏻‍♂️

aholstrup1:
Update documentation to reflect that when setting CICDSchedule the CI/CD workflow will stop triggering on push unless CICDPushBranches is also set

Fixes #1048


freddydk:
I can see that the current behavior is very much intended. Reason for the behavior is that why would you be running CI/CD on push if you are setting it up to be run daily or weekly?
Setting the CICDSchedule is really intended for people not running CI/CD on PR and push - but only doing nightly builds f.ex.
If you are setting up CI/CD on a schedule to do testing every night - it should be done using the Test Current workflow, which doesn't produce artifacts etc.

I might be arguing that this is intended behavior and something we should document instead.

aholstrup1:
I'm fine with both honestly :) We do allow setting both CICDSchedule and CICDPushBranches as long as both are defined in the AL-Go settings.

A drawback of schedules in GitHub is that they only trigger on the default branch. If you want to run CICDs in other branches you might want both CICDSchedule and CICDPushBranches :)

freddydk (approved):

ankosiorek (approved):

PR #1058, opened 2024-04-27
5 small bug fixes in preview

### LINKED ISSUE #1034: Reducing build time by skipping app compilation and installation in localdevenv.ps1 script.
If I do not want to compile my own apps when building the container and also do not want to install them (i.e. only install the dependencies). How should I set this up?

 I had already tried to set "DoNotPublishApps" to true. Unfortunately, this did not have the desired effect. Likewise, setting "appFolders" (localdev.settings.json) to empty was not effective, as the array is apparently only updated but not overwritten. 

The reason for my question is as follows:
The creation of the container with dependencies takes about 10 minutes, if the apps are also compiled and installed, we are talking about 30 minutes. That's a bit too long for many of our developers for just a small change in on of our apps. Hence the question of whether it can be deactivated.

### LINKED ISSUE #1038: Publish To AppSource - Include project list
Similar to #996, it would be more clear if the list of **'Publish to AppSource'** workflow runs would contains the **Project** setting (_only when different from * ?_) to indicate which  project(s) / app(s) have been included in the workflow runs in case of a multi-project repo?

image

---

- Publish To AppSource - **ProjectName1**
- Publish To AppSource - **ProjectName2**
- Publish To AppSource - **ProjectName1,ProjectName2**
- Publish To AppSource                       (= all projects)

  or               
  
- Publish To AppSource  - All projects (= all projects)

### LINKED ISSUE #1042: Downloading artifacts from GitHub doesn't work with branch names which include forward slashes
I'm having issues with deploying artifacts, which were created and uploaded to GitHub by the CI/CD pipeline.

When using the branch pattern `feat/mycoolbranchname` with a forward slash `/`, the artifacts won't get detected during the `Publish to Environment` Action (with version set to `latest` aka latest CI/CD build) at the start of the `Deploy` step.
It looks like forward slashes `/` get escaped into underscores `_` for branch names (and their variables in PS)

During the `Deploy` step, the escaped branch name seems to get passed into the [FindLatestSuccessfulCICDRun ](https://github.com/microsoft/AL-Go/blob/dd5c9ebf4bd06d33d45a26aa2fb3d0620277b1d0/Actions/Github-Helper.psm1#L815) function, where an API Call is made to the GitHub /actions/runs endpoint. 

`$runsURI = "https://api.github.com/repos/$repository/actions/runs?per_page=$per_page&page=$page&exclude_pull_requests=true&status=completed&branch=$branch"`

The issue is that the GitHub API expects the actual branch name for the branch parameter.
Respectively, for `feat/mycoolbranchname` the "escaped" result is `feat_mycoolbranchname`. The value `feat_mycoolbranchname` now gets passed into the request parameter and the API returns 0 results and the script fails.

An example:

grafik
Branch name starts with a forward slash (feat/phm_test)

grafik
Notice that the forward slash is now an underscore. (The `***` attempts to obfuscate the word "test", weird)

I verified via the GitHub API that there is in fact an artifact to download, if the branch name was correct.

Of course, not using a forward slash for example a branch with the name `feat-mycoolbranchname` works but currently limits the freedom in naming.

### LINKED ISSUE #1043: Determine Deployment Environments returning (404) Not Found [CI/CD]
When the CI/CD action is Determine Deployment Environments, an error pops in the log:

```log
Run microsoft/AL-Go/Actions/DetermineDeploymentEnvironments@dd5c9ebf4bd06d33d45a26aa2fb3d0620277b[1](https://github.com/MyCompany/PTE-BC-MyCompany/actions/runs/8722731937/job/23929261049#step:11:1)d0
Run $errorActionPreference = "Stop"; $ProgressPreference = "SilentlyContinue"; Set-StrictMode -Version 2.0
GenerateALDocArtifact=0
DeployALDocArtifact=1
Environment pattern to use: *
Requesting environments from GitHub
Error: The remote server returned an error: ([40](https://github.com/MyCompany/PTE-BC-MyCompany/actions/runs/8722731937/job/23929261049#step:11:42)4) Not Found.

Not Found

Failed to get environments from GitHub API - Environments are not supported in this repository
Reading environments from settings
```

This is the setting in my .github/AL-Go-Settings.json
```json
{
  "type": "PTE",
  "templateUrl": "https://github.com/microsoft/AL-Go-PTE@preview",
  "CICDPushBranches": [
    "main",
    "uat/*"
  ],
  "environments": [
    "UAT_20240319",
    "Production"
  ],
  "DeployToUAT": {
    "EnvironmentName": "UAT_20240319",
    "continuousDeployment": true,
    "Branches": [
      "uat/uat"
    ]
  },
  "DeployToProduction": {
    "EnvironmentName": "Production",
    "continuousDeployment": false,
    "Branches": [
      "main"
    ]
  },
  "templateSha": "9c047553618e3eb937d33b03e4407354e3ef4531"
}
```

After all, the actions succeeds, but there is an error message in the annotations

image

### LINKED ISSUE #1047: 403 forbidden with project dependencies
I found 2 settings are missing when configuring GHTOKENWORKFLOW

1)
when using fine grained tokens on organisation you can break things by restricting the usage of the PAT tokens.
on the following page:
https://github.com/organizations/enterprise/settings/personal-access-tokens

When using fine grained tokens:
- Allow access via fine-grained personal access tokens
- Do not require administrator approval

When using using classic tokens:
Allow access via personal access tokens (classic)

screenshot if you allow both:
image

2) fined grained access tokens do not work when having project dependencies in other internal repos.
It just can't download the artifacts, i keep getting 403 forbidden.
like below error:

image

this is resolved by granting extra rights to fine grained token:
image

image

Please reflect that in the doc on this page 
https://github.com/microsoft/AL-Go/blob/main/Workshop/Prerequisites.md
or this one in step 6
https://github.com/microsoft/AL-Go/blob/main/Scenarios/UpdateAlGoSystemFiles.md

freddydk:
Fixes #1038
Fixes #1043 (do not write useless annotations)
Fixes #1047
Fixes #1042
Fixes #1034

Also check rate limits during end2end tests (waitWorkflow) and wait for rate limit to reset if < 10%
We might want to do this for AL-Go more generally going forward if people run into rate limits.


Templates/AppSource App/.AL-Go/localDevEnv.ps1
```diff
@@ -10,7 +10,8 @@ Param(
     [pscredential] $credential = $null,
     [string] $licenseFileUrl = "",
     [switch] $fromVSCode,
-    [switch] $accept_insiderEula
+    [switch] $accept_insiderEula,
+    [switch] $clean
```
aholstrup1, line 14: I think we should consider adding some docs to the localdevenv script. I don't think it's that obvious what "clean" entails in this context.

aholstrup1 (approved):
Looks okay overall. Consider changing the name of the PR though. Once this goes into the git history "Quickfixes" isn't that descriptive

mazhelez (dismissed):

Templates/AppSource App/.AL-Go/localDevEnv.ps1
freddydk: Yeah - will add this after v5.1

mazhelez (approved):

PR #1064, opened 2024-05-02
Automate release process

freddydk:
Automate:
- The generation of modifying release notes with the version number (and revert)
- Creation of a release in AL-Go repo

Adds:
- Check that a successful end 2 end test was run on the same commit you are releasing

### DIFF
--- a/.github/workflows/Deploy.yaml
+++ b/.github/workflows/Deploy.yaml
@@ -123,4 +123,28 @@ jobs:
           $releaseNotes = (Get-Content -Encoding utf8 -Path $releaseNotesFile) -join "`n"
           $lastVersion = $releaseNotes.indexof("`n## ")
           $releaseNotes = $releaseNotes.Substring(0, $lastVersion)
-          Write-Host $releaseNotes
+          Add-Content -encoding UTF8 -Path $ENV:GITHUB_ENV -Value "ReleaseNotes=([Uri]::EscapeDataString($releaseNotes))"
+
+      - name: Create release
+        uses: actions/github-script@v7
+        id: createrelease
+        env:
+          branch: ${{ github.event.inputs.branch }}
+          bodyMD: ${{ env:ReleaseNotes }}
+        with:
+          github-token: ${{ secrets.OrgPAT }}
+          script: |
+            var bodyMD = process.env.bodyMD
+            const createReleaseResponse = await github.rest.repos.createRelease({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              tag_name: '${{ env.branch }}',
+              name: '${{ env.branch }}',
+              body: decodeURIComponent(bodyMD),
+              draft: false,
+              prerelease: false
+            });
+            const {
+              data: { id: releaseId, html_url: htmlUrl, upload_url: uploadUrl }
+            } = createReleaseResponse;
+            core.setOutput('releaseId', releaseId);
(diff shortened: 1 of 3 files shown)


.github/workflows/Deploy.yaml
```diff
@@ -28,8 +33,32 @@ defaults:
...
 
 jobs:
+  CheckEndToEnd:
+    runs-on: [ ubuntu-latest ]
+    steps:
+      - name: Check successful end 2 end tests have run
+        if: github.repository_owner == 'microsoft'
+        env:
+          GH_TOKEN: ${{ github.token }}
+          repository: ${{ github.repository }}
```
aholstrup1, line 43: I think GitHub already has $ENV:GITHUB_REPOSITORY for this purpose

.github/workflows/Deploy.yaml
```diff
@@ -90,3 +119,38 @@ jobs:
...
+        run: |
+          $errorActionPreference = "Stop"; $ProgressPreference = "SilentlyContinue"; Set-StrictMode -Version 2.0
+          $releaseNotesFile = Join-Path $env:GITHUB_WORKSPACE "RELEASENOTES.md"
+          $releaseNotes = (Get-Content -Encoding utf8 -Path $releaseNotesFile) -join "`n"
+          $lastVersion = $releaseNotes.indexof("`n## ")
+          $releaseNotes = $releaseNotes.Substring(0, $lastVersion)
+          Add-Content -encoding UTF8 -Path $env:GITHUB_ENV -Value "ReleaseNotes=$([Uri]::EscapeDataString($releaseNotes))"
+
+      - name: Create release
+        if: github.repository_owner == 'microsoft' && github.event.inputs.createRelease
```
aholstrup1, line 134:
```suggestion
        if: github.repository_owner == 'microsoft' && github.event.inputs.createRelease == 'true'
```

Not sure if I remember this correctly, but aren't the boolean inputs to GitHub either 'true' or 'false' (as strings)? 

For example in E2E do: 
`github.event.inputs.runUpgradeTests == 'true'`

.github/workflows/Deploy.yaml
```diff
@@ -90,3 +119,38 @@ jobs:
             Write-Host "::Error::Error deploying repositories. The error was $($_.Exception.Message)"
             exit 1
           }
+
+      - name: Calculate Release Notes
+        if: github.repository_owner == 'microsoft'
```
aholstrup1, line 124: Check for github.event.inputs.createRelease - No need to run this if we aren't creating a release

aholstrup1 (approved):

mazhelez (approved):

.github/workflows/Deploy.yaml
```diff
@@ -90,3 +122,38 @@ jobs:
...
+
+      - name: Calculate Release Notes
+        if: github.repository_owner == 'microsoft' && github.event.inputs.createRelease == 'true'
+        run: |
+          $errorActionPreference = "Stop"; $ProgressPreference = "SilentlyContinue"; Set-StrictMode -Version 2.0
+          $releaseNotesFile = Join-Path $env:GITHUB_WORKSPACE "RELEASENOTES.md"
+          $releaseNotes = (Get-Content -Encoding utf8 -Path $releaseNotesFile) -join "`n"
+          $lastVersion = $releaseNotes.indexof("`n## ")
+          $releaseNotes = $releaseNotes.Substring(0, $lastVersion)
+          Add-Content -encoding UTF8 -Path $env:GITHUB_ENV -Value "ReleaseNotes=$([Uri]::EscapeDataString($releaseNotes))"
```
mazhelez, line 134: `GITHUB_ENV` has some length limitation. Could release notes get big enough to reach this limitation?

freddydk: I think the limit is 64Kb.
We only use the env variable for the release notes for the latest version, which isn't very big.
Today the release notes for ALL versions is 41kb.

aholstrup1 (approved):

PR #1076, opened 2024-05-14
Deploybug

freddydk:
2 bugs when deploying to preview

- use of unknown variable (reserveReleaseNotes)
- use of wrong name of end 2 end tests workflow when searching for workflow results

Also, when testing that end 2 end tests have run - check that no jobs were skipped (running only partial tests)
autodeploy to preview after successful end 2 end tests.


.github/workflows/Deploy.yaml
```diff
@@ -3,6 +3,10 @@ name: Deploy
 run-name: "Deploy from [${{ github.ref_name }}] to [${{ inputs.branch }}]"
 
 on:
+  workflow_run:
```
aholstrup1, line 6: Won't this trigger even if I run E2E tests on some other branch? 
image

Actions/WorkflowInitialize/WorkflowInitialize.ps1
```diff
@@ -9,18 +9,27 @@ try {
     . (Join-Path -Path $PSScriptRoot -ChildPath "..\AL-Go-Helper.ps1" -Resolve)
     . (Join-Path -Path $PSScriptRoot -ChildPath "..\AL-Go-TestRepoHelper.ps1" -Resolve)
 
-    $ap = "$ENV:GITHUB_ACTION_PATH".Split('\')
+    $ap = "$ENV:GITHUB_ACTION_PATH".Split([System.IO.Path]::DirectorySeparatorChar)
```
aholstrup1, line 12: Is this related to the AL-Go deployment bug? 🤔 If not, it probably belongs in a different PR

Actions/WorkflowInitialize/WorkflowInitialize.ps1
```diff
@@ -9,18 +9,27 @@ try {
...
-    elseif ($branch -eq "preview") {
-        $verstr = "p"
-    }
     else {
-        $verstr = $branch
+        Write-Host "Using AL-Go for GitHub $branch"
+        if ($owner -ne "microsoft") {
+            $verstr = "d"
+        }
+        elseif ($branch -eq "preview") {
```
aholstrup1, line 27: We don't really refer to AL-Go actions in preview as e.g. "WorkflowInitialize@Preview" anymore I think. In BCApps we get a commit sha instead. I wonder if we should handle that scenario as well / instead?

.github/workflows/Deploy.yaml
freddydk: Yes it will - but the auto-trigger will check that a successful end 2 end test has been run on the latest sha (see checkend2end)

Actions/WorkflowInitialize/WorkflowInitialize.ps1
freddydk: I will move the workflowinitialize bug fixes to a different PR.

Actions/WorkflowInitialize/WorkflowInitialize.ps1
freddydk: doesn't belong here - sry

.github/workflows/Deploy.yaml
```diff
@@ -27,49 +31,99 @@ on:
...
     steps:
       - name: Check successful end 2 end tests have run
-        if: github.repository_owner == 'microsoft' && github.event.inputs.requireEndToEndTests == 'true'
+        if: github.repository_owner == 'microsoft' && needs.Inputs.outputs.requireEndToEndTests == 'true'
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
           $errorActionPreference = "Stop"
-          $end2endRuns = (gh api /repos/$($env:GITHUB_REPOSITORY)/actions/runs?per_page=100 | ConvertFrom-Json).workflow_runs | Where-Object { $_.name -eq 'End 2 End Tests' }
+          $end2endRuns = (gh api /repos/$($env:GITHUB_REPOSITORY)/actions/runs?per_page=100 | ConvertFrom-Json).workflow_runs | Where-Object { $_.name -eq 'End to end tests' }
```
mazhelez, line 98: [List workflow runs for a workflow](https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-workflow) is better suited here.

.github/workflows/Deploy.yaml
```diff
@@ -27,49 +31,99 @@ on:
...
+            throw "End to end tests run for the latest commit on main is not completed"
           }
           if ($latestRun.conclusion -ne 'success') {
-            throw "End 2 End Tests run for the latest commit on main did not succeed"
+            throw "End to end tests run for the latest commit on main did not succeed"
+          }
+          $allJobs = (gh api /repos/$($env:GITHUB_REPOSITORY)/actions/runs/$($latestRun.id)/jobs | ConvertFrom-Json).jobs
+          foreach($job in $allJobs) {
+            if ($job.conclusion -ne 'success') {
+              throw "Some jobs in the end to end tests run for the latest commit was skipped, failed or cancelled"
```
mazhelez, line 113: For better navigation, consider including a link to the failed E2E tests run and/or to the non-successful jobs.

.github/workflows/Deploy.yaml
```diff
@@ -27,49 +31,99 @@ on:
...
         required: false
-        default: true
+        default: false
       defaultBcContainerHelperVersion:
         description: 'Which version of BcContainerHelper to use? (latest, preview, private, a specific version number or a direct download URL like https://github.com/freddydk/navcontainerhelper/archive/master.zip). Leave empty to use latest (or preview for preview branches)'
         required: false
-        default: ''
+        default: 'latest'
+        
+permissions:
```
mazhelez, line 40: Assign the needed permissions on job-level. I suppose not all jobs need all the permissions.

mazhelez (commented):
A couple of things to address. See comments.

.github/workflows/Deploy.yaml
aholstrup1: Mmh but for example this workflow: 
https://github.com/microsoft/AL-Go/actions/runs/9062344772

The head_sha is 654dd546357a63bd9421137709ad293245095bb2 which is from the main branch but it looks like you ran the E2E test on a different ref (b4407559407171bffd82f47f3064f9c4fae01a7b)? Wouldn't CheckEndToEnd pass in that case?

.github/workflows/Deploy.yaml
freddydk: ok

.github/workflows/Deploy.yaml
freddydk: You might be right...
You probably are - will check that.

freddydk: There is no good way of getting the input from an end 2 end test run - so I have added a job to the end 2 end tests which gets skipped if we are running on a specified sha - basically meaning that this end 2 end test isn't useful:
image

This is how it looks when running on an empty ref:
image

freddydk: Good catch

.github/workflows/Deploy.yaml
```diff
@@ -27,49 +31,104 @@ on:
...
+        if: github.repository_owner == 'microsoft' && needs.Inputs.outputs.requireEndToEndTests == 'true'
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
           $errorActionPreference = "Stop"
-          $end2endRuns = (gh api /repos/$($env:GITHUB_REPOSITORY)/actions/runs?per_page=100 | ConvertFrom-Json).workflow_runs | Where-Object { $_.name -eq 'End 2 End Tests' }
+          $end2endRuns = (gh api /repos/$($env:GITHUB_REPOSITORY)/actions/workflows/E2E.yaml/runs?per_page=100 | ConvertFrom-Json).workflow_runs
           $latestSha = (gh api /repos/$($env:GITHUB_REPOSITORY)/commits/main | ConvertFrom-Json).sha
-          $latestRun = $end2endruns | Where-Object { $_.head_sha -eq $latestSha }
+          $latestRun = $end2endruns | Where-Object { $_.head_sha -eq $latestSha } | select-object -first 1
```
mazhelez, line 100: The API has `status` and `head_sha` parameter that could be utilized in this case. See https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-workflow


I think it's best to use these parameters as latest run might not be successful, but the previous run (on the same SHA) might have been successful. It's weird case (why would we run e2e tests on the same SHA if the previous run succeeded), but who knows... we could reach the point where we run e2e a couple of times in hopes that one of them is fully successful.

.github/workflows/E2E.yaml
```diff
@@ -79,6 +79,17 @@ jobs:
           Add-Content -Encoding UTF8 -Path $env:GITHUB_OUTPUT -Value "githubOwner=$githubOwner"
           Write-Host "githubOwner=$githubOwner"
 
+  CheckRef:
+    runs-on: [ ubuntu-latest ]
+    needs: [ Check ]
+    if: github.event.inputs.ref == ''
```
mazhelez, line 85:
```suggestion
    if: (github.event.inputs.ref == '') || (github.event.inputs.ref == github.sha)
```

I myself sometimes provide the latest SHA as an argument, even though it doesn't matter.

.github/workflows/Deploy.yaml
mazhelez: Alternative approach: include the SHA (the input) in the display title of the E2E Tests run.

That way it will be visible which E2E Test run is on _main_ and will save a call to the jobs API.

mazhelez: Did a small prototype: https://github.com/mazhelez/BCApps/blob/main/.github/workflows/Test.yaml

image

.github/workflows/Deploy.yaml
mazhelez: With the other discussion in place, the `head_sha` won't be needed, but the status would be useful.

.github/workflows/Deploy.yaml
freddydk: Good idea - changed the E2E to 
image

I still need to check that jobs are not skipped though - else this would be successful:
image

.github/workflows/Deploy.yaml
```diff
@@ -27,49 +31,101 @@ on:
...
           }
           if ($latestRun.status -ne 'completed') {
-            throw "End 2 End Tests run for the latest commit on main is not completed"
+            throw "End to end tests run for the latest commit on main is not completed - see $($latestRun.html_url)"
           }
           if ($latestRun.conclusion -ne 'success') {
-            throw "End 2 End Tests run for the latest commit on main did not succeed"
+            throw "End to end tests run for the latest commit on main did not succeed - see $($latestRun.html_url)"
+          }
+          $allJobs = (gh api /repos/$($env:GITHUB_REPOSITORY)/actions/runs/$($latestRun.id)/jobs | ConvertFrom-Json).jobs
```
mazhelez, line 110: This would return only 30 jobs by default.

Alexander used `--paginate` in his PR: https://github.com/microsoft/AL-Go/blob/f82966816d304985bde80f66e82e5b36b0673141/Actions/WorkflowPostProcess/WorkflowPostProcess.ps1#L18

.github/workflows/E2E.yaml
```diff
@@ -1,5 +1,7 @@
 name: End to end tests
 
+run-name: End to end tests - ${{ github.event.inputs.ref != '' && github.event.inputs.ref || github.sha }}
```
mazhelez, line 3: Not sure what this is supposed to mean 🤔 

And I tested it a bit and now even more confused:
image

https://github.com/mazhelez/BCApps/actions/runs/9085072498/job/24967509935

.github/workflows/Deploy.yaml
```diff
@@ -27,49 +31,101 @@ on:
...
     steps:
       - name: Check successful end 2 end tests have run
-        if: github.repository_owner == 'microsoft' && github.event.inputs.requireEndToEndTests == 'true'
+        if: github.repository_owner == 'microsoft' && needs.Inputs.outputs.requireEndToEndTests == 'true'
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
           $errorActionPreference = "Stop"
-          $end2endRuns = (gh api /repos/$($env:GITHUB_REPOSITORY)/actions/runs?per_page=100 | ConvertFrom-Json).workflow_runs | Where-Object { $_.name -eq 'End 2 End Tests' }
+          $end2endRuns = (gh api /repos/$($env:GITHUB_REPOSITORY)/actions/workflows/E2E.yaml/runs?per_page=100 | ConvertFrom-Json).workflow_runs
```
mazhelez, line 98: I still think the filter on the SHA (that the workflow file is from _main_) should be here.

Scenario is: someone runs E2E test on a branch in AL-Go (e.g. PPPreview) with SHA that is from latest _main_? This workflow run shouldn't be considered, should it?

.github/workflows/Deploy.yaml
```diff
@@ -27,49 +31,101 @@ on:
...
           $latestSha = (gh api /repos/$($env:GITHUB_REPOSITORY)/commits/main | ConvertFrom-Json).sha
-          $latestRun = $end2endruns | Where-Object { $_.head_sha -eq $latestSha }
+          $latestRun = $end2endruns | Where-Object { $_.Name -eq "End to end tests - $latestSha" } | select-object -first 1
```
mazhelez, line 100: Shouldn't we check _all_ E2E tests run that were on the latest SHA?

What if I trigger E2E tests run because I didn't notice you ran it 10 minutes ago and mine fails due to an instability, but yours succeeds? Your workflow run should be good enough for deployment.

mazhelez (commented):
One more iteration needed. See comments.

.github/workflows/Deploy.yaml
freddydk: good catch

.github/workflows/E2E.yaml
freddydk: This construct: ${{ github.event.inputs.ref != '' && github.event.inputs.ref || github.sha }}

returns github.event.inputs.ref if that is not empty - else it returns github.sha
So - it will display the sha on which the tests are run.

.github/workflows/Deploy.yaml
freddydk: Correct - it shouldn't

.github/workflows/Deploy.yaml
freddydk: I don't think so - I don't think we very very rarely will be in that situation.

I actually tried changing the code to check all - but then the code becomes more complex when trying to display which end 2 end test it is basing the verdict on. Now I can say; This end2end failed or this end2end is not complete - testing all means that I would say two has failed and one isn't complete???

.github/workflows/E2E.yaml
freddydk: Problem in your test is that your input variable is called sha - you use mine called ref.
I suggested a change to your repo.

mazhelez: True, my bad.

.github/workflows/Deploy.yaml
mazhelez: What about the following:
- If there is no E2E tests run on the latest SHA => throw "No E2E tests run found on SHA: _abcd_. Run E2E tests workflow."
- If none of the E2E tests runs on the latest SHA are "good" (rarely it has to check more than one run, but still possible) => "throw "None of the E2E tests runs on SHA _abcd_ were completed or successful. See runs at $runs.html_url"
- If there is a "good" E2E tests run => "Found successful E2E tests run on SHA abcd (link: $run.html_url)."

I imagine it gets complex because the script is inline. The alternative is a local action (I am a fan), so you can consider adding one for checking for successful E2E tests run. The other major pro of having an action instead of an inline script is that the code gets analyzed and all.

.github/workflows/Deploy.yaml
```diff
@@ -27,49 +31,101 @@ on:
...
     steps:
       - name: Check successful end 2 end tests have run
-        if: github.repository_owner == 'microsoft' && github.event.inputs.requireEndToEndTests == 'true'
+        if: github.repository_owner == 'microsoft' && needs.Inputs.outputs.requireEndToEndTests == 'true'
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
           $errorActionPreference = "Stop"
-          $end2endRuns = (gh api /repos/$($env:GITHUB_REPOSITORY)/actions/runs?per_page=100 | ConvertFrom-Json).workflow_runs | Where-Object { $_.name -eq 'End 2 End Tests' }
+          $end2endRuns = (gh api /repos/$($env:GITHUB_REPOSITORY)/actions/workflows/E2E.yaml/runs?per_page=100 | ConvertFrom-Json).workflow_runs | Where-Object { $_.head_branch -eq 'main' }
```
mazhelez, line 98: The branch (or the SHA) could be a query param.

.github/workflows/Deploy.yaml
freddydk: Typically, there will be a few runs, where the last is the 2nd or 3rd attempt, which is successful, still running or failed.
Some times, there will be a partial end 2 end test which is successful (running some of the tests) - this should be excluded.
We have never (to this date and to my knowledge) been running two end 2 end tests at the same time.
We have never (to this date and to my knowledge) been running a partial end 2 end test on the same sha as an already successful full end 2 end test.
Also, if I run end 2 end tests personally - it is usually on the SHA of my PR (not on main).
So, I think you are trying to fix a problem which doesn't exist.

I do agree that we should refactor the internal workflows (including this) to not have inline code, but I won't have the time to do a major refactoring like this at this time - we can add a slice for that work.

aholstrup1 (approved):

.github/workflows/Deploy.yaml
mazhelez: How was this addressed?

freddydk: by adding --paginate - but that was undone when trying to solve the other comment, thanks

freddydk: redone

m-zhelezova (approved):

mazhelez (approved):

aholstrup1 (approved):

PR #1077, opened 2024-05-14
AL-Go version in WorkflowInitialize

freddydk:
AL-Go version is not displayed correctly in WorkflowInitialize

Example (when using the AL-Go PPPreview)
image

When running Linux it would fail, but swallow the exception

### DIFF
--- a/Actions/WorkflowInitialize/WorkflowInitialize.ps1
+++ b/Actions/WorkflowInitialize/WorkflowInitialize.ps1
@@ -11,6 +11,8 @@ try {
 
     $actionsRef = $ENV:GITHUB_ACTION_REF
     $actionsRepo = $ENV:GITHUB_ACTION_REPOSITORY
+    Write-Host $actionsRef
+    Write-Host $actionsRepo    
     if ($actionsRepo -like 'microsoft/AL-Go-Actions') {
         Write-Host "Using AL-Go for GitHub $actionsRef"
         $verstr = $actionsRef
--- a/Actions/WorkflowInitialize/WorkflowInitialize.ps1
+++ b/Actions/WorkflowInitialize/WorkflowInitialize.ps1
@@ -9,15 +9,15 @@ try {
     . (Join-Path -Path $PSScriptRoot -ChildPath "..\AL-Go-Helper.ps1" -Resolve)
     . (Join-Path -Path $PSScriptRoot -ChildPath "..\AL-Go-TestRepoHelper.ps1" -Resolve)
 
-    # $ENV:GITHUB_ACTION_PATH is (for Direct AL-Go or preview)
-    #                                        -5     -4    -3     -2       -1
-    # linux:            /home/runner/work/freddydk/AL-Go/AL-Go/Actions/WorkflowInitialize
-    # windows:  C:\Users\runneradmin\work\freddydk\AL-Go\AL-Go\Actions\WorkflowInitialize
-    # and (for normal)
-    #                                        -3         -2          -1
-    # linux:           /home/runner/work/microsoft/AL-Go-Actions/WorkflowInitialize
-    # windows: C:\Users\runneradmin\work\microsoft\AL-Go-Actions\WorkflowInitialize
-    Write-Host $ENV:GITHUB_ACTION_PATH
+    # $ENV:GITHUB_ACTION_PATH is
+    #                                      -4       -3         -2       -1
+    # linux:            /home/runner/work/owner/AL-Go-Actions/branch/WorkflowInitialize
+    # windows:  C:\Users\runneradmin\work\owner\AL-Go-Actions\branch\WorkflowInitialize
+    #
+    # For Preview (or Direct AL-Go development branch)
+    #                                      -5    -4     -3     -2       -1
+    # linux:            /home/runner/work/owner/AL-Go/branch/Actions/WorkflowInitialize
+    # windows:  C:\Users\runneradmin\work\owner\AL-Go\branch\Actions\WorkflowInitialize
     $ap = "$ENV:GITHUB_ACTION_PATH".Split([System.IO.Path]::DirectorySeparatorChar)
     $branch = $ap[$ap.Count-2]
     $owner = $ap[$ap.Count-4]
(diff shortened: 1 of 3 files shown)


mazhelez (dismissed):
Thanks for the code comment explaining the indices! It's quite helpful.

Actions/WorkflowInitialize/WorkflowInitialize.ps1
```diff
@@ -9,18 +9,36 @@ try {
     . (Join-Path -Path $PSScriptRoot -ChildPath "..\AL-Go-Helper.ps1" -Resolve)
     . (Join-Path -Path $PSScriptRoot -ChildPath "..\AL-Go-TestRepoHelper.ps1" -Resolve)
 
-    $ap = "$ENV:GITHUB_ACTION_PATH".Split('\')
+    # $ENV:GITHUB_ACTION_PATH is
+    #                                      -4       -3         -2       -1
+    # linux:            /home/runner/work/owner/AL-Go-Actions/branch/WorkflowInitialize
+    # windows:  C:\Users\runneradmin\work\owner\AL-Go-Actions\branch\WorkflowInitialize
```
aholstrup1, line 15: For me the action path isn't consistent with / and \\. I'm getting this action path on a windows runner 🤷‍♂️  

`D:\a\_actions\aholstrup1\AL-Go\branch\Actions/WorkflowInitialize`

Maybe we need to replace all \\ with / or vice versa? To ensure it's consistent

Actions/WorkflowInitialize/WorkflowInitialize.ps1
```diff
@@ -9,18 +9,36 @@ try {
...
+    # For Preview (or Direct AL-Go development branch)
+    #                                      -5    -4     -3     -2       -1
+    # linux:            /home/runner/work/owner/AL-Go/branch/Actions/WorkflowInitialize
+    # windows:  C:\Users\runneradmin\work\owner\AL-Go\branch\Actions\WorkflowInitialize
+    $ap = "$ENV:GITHUB_ACTION_PATH".Split([System.IO.Path]::DirectorySeparatorChar)
     $branch = $ap[$ap.Count-2]
     $owner = $ap[$ap.Count-4]
-
-    if ($owner -ne "microsoft") {
+    if ($branch -eq 'Actions' -and $owner -eq 'AL-Go') {
```
aholstrup1, line 24: I'm not sure I understand this check. Could you add a small comment or just explain it in the reply to this comment :)

 Shouldn't `$ap[$ap.Count-2]` always return Actions for example?

Actions/WorkflowInitialize/WorkflowInitialize.ps1
```diff
@@ -9,18 +9,36 @@ try {
...
-    elseif ($branch -eq "preview") {
-        $verstr = "p"
-    }
     else {
-        $verstr = $branch
+        Write-Host "Using AL-Go for GitHub $branch"
+        if ($owner -ne "microsoft") {
+            $verstr = "d"
+        }
+        elseif ($branch -eq "preview") {
```
aholstrup1, line 36: Al-Go preview will usually use a commit sha instead of "preview". I wonder if we should handle that scenario as well / instead?

Actions/WorkflowInitialize/WorkflowInitialize.ps1
freddydk: Will just split('/\') - that should take care of this.

Actions/WorkflowInitialize/WorkflowInitialize.ps1
freddydk: fixed

Actions/WorkflowInitialize/WorkflowInitialize.ps1
freddydk: See comments above - $ap[$ap.count-2] will be the Actions subfolder if using direct AL-Go or preview.
It will be the branch if pointing to AL-Go-Actions

Actions/WorkflowInitialize/WorkflowInitialize.ps1
```diff
@@ -9,18 +9,43 @@ try {
...
+    #                                      -4       -3         -2       -1
+    # linux:            /home/runner/work/owner/AL-Go-Actions/branch/WorkflowInitialize
+    # windows:  C:\Users\runneradmin\work\owner\AL-Go-Actions\branch\WorkflowInitialize
+    #
+    # For Preview (or Direct AL-Go development branch)
+    #                                      -5    -4     -3     -2       -1
+    # linux:            /home/runner/work/owner/AL-Go/branch/Actions/WorkflowInitialize
+    # windows:  C:\Users\runneradmin\work\owner\AL-Go\branch\Actions\WorkflowInitialize
+    # or:       C:\Users\runneradmin\work\owner\AL-Go\branch\Actions/WorkflowInitialize
+    $ap = "$ENV:GITHUB_ACTION_PATH".Split('/\')
```
mazhelez, line 22: Just out of curiosity, why not use `github.action_ref` and `github.action_repository` (from https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) instead of splitting the action path? Seems a bit cleaner and less likely to break (in case GitHib go wild and change the action path structure).

Then the check below might not be needed (referring to `if ($branch -eq 'Actions' -and $owner -eq 'AL-Go')`).

As a minimum, consider renaming `$branch` to `$ref` as it's indeed a ref.

mazhelez (approved):
Looks okay. Left a suggestion that will reduce the complexity a bit and make the code more robust. However, up to you to implement it or not.

Actions/WorkflowInitialize/WorkflowInitialize.ps1
freddydk: Good point - these didn't exist earlier but I remember that you also suggested using them in a different situation.
The code became a lot simpler with this, thanks.

mazhelez (approved):

Actions/WorkflowInitialize/WorkflowInitialize.ps1
freddydk: and... - they are actually not available as ENV variables - they are as github.action_ref and github.action_repository in action.yaml.
Fixed this now - looks much simpler-

mazhelez (approved):

aholstrup1 (approved):

PR #1097, opened 2024-06-05
Add support for custom template files

cegekaJG:
This PR allows users to define a list of paths using the new `templateFiles` parameter in the repository settings. Using the existing file comparison of the action "CheckForUpdates", any files matched will be compared with their template counterpart and updated, if possible.
This PR also updates the `settings.md` file and JSON schema to include the new parameter, as well as a new test case to test whether given paths are being properly interpreted.


freddydk:
Thanks - we wíll have a look at this and discuss the overall approach to custom templates (see also https://github.com/microsoft/AL-Go/pull/853) to see what the right approach is.

freddydk:
@cegekaJG - Am I right in assuming that you are using your own fork of AL-Go for GitHub?
And then you want to sync additional files from that repo to your repositories?
Is this correct? is this what you are trying to achieve?

Problem with the approach to have this in the repo settings is that you have to have this in all repositories (or in variables) - it should really be part of the template.

I think it is a good idea to have a list of additional folders and files (when having f.ex. indirect templates), but I think this should be part of the template repo - I will include this requirement in the indirect template PR, which I will complete after the summer vacation.

cegekaJG:
I agree, but this was the easiest way I could think of without affecting old repositories. Additionally, it's important to differentiate between files you want to copy from the template at the start, and files you always want to sync with the template.
If I am using custom deploy scripts, I'd want all repos to use the master version, but the workspace config file obviously needs to differ from the template.

freddydk:
Have added a link to this PR in the other PR comments - will take the scenario into account there

PR #1108, opened 2024-06-19
Update logic for Increment Version Number

### LINKED ISSUE #1105: Increment Version Number - repoVersion in .github/AL-Go-Settings.json is not updated
**Problem**
If you define the repoVersion in .github/AL-Go-Settings.json it isn't updated when running the Increment Version Number workflow. 

In our case, the would like to keep the repoVersion the same across all Al-Go projects so having repoVersion defined on repo level makes sense for us (even if it is a project setting)

aholstrup1:
**Problem**
If you define the repoVersion in .github/AL-Go-Settings.json it isn't updated when running the Increment Version Number workflow. In our case, the would like to keep the repoVersion the same across all Al-Go projects so having repoVersion defined on repo level makes sense for us (even if it is a project setting)

**Proposed Solution**
* If repoVersion exists on project level (recommended), we only update it there.
* If repoVersion doesn't exist in project level but exists on repo level, we update it on repo level (what we want in our case)
* if repoVersion is neither defined on project level or repo level, force create it on project level

Fixes #1105


Actions/IncrementVersionNumber/IncrementVersionNumber.ps1
```diff
@@ -47,23 +47,35 @@ try {
...
-
-            # Set repoVersion in project settings according to the versionNumber parameter
-            Set-VersionInSettingsFile -settingsFilePath $projectSettingsPath -settingName 'repoVersion' -newValue $versionNumber
+            if (Test-SettingExists -settingsFilePath $projectSettingsPath -settingName 'repoVersion') {
+                # If 'repoVersion' exists in the project settings, update it there
+                Set-VersionInSettingsFile -settingsFilePath $projectSettingsPath -settingName 'repoVersion' -newValue $versionNumber
+            } elseif ($repoVersionExistsInRepoSettings) {
+                # If 'repoVersion' is not found in project settings but it exists in repo settings, update it there instead
+                Write-Host "Setting 'repoVersion' not found in $projectSettingsPath. Updating it on repo level instead"
+                $updateRepoVersionInRepoSettings = $true
```
mazhelez, line 72: At this point the version isn't set, so line 81 (`$projectSettings = ReadSettings -baseFolder $baseFolder -project $project` ) will fetch the old version. That might have effect in `Set-VersionInAppManifests`.

Can you please verify?

mazhelez (dismissed):
Looks okay.

Is there a way to cover the functionality in tests?

freddydk:
Looking at the PR description:
- if repoVersion is neither defined on project level or repo level, force create it on repo level
- if projects is * then we update the repoVersion both on project level and repo level (if they already exist)

According to the code - it looks like it is force created on the project level if it doesn't exist (which I think is correct) - meaning that the description is wrong, right?

On the last line - I don't think we are updating the repoversion on the repo level if it exists in all projects - do we?

aholstrup1:
> Looking at the PR description:
> 
> * if repoVersion is neither defined on project level or repo level, force create it on repo level
> * if projects is * then we update the repoVersion both on project level and repo level (if they already exist)
> 
> According to the code - it looks like it is force created on the project level if it doesn't exist (which I think is correct) - meaning that the description is wrong, right?
> 
> On the last line - I don't think we are updating the repoversion on the repo level if it exists in all projects - do we?

You're right :) The description was wrong. It should reflect the code now.

mazhelez (approved):

freddydk (changes requested):
Please update releasenotes with a comment on this change (or at least the issue fixed)

freddydk (dismissed):

freddydk (approved):

mazhelez (approved):

freddydk (approved):

mazhelez (approved):

PR #1116, opened 2024-06-24
Support deployment scope

### LINKED ISSUE #866: Accessing GitHub Environment Variables in DeployToCustom Scenarios for PowerShell Scripts
Is there a way to make github environment variables accessible in DeployToCustom scenarios? In on OnPrem scenarios we need, insteaf of an AuthContext an Instance Name of Business Central to know the target of our deployment. I could misuse "AuthContext" but this doesn't look like a technical sound solution.

We have for example variables like "BC_Instance" that we need in "DeployToCustom.ps1" script to execute an deployment. I tried to find a way in the AL-Go [Deploy.ps1](https://github.com/microsoft/AL-Go/blob/main/Actions/Deploy/Deploy.ps1) but did not find a viable way.

I assumed this code does the trick but unfortunately $env: is not the github environment in this context
` $secrets = $env:Secrets | ConvertFrom-Json ` (Line 29)

This is the point at which the custom script "DeployTo..." is called, unfortunately there is no option there either. So I currently have to call the environment variables in my own script via the GitHub API. This requires the repository ID (available), the environment name (available) and the GitHub workflow token. Unfortunately, the token is not available. Accordingly, this is not a viable option for me either. Do you have an approach for me here?
```
$customScript = Join-Path $ENV:GITHUB_WORKSPACE ".github/DeployTo$($deploymentSettings.EnvironmentType).ps1"
if (Test-Path $customScript) {
    Write-Host "Executing custom deployment script $customScript"
    $parameters = @{
        "type" = $type
        "AuthContext" = $authContext
        "Apps" = $apps
    } + $deploymentSettings
    . $customScript -parameters $parameters
```

Suggestion: github environment variables are passed through to "DeployToCustom.ps1"

### LINKED ISSUE #980: Allow Scope to be PTE in continuousDeployment for PTE extensions in Sandbox (enhancement request)
## Use Case

Developer asks new online environment on the customer tenant.
This should be a copy from the current `UAT` environment.
Since the current `UAT` is part of the CI/CD deployment, the latest version of our app has been published in sope `Dev`.

image

When creating a copy from the current `UAT`, all apps, published in scope `Dev` are removed in the new environment that is created.

If there are multiple PTE-apps installed in `Dev` scope, it is quit a hassle to reinstall all apps in the new environment.

## Request

add a setting in the `DeployTo<environmentname>` that allows us to define the scope, e.g.: **`"scope": "PTE"`**

```json
"DeployToUAT": {
    "EnvironmentName": "UAT",
    "continuousDeployment": true,
    "scope": "PTE"
  }
```

### LINKED ISSUE #1079: AppSource App deployment failes with PerTenantExtensionCop Error PTE0001 and PTE0002
I have an app that was generated from the generated from [AL-Go-AppSource](https://github.com/microsoft/AL-Go-AppSource) template.

When the CICD runs we get an error during deployment, telling: `Deploying to UAT failed. Unable to publish app. Please open the Extension Deployment Status Details page in Business Central to see the detailed error message.`

image

In the `Status Details page` of Business Central I get these error messages:
```log
Package validation failed due to the following error(s): src/codeunit/NALRaptEmailMgtFunctions.Codeunit.al (4,10) - Error PTE0001: Codeunit 'NAL Rapt. Email Mgt. Functions' has an ID of [2057753]. It must be within the range '[50000..99999]'.
src/codeunit/NALRaptFwdLinkMgt.Codeunit.al (5,10) - Error PTE0001: Codeunit 'NAL Rapt. Fwd Link Mgt.' has an ID of [2057754]. It must be within the range '[50000..99999]'.
src/tableextension/NALRaptPPInvViewDoc.TableExt.al (5,15) - Error PTE0002: Field 'NAL Raptor Batch Run' has an ID of [2057742]. It must be within the range '[50000..99999]'.
src/tableextension/NALRaptPPInvViewDoc.TableExt.al (10,15) - Error PTE0002: Field 'NAL Raptor Batch Run Date' has an ID of [2057743]. It must be within the range '[50000..99999]'.
...
```

Any idea what caused this error message?

### LINKED ISSUE #1083: SyncMode for custom deployments?
Do the parameters passed to the custom deployment script `.github/DeployTo{environmentType}` include `SyncMode`? The current documentation doesn't mention it:
https://github.com/microsoft/AL-Go/blob/6a9ef0c91e5b54d0fd3abb3d4495417a1db28f17/Scenarios/settings.md?plain=1#L253-L265
However, the `Actions/Deploy/Deploy.ps1` code does append the hashtable `$deploymentParameters`, which does seem to include the parameter in question.
If the parameter really isn't included, it really should be - it's just as necessary for OnPrem environments. If it is, the documentation needs to be updated.

On top of that, could you add a way to override the sync mode for a single deployment? Maybe as a parameter when manually triggering an action that includes the job. When using ForceSync, it generally is a one-time thing and leaving it enabled in a config file could cause problems, so letting developers trigger it explicitly outside of the repo files would make it easier and safer.

### LINKED ISSUE #1109: Why filter deployment settings?
In the action "DetermineDeploymentEnvironments", a step defines a default list of parameters and overwrites them with custom values if they are defined in the configuration.

https://github.com/microsoft/AL-Go/blob/c1020361df06b369c3fed56cefced85e1709f03b/Actions/DetermineDeploymentEnvironments/DetermineDeploymentEnvironments.ps1#L139-L165

Wouldn't it be better, however, to just add every property of the deployment config to the environment settings? That way, it's possible to define custom parameters that can be used by custom deployment scripts.
I don't know if there is any code that relies on a known quantity of properties, so I don't see a risk in allowing this.

freddydk:
Add scope property to DeployTo settings and allow custom properties in DeployTo

Fixes #1079
Fixes #980
Fixes #1083
Fixes #866
Fixes #1109


Actions/Deploy/Deploy.ps1
```diff
@@ -19,12 +19,13 @@ $deploymentEnvironments = $deploymentEnvironmentsJson | ConvertFrom-Json | Conve
...
 $envName = $environmentName.Split(' ')[0]
 $secrets = $env:Secrets | ConvertFrom-Json
+$settings = $env:Settings | ConvertFrom-Json
 
 $authContext = $null
 foreach($secretName in "$($envName)-AuthContext","$($envName)_AuthContext","AuthContext") {
     if ($secrets."$secretName") {
         $authContext = [System.Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String($secrets."$secretName"))
-        break
+            break
```
aholstrup1, line 28:
```suggestion
        break
```

aholstrup1 (dismissed):

aholstrup1 (approved):

aholstrup1 (approved):

mazhelez (approved):

PR #1120, opened 2024-06-25
Remove unused parameter in Deploy.ps1

cegekaJG:
This PR removes the parameter `$token` from "Deploy.ps1", which currently isn't being read by any script, or specified by any workflow.


Actions/Deploy/Deploy.ps1
```diff
@@ -1,6 +1,4 @@
 Param(
-    [Parameter(HelpMessage = "The GitHub token running the action", Mandatory = $false)]
-    [string] $token,
```
mazhelez, line 3: It's not used in the script but it's passed in the action YAML (Actions/Deploy/action.yaml).

cegekaJG: You're correct, I missed that. But no workflow overrides the default value and the token still goes unused, so shouldn't it go? Or is there something on the roadmap that would require it?

Actions/Deploy/action.yaml
```diff
@@ -5,10 +5,6 @@ inputs:
     description: Shell in which you want to run the action (powershell or pwsh)
     required: false
     default: powershell
-  token:
```
mazhelez, line 8: Lines 32 and 40 need to be cleaned up as well. Also, run the tests in Tests\Deploy.Action.Test.ps1 locally to verify they pass.

Actions/Deploy/Deploy.ps1
freddydk: It also needs to be removed from the README.md (in the Actions/Deploy folder)

Actions/Deploy/action.yaml
freddydk: @cegekaJG - there is a CI test failing. That test checks whether the .ps1 script matches the .yaml file with parameters etc. once you have removed the token in the lines mentioned by Maria, the test will pass.

cegekaJG: My mistake, it should work now. The test was successful as well.

freddydk (changes requested):
@cegekaJG please remove `token` from README.md in Actions/Deliver

freddydk (approved):

PR #1122, opened 2024-06-26
Use github.sha instead of github.ref in reusable workflows

mazhelez:


Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
```diff
@@ -93,7 +93,6 @@ jobs:
     with:
       shell: ${{ needs.Initialization.outputs.githubRunnerShell }}
       runsOn: ${{ needs.Initialization.outputs.githubRunner }}
-      checkoutRef: refs/pull/${{ github.event.number }}/merge
```
freddydk, line 96: Is this really correct???

freddydk: I think we went back and forth a few times back then to find this

Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
```diff
@@ -38,6 +38,7 @@ jobs:
       telemetryScopeJson: ${{ steps.init.outputs.telemetryScopeJson }}
       githubRunner: ${{ steps.ReadSettings.outputs.GitHubRunnerJson }}
       githubRunnerShell: ${{ steps.ReadSettings.outputs.GitHubRunnerShell }}
+      checkoutRef: ${{ github.event.pull_request.merge_commit_sha }}
       projects: ${{ steps.determineProjectsToBuild.outputs.ProjectsJson }}
```
freddydk, line 42: I think you placed this in the wrong location (under outputs) - should be a parameter for _BuildALGo...

freddydk (changes requested):

Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
mazhelez: Yes, yes... figured it out the long way 😄

freddydk:
I guess this should also resolve the bug we have with PR handler fails if the PR is merged before completing - or was this already resolved earlier?

freddydk (approved):

aholstrup1 (approved):

PR #1132, opened 2024-07-03
Skip Pull Request check in draft mode

### LINKED ISSUE #1051: [Feature Request] Don't trigger Pull Request Status Check for drafts
I tend to create a PR very early to give my team and myself an overview of what I'm working on, but I don't need to run the build pipeline until I'm ready.
The "Pull Request Status Check" workflow shouldn't trigger if a PR is still a draft, otherwise there's no way to prevent the workflow from running after creating a PR. If other users rely on status checks for drafts (not sure if that's the case, skipping status checks until the PR is ready for review is the default behavior), the AL Go settings could include a flag that adds a condition for the workflow after running a system file update.

cegekaJG:
This PR adds a condition to "PullRequestHandler.yaml" that ensures pull requests in draft mode won't trigger the status check. This reduces unnecessary build workflows for feature branches. Additionally, the PR explicitly lists all types of the `pull_request_target` event. Listing the types isn't supposed to confirm each type as a valid reason to trigger the workflow, but rather to invite maintainers to judge which events can be ignored for this workflow.
Resolves #1051

### DIFF
--- a/Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
+++ b/Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
@@ -37,7 +37,7 @@ jobs:
 
   Initialization:
     needs: [ PregateCheck ]
-    if: (!failure() && !cancelled())
+    if: (!failure() && !cancelled()) && (github.event.pull_request.draft == false)
     runs-on: [ windows-latest ]
     outputs:
       telemetryScopeJson: ${{ steps.init.outputs.telemetryScopeJson }}
@@ -87,7 +87,7 @@ jobs:
 
   Build:
     needs: [ Initialization ]
-    if: (!failure()) && (!cancelled()) && fromJson(needs.Initialization.outputs.buildOrderJson)[0].projectsCount > 0
+    if: (!failure()) && (!cancelled()) && fromJson(needs.Initialization.outputs.buildOrderJson)[0].projectsCount > 0 && (github.event.pull_request.draft == false)
     strategy:
       matrix:
         include: ${{ fromJson(needs.Initialization.outputs.buildOrderJson)[0].buildDimensions }}
@@ -111,7 +111,7 @@ jobs:
 
   StatusCheck:
     needs: [ Initialization, Build ]
-    if: (!cancelled())
+    if: (!cancelled()) && (github.event.pull_request.draft == false)
     runs-on: [ windows-latest ]
     name: Pull Request Status Check
     steps:
--- a/Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
+++ b/Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
@@ -3,6 +3,24 @@ name: 'Pull Request Build'
 on:
   pull_request_target:
     branches: [ 'main' ]
+    types:
+      - assigned
+      - unassigned
+      - labeled
+      - unlabeled
+      - opened
+      - edited
+      - closed
+      - reopened
+      - synchronize
+      - converted_to_draft
+      - ready_for_review
+      - locked
+      - unlocked
+      - review_requested
+      - review_request_removed
+      - auto_merge_enabled
+      - auto_merge_disabled
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
--- a/Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
+++ b/Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
@@ -25,7 +25,7 @@ env:
 
 jobs:
   PregateCheck:
-    if: (github.event.pull_request.base.repo.full_name != github.event.pull_request.head.repo.full_name) && (github.event_name != 'pull_request')
+    if: (github.event.pull_request.base.repo.full_name != github.event.pull_request.head.repo.full_name) && (github.event_name != 'pull_request') && (github.event.pull_request.draft == false)
     runs-on: windows-latest
     steps:
       - uses: microsoft/AL-Go-Actions/VerifyPRChanges@main
--- a/Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
+++ b/Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
@@ -4,23 +4,10 @@ on:
   pull_request_target:
     branches: [ 'main' ]
     types:
-      - assigned
-      - unassigned
-      - labeled
-      - unlabeled
-      - opened
-      - edited
-      - closed
-      - reopened
       - synchronize
-      - converted_to_draft
+      - opened
+      - reopened
       - ready_for_review
-      - locked
-      - unlocked
-      - review_requested
-      - review_request_removed
-      - auto_merge_enabled
-      - auto_merge_disabled
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number }}


Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
```diff
@@ -25,7 +43,7 @@ env:
 
 jobs:
   PregateCheck:
```
mazhelez, line 45: The `PregateCheck` exists to fail external PRs (from forks) that edit certain files.

With the new check the check will not be triggered for draft PRs, which is not desirable behavior.

Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
```diff
@@ -3,6 +3,24 @@ name: 'Pull Request Build'
 on:
   pull_request_target:
     branches: [ 'main' ]
+    types:
+      - assigned
```
mazhelez, line 7: From https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

> By default, a workflow only runs when a pull_request_target event's activity type is opened, synchronize, or reopened

This list contains the events that would trigger a new Pull Request Build run. Why would you want a new build triggered when the PR is labeled, or assigned?

cegekaJG: Oh, sorry, my mistake. I assumed the default array of types would be all of them, not just those. I do want to include `ready_for_review`, because that's when the PR exits draft mode.

Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
cegekaJG: I don't see the sense in running status checks for PRs in draft mode. Even if the PR was editing blacklisted files, there's no risk of introducing breaking changes since it's not ready to be merged to begin with, right?

mazhelez: Running status checks on draft have some advantages.

E.g. some developers prefer to have green build _before_ setting the PR as ready to review. Others, want to get feedback as soon as they start developing.

Since GitHub decided to run PR workflows even on draft PRs, we decided that there isn't anything to lose there.

Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
cegekaJG: Assuming drafts don't trigger the PR handler, there wouldn't be a successful PR build, hence the need to trigger as soon as the PR leaves draft state. I agree that you would trigger a superfluous build by setting a ready PR to draft and back, but that's an edge case.

Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
cegekaJG: My motivation for this change is the fact that runtimes can be very long and I want to save as much compute time on GitHub's runners as possible. I understand that this can be mitigated by pushing as little as possible, but I'd like our developers to have the ability to keep their remote branches up to date without restrictions. Additionally, we have custom workflows that automate some processes by preparing pull requests, and those can end up in draft mode if manual input is required - any status checks would be pointless at this stage, so I'd like to avoid them.

mazhelez: That's a valid point, but, as I said, that might not be aligned with what other developers what to experience.

So, generally you want to following scenarios to be supported:
1. When a draft PR is opened, no PR Build is triggered.
2. When a draft PR is converted to "ready for review", a PR Build is triggered.
3. When a non-draft PR is opened, a PR Build is triggered.

With your changes, 1. will not be satisfied, because GitHub doesn't differentiate between draft and non-draft PRs for the "opened" trigger. (Feel free to test that).
If a PR Build isn't triggered when the PR is opened, then how would you cover 3.?

Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
mazhelez: From the other comment, how would not trigger PR Builds on draft PRs?

Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
cegekaJG: Oh, I didn't realize the `draft` parameter of a newly opened pull request always evaluates to false. In that case, it's impossible to prevent premature status checks.

mazhelez: Maybe not prevent it, but make it not run the whole build. But that requires different effort.

cegekaJG: Would adding a draft condition to the build job suffice? Or are there other sections I need to adapt?

cegekaJG: Sorry, but I've tested again and GitHub does in fact differentiate between draft and non-draft PRs on the "opened" trigger.
The following workflow will be skipped if a new pull request is opened in draft mode:

```yml
name: Run Pre-Commit Hooks

on:
    pull_request:
        types: [opened, synchronize, reopened]
        branches:
            - main

jobs:
    pre-commit:
        name: Run Pre-Commit
        if: github.event.pull_request.draft == false
        runs-on: ubuntu-latest

        steps:
            - name: Checkout code
              uses: actions/checkout@v4
```

mazhelez: Then how is it different from now?

> By default, a workflow only runs when a pull_request event's activity type is **opened**, **synchronize**, or **reopened**

Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
cegekaJG: I've now updated all jobs of the workflow with the draft check. If there's some job that shouldn't be skipped due to a particular reason, let me know and we can think of a solution.

Templates/Per Tenant Extension/.github/workflows/PullRequestHandler.yaml
mazhelez: Can you elaborate a bit? Is the workflow skipped or is it never triggered?

My question was in regards of the triggers. If the default trigger types are _opened_, _synchronize_, or _reopened_, then why is the PR Build triggered on draft PRs in the first place?

cegekaJG: The triggers don't consider the status of the PR so the workflow is initiated all the same, but the jobs will be skipped because the `if`-condition isn't met.

mazhelez: Oh, I see :)
That was the solution we would have gone with.

cegekaJG:
https://github.com/microsoft/AL-Go/actions/runs/10140000167
Does the test fail because the templates are using microsoft/AL-Go@main as the template URL? Running the tests locally doesn't result in an error.

freddydk:
> https://github.com/microsoft/AL-Go/actions/runs/10140000167 Does the test fail because the templates are using microsoft/AL-Go@main as the template URL? Running the tests locally doesn't result in an error.

The error is because the PullRequestHandler needs to be identical in the AppSource App template folder and in the Per Tenant Extension folder.

Your PR only contains changes to the PTE folder.

freddydk:
I think the PR makes sense.
Being able to avoid builds while working on a PR in draft mode, can save a lot of minutes while you don't want them.

Big question is whether it should be possible to run builds on draft mode? with this PR, I guess it is always blocked?

cegekaJG:
I do agree there needs to be a way to allow for the pipeline to run in draft mode. We could either run the workflow partially and exit as soon as possible, or supplement Update-Workflow with the ability to append or remove the draft condition.

freddydk:
> there needs to be a way to allow for the pipeline to run in draft mode. We could either run the workflow partially and exit as soon as possible, or supplement Update-Workflow with the ability to append or remove the draft condition.

We could also create a setting called runPullRequestBuildOnDraftPR and default it to false?
Then most people wouldn't change that unless they really want this.

aholstrup1:
Closing this Pull Request for now since it's been stale for a while. In order to merge this PR, we would want this behavior to be configurable as mentioned in https://github.com/microsoft/AL-Go/pull/1132#issuecomment-2309879010

The feature suggestion #1051 is still in our backlog so the suggestion is not forgotten just because the PR will be closed now :)

PR #1145, opened 2024-07-16
Fix placeholder syntax in PR Handler action

cegekaJG:
This PR fixes an error introduced by #1130.
The new reference values use the `format()` command to generate a git reference, but the placeholder syntax was incorrect.
The two instances of `${0}` per template have been replaced with `{0}`.

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -9,6 +9,7 @@
 - Issue 1109 Why filter deployment settings?
 - Fix issue with github ref when running reusable workflows
 - Issue 1098 Support for specifying the name of the AZURE_CREDENTIALS secret by adding a AZURE_CREDENTIALSSecretName setting
+- Fix placeholder syntax for git ref in PullRequestHandler.yaml
 
 ### Dependencies to PowerShell modules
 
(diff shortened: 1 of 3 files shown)


mazhelez (dismissed):
Maybe also add a line in the release notes.

cegekaJG:
> Maybe also add a line in the release notes.

Do you mean a line explaining this PR in the "Issues"-section? I figured it wouldn't be necessary since this only fixes a change that is still only in the preview.

mazhelez:
> > Maybe also add a line in the release notes.
> 
> Do you mean a line explaining this PR in the "Issues"-section? I figured it wouldn't be necessary since this only fixes a change that is still only in the preview.

Exactly because it is already shipped in _preview_. Repositories running on _preview_ will get the `There are updates for your AL-Go system, run 'Update AL-Go System Files' workflow to download the latest version of AL-Go.` message and the release notes should explain what these changes are.

It's not a requirement, but a recommendation.

freddydk (dismissed):

mazhelez (approved):

freddydk (approved):

PR #1146, opened 2024-07-17
Allow "ubuntu-latest" runners for all workflows

cegekaJG:
This pull request adds a check to allow updating the `runs-on` and `shell` parameters of all workflows. Previously, the workflows "Update AL Go System Files" and "Pull Request Handler" were excluded when changing the two parameters to prevent deadlocks and security issues.
With this change, a check will only allow changing the parameters of the blacklisted workflows if the `runs-on` parameter is either `windows-latest` or `ubuntu-latest`. Additionally, a check in the function `ModifyRunsOnAndShell` ensures only the value `pwsh` is allowed when using an `ubuntu-latest` runner.

### DIFF
--- a/.github/workflows/E2E.yaml
+++ b/.github/workflows/E2E.yaml
@@ -200,7 +200,19 @@ jobs:
           Write-Host "repoName='$repoName'"
           Write-Host "Repo URL: https://github.com/${{ needs.Check.outputs.githubowner }}/$repoName"
 
-      - name: Run test
+      - name: Run test on Linux
+        run: |
+          try {
+            . (Join-Path "." "e2eTests/scenarios/${{ matrix.scenario }}/runtest.ps1") -github -linux -githubOwner '${{ needs.Check.outputs.githubowner }}' -repoName ${{ steps.calculateParams.outputs.repoName }} -token '${{ Secrets.E2EPAT }}' -pteTemplate '${{ needs.Check.outputs.githubowner }}/${{ needs.SetupRepositories.outputs.perTenantExtensionRepo }}' -appSourceTemplate '${{ needs.Check.outputs.githubowner }}/${{ needs.SetupRepositories.outputs.appSourceAppRepo }}' -adminCenterApiToken '${{ Secrets.adminCenterApiCredentials }}'
+          }
+          catch {
+            Write-Host $_.Exception.Message
+            Write-Host $_.ScriptStackTrace
+            Write-Host "::Error::$($_.Exception.Message)"
+            $host.SetShouldExit(1)
+          }
+
+      - name: Run test on Windows
         run: |
           try {
             . (Join-Path "." "e2eTests/scenarios/${{ matrix.scenario }}/runtest.ps1") -github -githubOwner '${{ needs.Check.outputs.githubowner }}' -repoName ${{ steps.calculateParams.outputs.repoName }} -token '${{ Secrets.E2EPAT }}' -pteTemplate '${{ needs.Check.outputs.githubowner }}/${{ needs.SetupRepositories.outputs.perTenantExtensionRepo }}' -appSourceTemplate '${{ needs.Check.outputs.githubowner }}/${{ needs.SetupRepositories.outputs.appSourceAppRepo }}' -adminCenterApiToken '${{ Secrets.adminCenterApiCredentials }}'
--- a/.github/workflows/E2E.yaml
+++ b/.github/workflows/E2E.yaml
@@ -181,7 +181,7 @@ jobs:
           Add-Content -Encoding UTF8 -Path $env:GITHUB_OUTPUT -Value "scenarios=$scenariosJson"
           Write-Host "scenarios=$scenariosJson"
 
-  Scenario:
+  ScenariosOnWindows:
     runs-on: [ windows-latest ]
     needs: [ Check, SetupRepositories, Analyze ]
     if: github.event.inputs.runScenarios == 'true'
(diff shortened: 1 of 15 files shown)


Actions/CheckForUpdates/CheckForUpdates.HelperFunctions.ps1
```diff
@@ -271,6 +274,9 @@ function GetWorkflowContentWithChangesFromSettings {
     $baseName = [System.IO.Path]::GetFileNameWithoutExtension($srcFile)
     $yaml = [Yaml]::Load($srcFile)
     $workflowScheduleKey = "$($baseName)Schedule"
+    $workflowBlacklist = @('UpdateGitHubGoSystemFiles', 'Troubleshooting')
```
freddydk, line 277: I would move these lines down to the section where they are used to improve readability

Actions/CheckForUpdates/CheckForUpdates.HelperFunctions.ps1
```diff
@@ -289,7 +295,14 @@ function GetWorkflowContentWithChangesFromSettings {
         ModifyPullRequestHandlerWorkflow -yaml $yaml -repoSettings $repoSettings
     }
 
-    if ($baseName -ne "UpdateGitHubGoSystemFiles" -and $baseName -ne 'Troubleshooting') {
+    # Workflows in the blacklist may only run on runners listed in the whitelist
+    if($workflowBlacklist -contains $baseName) {
+        if($runnerWhitelist -notcontains $repoSettings."runs-on") {
+            $modifyRunsOnAndShell = $false
+        }
+    }
```
freddydk, line 303: I am not sure that the end 2 end tests covers this scenario. If Update AL-Go system files is set to Ubuntu-latest and it for some reason doesn't work - people will be dead in the water.
I will try to run end 2 end tests on this PR to see whether it is covered.

lippertmarkus: This wouldn't allow people to use self-hosted runners for AL-Go updates and Troubleshooting

lippertmarkus (changes requested):

Actions/CheckForUpdates/CheckForUpdates.HelperFunctions.ps1
freddydk: That is already the case today. These workflows can only run on GitHub hosted agents to avoid deadlocks if self hosted agents are not working (where they need to run update al-go system files to resolve)

RELEASENOTES.md
```diff
@@ -23,6 +23,11 @@ All authentication context secrets now supports managed identities and federated
 
 In the summary after a Test Run, you now also have the result of performance tests.
 
+### Ubuntu runners for all AL Go workflows
+
+Previously, the workflows "Update AL Go System Files" and "Pull Request Handler" were excluded when changing the two parameters to prevent deadlocks and security issues.
+From now on, a check will only allow changing the parameters of the blacklisted workflows if the `runs-on` parameter is either `windows-latest` or `ubuntu-latest`. Additionally, a check in the function `ModifyRunsOnAndShell` ensures only the value `pwsh` is allowed when using an `ubuntu-latest` runner.
```
mazhelez, line 29:
```suggestion
From now on, a check will only allow changing the parameters of the excluded workflows if the `runs-on` parameter is either `windows-latest` or `ubuntu-latest`. Additionally, only value `pwsh` for `shell` setting is allowed when using an `ubuntu-latest` runner (`runs-on` setting).
```

Actions/CheckForUpdates/CheckForUpdates.HelperFunctions.ps1
freddydk: I was right in assuming that no end 2 end test will cover the scenario that Update AL-Go System Files actually works on ubuntu-latest. I will create a PR (on your PR) with an end 2 end test

freddydk: Have modified end 2 end tests to use ubuntu for these workflows as well - and run all scenario tests on windows and linux - running end 2 end now - if they pass, I am OK with this.

freddydk: I think I have been running too many tests and exhausted my limits - will re-run later

freddydk: The Update AL-Go System Files actually didn't work properly under Linux due to this Powershell "bug": https://github.com/PowerShell/PowerShell/issues/6473#issuecomment-375930843

Under Linux, folders starting with a period are considered hidden and Resolve-Path cannot resolve a hidden path.
I have added a workaround to the PR and re-running end 2 end tests now.

freddydk: The consequence of this bug was that scripts and files directly under .github or under .AL-Go would not have been updated by Update AL-Go System Files.
Running a full end 2 end now https://github.com/microsoft/AL-Go/actions/runs/10088558197

freddydk (approved):

.github/workflows/E2E.yaml
```diff
@@ -196,14 +196,45 @@ jobs:
...
+            . (Join-Path "." "e2eTests/scenarios/${{ matrix.scenario }}/runtest.ps1") -github -githubOwner '${{ needs.Check.outputs.githubowner }}' -repoName '${{ steps.calculateParams.outputs.repoName }}' -token '${{ Secrets.E2EPAT }}' -pteTemplate '${{ needs.Check.outputs.githubowner }}/${{ needs.SetupRepositories.outputs.perTenantExtensionRepo }}' -appSourceTemplate '${{ needs.Check.outputs.githubowner }}/${{ needs.SetupRepositories.outputs.appSourceAppRepo }}' -adminCenterApiToken '${{ Secrets.adminCenterApiCredentials }}'
+          }
+          catch {
+            Write-Host $_.Exception.Message
+            Write-Host $_.ScriptStackTrace
+            Write-Host "::Error::$($_.Exception.Message)"
+            $host.SetShouldExit(1)
+          }
+
+  ScenariosOnLinux:
```
mazhelez, line 215: Minor: consider extracting in a local action to avoid coffee duplication.

e2eTests/scenarios/FederatedCredentials/runtest.ps1
```diff
@@ -44,6 +45,11 @@ Write-Host -ForegroundColor Yellow @'
 
 $errorActionPreference = "Stop"; $ProgressPreference = "SilentlyContinue"; Set-StrictMode -Version 2.0
 
+if ($linux) {
+    Write-Host 'This test is using bcsamples-bingmaps.appsource and should only run once (either Windows or Linux)'
```
mazhelez, line 49: Wouldn't that mean that this scenario will always be run only on Windows?

mazhelez (approved):

e2eTests/scenarios/FederatedCredentials/runtest.ps1
freddydk: This scenario runs on whatever bcsamples-bingmaps.appsource is setup to run on, but for federated credentials, there really are no Linux/Windows specific things.

.github/workflows/E2E.yaml
freddydk: Yeah, thought about that - these .yaml files could use some refactoring - maybe have Linux,Windows as a dimension in the matrix - but this one already took more time than I had.
In the end - the bug found would also affect Check For Updates running on Linux - so all good.

PR #1172, opened 2024-08-19
Issue 1164

### LINKED ISSUE #1164: [Bug]: Gertting secrets from Azure key vault fails in Preview
### AL-Go version

preview

### Describe the issue

As soon as I switched to preview, the Initialize step fails when it tries to read secrets from the key vault. 
If I switch back to main, it works again.

### Expected behavior

Initialize should not fail

### Steps to reproduce

Run preview CI/CD with a connected azure key vault

image

freddydk:
Fixes #1164

### DIFF
--- a/Actions/ReadSecrets/ReadSecrets.ps1
+++ b/Actions/ReadSecrets/ReadSecrets.ps1
@@ -26,7 +26,7 @@ try {
 
     $outSecrets = [ordered]@{}
     $settings = $env:Settings | ConvertFrom-Json | ConvertTo-HashTable
-    $keyVaultCredentials = GetKeyVaultCredentials -testSecretName 'algotestsecret'
+    $keyVaultCredentials = GetKeyVaultCredentials
     $getAppDependencyProbingPathsSecrets = $false
     $getTokenForPush = $false
     [System.Collections.ArrayList]$secretsCollection = @()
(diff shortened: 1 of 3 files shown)


Actions/ReadSecrets/ReadSecrets.ps1
```diff
@@ -26,7 +26,7 @@ try {
 
     $outSecrets = [ordered]@{}
     $settings = $env:Settings | ConvertFrom-Json | ConvertTo-HashTable
-    $keyVaultCredentials = GetKeyVaultCredentials
+    $keyVaultCredentials = GetKeyVaultCredentials -testSecretName 'algotestsecret'
```
mazhelez, line 29: Consider using a `-testConnection` (or similar) switch instead.

Actions/ReadSecrets/ReadSecretsHelper.psm1
```diff
@@ -84,6 +84,10 @@ function GetGithubSecret {
 }
 
 function GetKeyVaultCredentials {
+    Param(
+        [string] $testSecretName = $null
```
mazhelez, line 88: It looks like we always want to test if the provided credentials are enough to read secrets. So, a parameter isn't needed.

aholstrup1 (approved):

PR #1227, opened 2024-09-28
Add JSON schema for AL-Go settings

mazhelez:
Add JSON schema for AL-Go settings to improve the experience when modifying the default settings.
E.g.
image

- Add `$schema` property to the settings files in the AL-Go templates
- Modify the `$schema` property in settings files when updating AL-Go

Open questions: should "Update AL-Go System Files" also add the `$schema` property if it's missing in the settings files?


Actions/AL-Go-Helper.ps1
```diff
@@ -814,21 +822,37 @@ function ReadSettings {
...
+    if(Get-Command Test-Json -ErrorAction SilentlyContinue) { # Test-Json is only available in PowerShell 6.2 and later
+        if (!$silent.IsPresent) { Write-Host "Testing settings against the settings schema" }
+        $settingsJson = ConvertTo-Json -InputObject $settings -Depth 99
+        $settingsSchema = Get-Content -Path "$PSScriptRoot\settings.schema.json" -Raw
+
+        try{
+            Test-Json -json $settingsJson -schema $settingsSchema | Out-Null
+        }
+        catch {
+            throw "Settings are not valid. Error: $($_.Exception.Message)"
```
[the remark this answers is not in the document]
freddydk: I definitely don't think we should break anything in the first version.
We should write a warning and add telemetry to see if people are running into this.
Else - this could potentially block people from upgrading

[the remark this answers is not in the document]
mazhelez: Agreed to emit a warning instead.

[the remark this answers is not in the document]
mazhelez: What's your concern with output the actual error message?

As seen in the tests, the error doesn't contain the setting's value:
E.g.:
```
The JSON is not valid with the schema: The string value is not a match for the indicated regular expression at '/shell'
```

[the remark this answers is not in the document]
mazhelez: As for logging a warning, unfortunately, the telemetry functions use `ReadSettings`, so it isn't possible to emit telemetry from it.

Actions/AL-Go-Helper.ps1
```diff
@@ -532,63 +532,12 @@ function MergeCustomObjectIntoOrderedDictionary {
...
-    $workflowName = $workflowName.Trim().Split([System.IO.Path]::getInvalidFileNameChars()) -join ""
-
-    # Start with default settings
-    $settings = [ordered]@{
+function GetDefaultSettings
+(
+    [string] $repoName
+)
+{
+    return [ordered]@{
```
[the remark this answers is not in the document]
mazhelez: Another time. Not needed for this PR. Besides, as it is currently, there is an option to add comments easily.

Actions/settings.schema.json
```diff
@@ -0,0 +1,592 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "description": "Schema for AL-Go settings",
+    "type": "object",
```
freddydk, line 4: Should we consider having a repository schema file, a project schema and a workflow schema?
I mean a setting like "workflowSchedule" is only allowed in workflow specific settings files f.ex.
and other settings are supposed to be in repo settings only.

mazhelez: Thanks for bringing it up. I did think of that, but decided to keep it loose and don't check specific settings files and settings layers, but the settings object after all _layers_ are applied (org settings, repo settings, project settings etc.).

If we don't guard the specific settings level, there's much more flexibility on how to the settings can be organized. In a single-project setup, for example. all settings can be in AL-Go-settings.json (I think empty .AL-Go/settings.json might still be needed though).

About `workflowSchedule`, why not allowing the _default_ value to be set in a single place (e.g. the org settings) and only changed/removed on specific cases.

There are a lot of different setups, and my thinking is that allowing flexibility with AL-Go settings will pay off.

Actions/AL-Go-Helper.ps1
```diff
@@ -814,21 +822,37 @@ function ReadSettings {
...
-        Write-Host "Switching shell to pwsh for ubuntu"
+        if (!$silent.IsPresent) { Write-Host "Switching shell to pwsh for ubuntu" }
         $settings.githubRunnerShell = "pwsh"
     }
 
     if($settings.projectName -eq '') {
         $settings.projectName = $project # Default to project path as project name
     }
+
+    if(Get-Command Test-Json -ErrorAction SilentlyContinue) { # Test-Json is only available in PowerShell 6.2 and later
```
freddydk, line 843: Shouldn't we support running on PowerShell 5 as well, by spawning a pwsh session to do the test-json?

Actions/CheckForUpdates/CheckForUpdates.ps1
```diff
@@ -170,31 +175,53 @@ foreach($checkfile in $checkfiles) {
...
+                    continue
                 }
-                elseif ($dstFileExists) {
-                    # file exists, compare and add to $updateFiles if different
-                    $dstContent = Get-ContentLF -Path $dstFile
-                    if ($dstContent -cne $srcContent) {
-                        Write-Host "Updated $type ($(Join-Path $dstPath $filename)) available"
-                        $updateFiles += @{ "DstFile" = Join-Path $dstPath $filename; "content" = $srcContent }
+
+                switch ($type) {
```
freddydk, line 192: In the old code - all the modifications are done prior to this and the old structure here was:
if (it isn't needed) {
    // remove it if it exists
}
elseif (it already exists) {
    // mark it as updated if it was modified
}
else {
    // create it
}

Now part of the modifying files are moved into this code, which IMO complicates things quite a bit and makes it harder to read - also, you will always add the settings file to updateFiles (even if they haven't changed) and thus the Check for updates will always say that updates are available.

I think you should instead have a switch in line 165 on the type and call a function to get the updated settings file with the the right formatting and then leave this area for comparing and marking files for removal, updating or creation.

freddydk (changes requested):

Tests/AL-Go-Helper.Test.ps1
```diff
@@ -271,114 +391,31 @@
...
-        Pop-Location
-        Remove-Item -Path $tempName -Recurse -Force
+    It 'Settings schema is valid' {
+        Test-Json -json $schema | Should -Be $true
     }
 
-    It 'GetFoldersFromAllProjects' {
-        Mock Write-Host { }
-        Mock Out-Host { }
+    It 'Default settings match schema' {
```
aholstrup1, line 398: Can we also check that all default settings are present within the syntax file? Just so we don't forget to add new settings to the syntax file

aholstrup1 (approved):

Actions/CheckForUpdates/CheckForUpdates.ps1
```diff
@@ -162,20 +162,26 @@ foreach($checkfile in $checkfiles) {
...
+                        # For non-workflow files, just read the file content
+                        $srcContent = Get-ContentLF -Path $srcFile
+                    }
+                }
                 # Replace static placeholders
                 $srcContent = $srcContent.Replace('{TEMPLATEURL}', $templateUrl)
 
                 if ($isDirectALGo) {
-                    # If we are using direct AL-Go repo, we need to change the owner to the templateOwner, the repo names to AL-Go and AL-Go/Actions and the branch to templateBranch
+                    # If we are using direct AL-Go repo, we need to change the owner to the remplateOwner, the repo names to AL-Go and AL-Go/Actions and the branch to templateBranch
```
aholstrup1, line 184:
```suggestion
                    # If we are using direct AL-Go repo, we need to change the owner to the templateOwner, the repo names to AL-Go and AL-Go/Actions and the branch to templateBranch
```

aholstrup1 (dismissed):

Actions/AL-Go-Helper.ps1
mazhelez: Added support for validating settings in PS5.

However, I'm a bit concerned about the extra overhead of around 0.5 second. Here are some stats of `ReadSettings` (in seconds)

| | powershell | pwsh |
| :-- | :-- | :-- |
| Without Validation | 0.021232139 | 0.01290875 |
| With Validation | 0.533863531 | 0.016743805 |

Actions/CheckForUpdates/CheckForUpdates.ps1
mazhelez: Cleaned up and optimized a little bit.

Actions/AL-Go-Helper.ps1
freddydk: got it - maybe we should switch the default to pwsh instead of powershell at some point in time?

Actions/Github-Helper.psm1
```diff
@@ -815,25 +815,49 @@ function Set-ContentLF {
...
+
+    Process {
+        $content = $object | ConvertTo-Json -Depth 99
         if ($PSVersionTable.PSVersion.Major -lt 6) {
             try {
-                $path = $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath($path)
-                # This command will reformat a JSON file with LF line endings as PowerShell 7 would do it (when run using pwsh)
-                $command = "`$cr=[char]13;`$lf=[char]10;`$path='$path';`$content=Get-Content `$path -Encoding UTF8|ConvertFrom-Json|ConvertTo-Json -Depth 99;`$content=`$content -replace `$cr,'';`$content|Out-Host;[System.IO.File]::WriteAllText(`$path,`$content+`$lf)"
-                . pwsh -command $command
+                # This command will reformat a JSON content with LF line endings as PowerShell 7 would do it (when run using pwsh)
```
freddydk, line 850: This code returns the string with CR+LF - we need to remove the CR - else we will detect changes in settings file all the time.

freddydk (dismissed):

freddydk (approved):

aholstrup1 (approved):

Actions/AL-Go-Helper.ps1
mazhelez: I agree. Let's have it as is for now, but have it in mind for a future improvement.

PR #1250, opened 2024-10-10
Update Releasing.md

macel94:
validation does not pass if i use +1.0


freddydk (approved):

freddydk:
yeah, that is correct - we do not support +1.x - only +1 - thanks for catching this.

aholstrup1 (approved):

Issue #1252, opened 2024-10-13
[Bug]: Self-hosted-runner on Windows - Signing Tool not found


sschuh-mumdat:
### AL-Go version

6.0

### Describe the issue

Hi,

maybe someone give me a hint in the right direction...

**To the story:**
i'm about to setup and test a self-hosted runner on a Windows server.

Installed components:
		choco install git --force --params "/NoAutoCrlf"
		choco install 7zip
		choco install gh
		choco install pwsh -y
		choco install vcredist140 -y
		choco install dotnet -y
		choco install dotnet-sdk -y	

Feed:
	dotnet nuget add source https://api.nuget.org/v3/index.json --name nuget.org

Windows Defender/Antivirus is disabled.

When it goes to signing, following error occurs:

"The sign.exe can not be found."

```
Run microsoft/AL-Go-Actions/Sign@v6.0
Run C:\actions-runner\_work\_actions\microsoft\AL-Go-Actions\v6.0\Sign/../Invoke-AlGoAction.ps1 -ActionName "Sign" -Action {
Files to be signed
Module Az.Accounts is available in version 3.0.4
Importing module Az.Accounts (minimum version 2.15.1)
Module Az.KeyVault is available in version 6.2.0
Importing module Az.KeyVault (minimum version 5.2.0)
Connecting to Azure using clientId and clientSecret.
Successfully connected to Azure

Signing files
  Installing signing tool version 0.9.1-beta.24123.2 in C:\Users\github-service\AppData\Local\Temp\SigningTool-78[46](https://github.com/mumdat/bezahlmodul/actions/runs/11312907797/job/31460997018#step:9:49)30865
  No NuGet sources are defined or enabled
  Applying settings from C:\actions-runner\_work\bezahlmodul\bezahlmodul\.github\AL-Go-Settings.json
  Applying settings from C:\actions-runner\_work\bezahlmodul\bezahlmodul\.AL-Go\settings.json
  No settings found in C:\actions-runner\_work\bezahlmodul\bezahlmodul\.github\CICD.settings.json
  No settings found in C:\actions-runner\_work\bezahlmodul\bezahlmodul\.AL-Go\CICD.settings.json
  No settings found in C:\actions-runner\_work\bezahlmodul\bezahlmodul\.AL-Go\sschuh-mumdat.settings.json
  Enabling Microsoft telemetry...

Error: 
Unexpected error when running action. 
Error Message: 
[...]

PR #1270, opened 2024-10-22
Write output of sign install to log

aholstrup1:
* Update the Sign script to log the output of dotnet install to the logs. 
* Update the sign package to latest. The Sign team recently made an improvement to write a warning if _Visual C++ Runtime_ is not installed. If it isn't, signing will fail because NavSip depends on this assembly.  

Related to #1252

### DIFF
--- a/Actions/Sign/Sign.psm1
+++ b/Actions/Sign/Sign.psm1
@@ -16,10 +16,14 @@ function Install-SigningTool() {
     # Install the signing tool in the temp folder
     Write-Host "Installing signing tool version $version in $tempFolder"
     New-Item -ItemType Directory -Path $tempFolder | Out-Null
-    dotnet tool install sign --version $version --tool-path $tempFolder | Out-Null
+    $dotnetInstallOutput = dotnet tool install sign --version $version --tool-path $tempFolder
+    Write-Host $dotnetInstallOutput
 
     # Return the path to the signing tool
-    $signingTool = Join-Path -Path $tempFolder "sign.exe" -Resolve
+    $signingTool = Join-Path -Path $tempFolder "sign.exe"
+    if (-not (Test-Path -Path $signingTool)) {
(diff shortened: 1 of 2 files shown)


mazhelez (dismissed):

Actions/Sign/Sign.psm1
```diff
@@ -16,10 +16,14 @@ function Install-SigningTool() {
     # Install the signing tool in the temp folder
     Write-Host "Installing signing tool version $version in $tempFolder"
     New-Item -ItemType Directory -Path $tempFolder | Out-Null
-    dotnet tool install sign --version $version --tool-path $tempFolder | Out-Null
+    $dotnetInstallOutput = dotnet tool install sign --version $version --tool-path $tempFolder
```
freddydk, line 19: Wouldn't | Out-Host do the same?

Actions/Sign/Sign.psm1
```diff
@@ -16,10 +16,14 @@ function Install-SigningTool() {
...
+    Write-Host $dotnetInstallOutput
 
     # Return the path to the signing tool
-    $signingTool = Join-Path -Path $tempFolder "sign.exe" -Resolve
+    $signingTool = Join-Path -Path $tempFolder "sign.exe"
+    if (-not (Test-Path -Path $signingTool)) {
+        throw "Failed to install signing tool. If you are using a self-hosted runner, make sure the runner has .NET installed and nuget.org set up as a nuget source."
```
freddydk, line 25: Shouldn't we state version and maybe how nuget source could be added?

freddydk (dismissed):

Actions/Sign/Sign.psm1
aholstrup1: Good point. I'll adjust

Actions/Sign/Sign.psm1
aholstrup1: We do state how it should be done in the docs so maybe we can link them to the docs.

freddydk (approved):

mazhelez (approved):

PR #1287, opened 2024-11-01
Enable Git submodule support

dibyanugraha:
I use git submodule instead of putting the extension source codes directly in the repo, to my surprise it doesn't automatically initialize the submodule thus failing the automatic build.

These changes should make the workflow is able to initialize any git submodules inside the repo thus making the automatic build passes.

Tested in this repo https://github.com/dibyanugraha/d365-bc-toastui-calendar-control-addin-cicd


dibyanugraha:

freddydk:
I actually didn't know about submodules until now.
reading about it - wouldn't we have do add this to more places we do checkout?
I think you would need it during Initialize in CICD, Current, NextMinor and NextMajor as well, where we analyze the repo to determine projects and depth.

dibyanugraha:
You might be right, I'll update the rest of the yaml files. At first I tried to add it in the cicd.yaml but it still didn't work, then I found it worked by adding in _BuildALGoProject.yaml instead.

Yes I'm aware it's a public repo, nothing to worry about it, I don't plan on deploy it anywhere, it's merely my playground to test.

freddydk:
> You might be right, I'll update the rest of the yaml files. At first I tried to add it in the cicd.yaml but it still didn't work, then I found it worked by adding in _BuildALGoProject.yaml instead.
> 
> Yes I'm aware it's a public repo, nothing to worry about it, I don't plan on deploy it anywhere, it's merely my playground to test.

You also need to update the same yaml files in the AppSource App/.github/workflows folder.

dibyanugraha:
> > You might be right, I'll update the rest of the yaml files. At first I tried to add it in the cicd.yaml but it still didn't work, then I found it worked by adding in _BuildALGoProject.yaml instead.
> > Yes I'm aware it's a public repo, nothing to worry about it, I don't plan on deploy it anywhere, it's merely my playground to test.
> 
> You also need to update the same yaml files in the AppSource App/.github/workflows folder.

Done, sorry I missed those

freddydk:
One more thing, sorry for forgetting.

Could you add a section to releasenotes.md (under support for page scripting) following the same style, explaining what this is and how to use it (linking to github docs probably - not private repos).

Thanks.

dibyanugraha:
> One more thing, sorry for forgetting.
> 
> Could you add a section to releasenotes.md (under support for page scripting) following the same style, explaining what this is and how to use it (linking to github docs probably - not private repos).
> 
> Thanks.

Sure thing. I've added it though I'm not sure whether it's good enough to explain git submodule. You're welcome to make it fits.

RELEASENOTES.md
```diff
@@ -25,6 +25,28 @@ Page Scripting tests are now supported as part of CI/CD. By specifying pageScrip
 - `PageScriptingTestResults` is a JUnit test results file with all results combined.
 - `PageScriptingTestResultDetails` are the detailed test results (including videos) when any of the page scripting tests have failures. If the page scripting tests succeed - the details are not published.
 
+### Support for Git submodule
+
+[Git submodule](https://git-scm.com/book/en/v2/Git-Tools-Submodules) is now supported as part of CI/CD on your project.
+
+If you're already established your project repos outside the AL Go template, you may use submodule instead of placing and commiting your code directly to the repo generated by this template.
```
mazhelez, line 32:
```suggestion
If you already have created your project repos outside the AL-Go template, you may use a git submodule instead of moving your code directly to the repo generated using this template.
```

RELEASENOTES.md
```diff
@@ -25,6 +25,28 @@ Page Scripting tests are now supported as part of CI/CD. By specifying pageScrip
...
+
+To use the Git submodule just simply remove entirely your project folder and replace it by adding your repo as submodule by running git command:
```
mazhelez, line 34: This part should be a separate scenario under _Scenarios_ and only referenced here.

RELEASENOTES.md
dibyanugraha: This suggestion is no longer applicable since the part is deleted as the outcome of below suggestion 
https://github.com/microsoft/AL-Go/pull/1287#discussion_r1825718824

aholstrup1 (approved):

mazhelez (approved):

freddydk:
We have decided to take this as experimental support and merge it.
It will be included in the next preview deployment.

freddydk:
@dibyanugraha - we have received an issue #1344 - which is another partner using git submodules for other things.

Are you only using this for public repositories?
It seems like the current implementation will only work for public submodules - or?

PR #1333, opened 2024-11-21
Make appFolders optional

### LINKED ISSUE #1296: [Bug]: Make property "appFolders" optional
### Feature description

The property `appFolders` in the file `.AL-Go/settings.json` shouldn't be required. The documentation reads as follows:
> appFolders should be an array of folders (relative to project root), which contains apps for this project. Apps in these folders are sorted based on dependencies and built and published in that order.
If appFolders are not specified, AL-Go for GitHub will try to locate appFolders in the root of the project.
The doc explicitly states that not specifying `appFolders` is viable, but that only applies if it is defined as an empty array. Nothing about this indicates that this is a required property, it even lists the default value `[]`.

If the definition of the property is absolutely necessary, both the schema and documentation should be updated to reflect this; otherwise omitting it shouldn't cause an error.

Edit: To be specific, AL Go checks the presence of the `appFolders` property during the action "Create a new app".

freddydk:
Fixes #1296

### DIFF
--- a/Actions/CreateApp/CreateApp.ps1
+++ b/Actions/CreateApp/CreateApp.ps1
@@ -87,7 +87,7 @@ try {
                 $settingsJson.$_ = @()
             }
         }
-        if (@($settingsJson.appFolders)+@($settingsJson.testFolders+@($settingsJson.bcptTestFolders)) {
+        if (@($settingsJson.appFolders)+@($settingsJson.testFolders)+@($settingsJson.bcptTestFolders)) {
             if ($type -eq "Performance Test App") {
                 if ($SettingsJson.bcptTestFolders -notcontains $foldername) {
                     $SettingsJson.bcptTestFolders += @($folderName)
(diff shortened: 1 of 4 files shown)


Actions/CreateApp/CreateApp.ps1
```diff
@@ -82,7 +82,12 @@ try {
     try {
         $settingsJsonFile = Join-Path $projectFolder $ALGoSettingsFile
         $SettingsJson = Get-Content $settingsJsonFile -Encoding UTF8 | ConvertFrom-Json
-        if (@($settingsJson.appFolders)+@($settingsJson.testFolders)) {
+        'appFolders','testFolders','bcptTestFolders' | ForEach-Object {
+            if ($settingsJson.PSObject.Properties.Name -ne $_) {
```
mazhelez, line 86: Don't you mean `-notcontains` here? `-ne` will find all props that are not equal to the current one, which might always be the case.

```suggestion
            if ($settingsJson.PSObject.Properties.Name -notcontains $_) {
```

freddydk: You are right - I ment (-not (name -eq $_))

mazhelez (approved):

jonasblunck (approved):

PR #1346, opened 2024-12-03
Submodules support

freddydk:


Actions/AL-Go-Helper.ps1
```diff
@@ -666,6 +666,8 @@ function ReadSettings {
             "Account"                                   = ""
             "CertificateProfile"                        = ""
         }
+        "useSubModules"                                 = ""
+        "subModulesTokenSecretName"                     = "subModulesToken"
```
aholstrup1, line 670: Can't we just hardcode the secret name instead of it being a setting? Like the Azure_Credentials secret for example. That would save us from calling ReadSecrets before actions/checkout since we know what the secret name is

Actions/AL-Go-Helper.ps1
```diff
@@ -666,6 +666,8 @@ function ReadSettings {
             "Account"                                   = ""
             "CertificateProfile"                        = ""
         }
+        "useSubModules"                                 = ""
```
aholstrup1, line 669:
```suggestion
        "useSubModules"                                 = $false
```

freddydk: useSubModules can be `true` to checkout submodules or `recursive` to recursively checkout submodules - a boolean would not suffice

Actions/AL-Go-Helper.ps1
freddydk: All secrets (incl. azure_credentials) can be mapped to other secrets using the SecretName setting construct.
The default is subModulesToken, so people don't have to do anything and we could also remove this setting from AL-Go-Helper and it would still work. This setting is just to symbolize that it works like this

freddydk: and we also need to support getting this from an Azure KeyVault etc etc.

Actions/AL-Go-Helper.ps1
aholstrup1: Yeah, so the options should be:
* True
* False (Default)
* Recursive

freddydk: Why would "false" be better than "" as undefined?
"false" is an arbitrary value, which cannot be given to actions/checkout - why is it better than "" - it isn't a value people will ever see.

freddydk: Defaulting to "false" would mean the code would look like:

```
      - name: Checkout Submodules
        if: env.useSubModules != 'false' && env.useSubModules != ''
        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
        with:
```

as most other string settings in AL-Go uses an empty string as the default (I think) - but if you feel strongly about this - then fine.

freddydk: > Why would "false" be better than "" as undefined? "false" is an arbitrary value, which cannot be given to actions/checkout - why is it better than "" - it isn't a value people will ever see.

I stand corrected - 'false' can be given to actions/checkout - but we would never do that. I can however see that if anybody reads the documentation for actions/checkout - then they might assume that 'false' is a good value.

freddydk: fixed

Actions/AL-Go-Helper.ps1
```diff
@@ -666,6 +666,8 @@ function ReadSettings {
             "Account"                                   = ""
             "CertificateProfile"                        = ""
         }
+        "useSubModules"                                 = "false"
```
mazhelez, line 669:
```suggestion
        "useSubmodules"                                 = "false"
```

Or maybe even `useGitSubmodules`?

Actions/AL-Go-Helper.ps1
```diff
@@ -666,6 +666,8 @@ function ReadSettings {
...
+        "subModulesTokenSecretName"                     = "subModulesToken"
```
mazhelez, line 670:
```suggestion
        "submodulesTokenSecretName"                     = "subModulesToken"
```

Templates/AppSource App/.github/workflows/CICD.yaml
```diff
@@ -66,7 +65,25 @@ jobs:
...
+          get: type,powerPlatformSolutionFolder,useSubModules
+
+      - name: Read subModules token
+        id: ReadSubModulesToken
+        if: env.useSubModules != 'false' && env.useSubModules != ''
+        uses: microsoft/AL-Go-Actions/ReadSecrets@main
+        with:
+          shell: ${{ inputs.shell }}
+          gitHubSecrets: ${{ toJson(secrets) }}
+          getSecrets: '-subModulesToken'
```
mazhelez, line 77: What's the dash in the beginning for?

Templates/AppSource App/.github/workflows/CICD.yaml
```diff
@@ -66,7 +65,25 @@ jobs:
...
           shell: powershell
-          get: type, powerPlatformSolutionFolder
+          get: type,powerPlatformSolutionFolder,useSubModules
+
+      - name: Read subModules token
+        id: ReadSubModulesToken
+        if: env.useSubModules != 'false' && env.useSubModules != ''
+        uses: microsoft/AL-Go-Actions/ReadSecrets@main
+        with:
+          shell: ${{ inputs.shell }}
```
mazhelez, line 75: There are no inputs at this point.

Templates/AppSource App/.github/workflows/CICD.yaml
```diff
@@ -66,7 +65,25 @@ jobs:
...
+          gitHubSecrets: ${{ toJson(secrets) }}
+          getSecrets: '-subModulesToken'
+
+      - name: Checkout Submodules
+        if: env.useSubModules != 'false' && env.useSubModules != ''
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          ref: ${{ inputs.checkoutRef }}
```
mazhelez, line 83: Same here. Where does `checkoutRef` come from?

mazhelez (commented):
PR is not completely ready yet.

Minor: _subModules_ should be _submodules_ all around.

How about having a custom AL-Go-specific composite checkout action that handles the git submodules logic? 
- actions/checkout without submodules to be able to read the settings
- depending on the new `useSubmodules` setting and secret, do the appropriate checkout.

 This way the workflows won't be so polluted, and the logic will be centralized.

Templates/AppSource App/.github/workflows/CICD.yaml
```diff
@@ -66,7 +65,25 @@ jobs:
...
+          lfs: true
+          submodules: ${{ env.useSubModules }}
+          token: '${{ fromJson(steps.ReadSubModulesToken.outputs.Secrets).subModulesToken }}'
```
mazhelez, line 86: How do you handle the case where the git submodule is a public repo and no extra token is needed?

aholstrup1: It falls back to using github_token as far as I can see.

Templates/AppSource App/.github/workflows/CICD.yaml
freddydk: copy/paste error

Templates/AppSource App/.github/workflows/CICD.yaml
freddydk: For asking readsecrets to not base64 encode the secret

Templates/AppSource App/.github/workflows/CICD.yaml
freddydk: copy/paste error

freddydk:
> Minor: _subModules_ should be _submodules_ all around.

Have renamed and added git in front of everything

> How about having a custom AL-Go-specific composite checkout action that handles the git submodules logic?
> 
> * actions/checkout without submodules to be able to read the settings
> * depending on the new `useSubmodules` setting and secret, do the appropriate checkout.
> 
> This way the workflows won't be so polluted, and the logic will be centralized.

We could do that - not sure how much we gain from that - would rather see whether this solves the issue before adding refactoring task to this.

aholstrup1 (approved):

mazhelez (approved):

freddydk:
> As I mentioned, if it's a single action, the workflows won't be so polluted (just the current actions/checkout will be replaced with AL-Go-Actions/checkout that will take care of reading the needed settings, secrets and checking out the code). Currently, the following is duplicated across build workflows and it doesn't have to be:

And yes, I acknowledge that - but this is really a bug fix to make sure that people upgrading won't have to downgrade.
I will add an item to the backlog to fix this - it isn't faster to fix it now than fixing it later - and this would include quite some new risks,

mazhelez (approved):

aholstrup1 (approved):

PR #1037, opened 2024-04-15
Run "Update AL-Go System Files" on multiple branches

mazhelez:
"Update AL-Go System Files" workflow could only be run on one branch at a time. Moreover, the schedule mechanism works only on the default branch (GitHub Actions limitation).

This PR alters the workflow to run the update logic on multiple branches, defined in a new AL-Go setting `updateALGoBranches`.

PR #1351, opened 2024-12-06
Add support for including a suffix in the Commit message/PR message

aholstrup1:
Add support for including a suffix in the Commit message/PR message

This is useful if you have e.g. the Azure Boards integration set up in your repository. Then you can add "AB#<WorkItemId>" to your commit/PR messages. Currently in BCApps we require a workitem so we have to manually edit the PRs for "Update AL-Go System Files" so it includes the workitem. See for example: https://github.com/microsoft/BCApps/pull/2455

Also seems applicable to JIRA integrations: https://github.com/microsoft/AL-Go/discussions/1140


Actions/AL-Go-Helper.ps1
```diff
@@ -1355,9 +1356,23 @@ function CommitFromNewFolder {
     invoke-git add *
     $status = invoke-git -returnValue status --porcelain=v1
     if ($status) {
+        $title = $commitMessage
+
+        # Add commit message suffix if specified in settings
+        $settings = ReadSettings
+        if ($settings.commitMessageSuffix) {
```
freddydk, line 1363: I would have expected this would be an input field, rather than a setting???

aholstrup1: Right... I think both could make sense. 

If you run "Update AL-GO System Files" on a schedule I'm not sure the input would make sense. In that scenario I think a setting like this is better. But you're right... In the non-scheduled case an input would be a better option since it would give you more flexibility. 

Adding support for using the input field would be a bit of a heavier lift though (I think). IMO it could make sense to start with a simpler approach and expand it later if needed. What do you think? 

Motivation for the PR is really that I'm tired of having to manually edit the 10+ PRs we get whenever there are AL-Go updates to BCApps.

mazhelez: And I'm tired of adding the "Automation" label + enabling the auto-merge on every PR from AL-Go to BCApps.
How about a more structured setting, as in https://github.com/microsoft/AL-Go/pull/1037/files#diff-d6e7039af6c0c11ad1a23decab42e0c9d1faeb7756089f0ebf640a2694665014

Pros of the settings: can be used in a scheduled workflows
Pros of the workflow input: it's more transparent what the workflow will do

aholstrup1: That too.. 
Don't know about this particular structured setting though. Presumably you'd need one for every workflow that creates PRs (since it's specific to Update AL-Go System Files)? Couldn't we leverage the "settings hierarchy"? If you want a setting to only apply to "Update AL-Go System Files" you need to create the `Update AL-Go System Files.settings.json`

Then we could have a single setting like: 
```
"workflowCommitSettings" = [ordered] @{
    "commitMessageSuffix": "Fixes AB#Something"
    "PRLabels": @("Automation")
    "AutocompletePR": $true
    "Whatever else": ...
 }
```

Actions/AL-Go-Helper.ps1
```diff
@@ -661,6 +661,12 @@ function ReadSettings {
             "defaultReleaseMD"                          = "## Release reference documentation\n\nThis is the generated reference documentation for [{REPOSITORY}](https://github.com/{REPOSITORY}).\n\nYou can use the navigation bar at the top and the table of contents to the left to navigate your documentation.\n\nYou can change this content by creating/editing the **{INDEXTEMPLATERELATIVEPATH}** file in your repository or use the alDoc:defaultReleaseMD setting in your repository settings file (.github/AL-Go-Settings.json)\n\n{RELEASENOTES}"
         }
         "trustMicrosoftNuGetFeeds"                      = $true
+        "commitMessageSuffix"                           = ""
```
freddydk, line 664: double commitMessageSuffix?

Scenarios/settings.md
```diff
@@ -79,6 +79,7 @@ The repository settings are only read from the repository settings file (.github
...
 | <a id="useGitSubmodules"></a>useGitSubmodules | If your repository is using Git Submodules, you can set the `useGitSubmodules` setting to `"true"` or `"recursive"` in order to use these submodules during build workflows. If `useGitSubmodules` is not set, git submodules are not initialized. If the submodules reside in private repositories, you need to define a `gitSubmodulesToken` secret. Read [this](https://aka.ms/algosecrets#gitSubmodulesToken) for more information. |
+| <a id="commitMessageSuffix"></a>commitMessageSuffix | If you want AL-Go to include a particular message in commits and pull requests opened by AL-Go you can specify a `commitMessageSuffix`.  |
```
freddydk, line 82: Use commitOptions.
It should probably be described that this setting can be defined in workflow specific settings file in .github/workflowname.settings.json.

RELEASENOTES.md
```diff
@@ -6,6 +6,7 @@
 ### New Repository Settings
 
 - `useGitSubmodules` can be either `true` or `recursive` if you want to enable Git Submodules in your repository. If your Git submodules resides in a private repository, you need to create a secret called `gitSubmodulesToken` containing a PAT with access to the submodule repositories. Like with all other secrets, you can also create a setting called `gitSubmodulesTokenSecretName` and specify the name of another secret, with these permissions (f.ex. ghTokenWorkflow).
+- `commitMessageSuffix` - A string you want to append to the end of commits/pull requests created by AL-Go. This can be useful if you are using the Azure Boards integration (or similar integration) to link commits to workitems.
```
freddydk, line 9: commitOptions

Actions/AL-Go-Helper.ps1
```diff
@@ -661,6 +661,12 @@ function ReadSettings {
...
+        "commitOptions"                                 = [ordered]@{
+            "commitMessageSuffix"                       = ""
+            "commitAutoMerge"                           = $false
+            "commitPullRequestLabels"                   = @()
+        }
```
freddydk, line 669: Maybe add "alwaysUseGhTokenWorkflow" and "alwaysUseDirectCommit" in commitOptions - where false will use the input values and true will always use...

Actions/AL-Go-Helper.ps1
```diff
@@ -661,6 +661,12 @@ function ReadSettings {
             "defaultReleaseMD"                          = "## Release reference documentation\n\nThis is the generated reference documentation for [{REPOSITORY}](https://github.com/{REPOSITORY}).\n\nYou can use the navigation bar at the top and the table of contents to the left to navigate your documentation.\n\nYou can change this content by creating/editing the **{INDEXTEMPLATERELATIVEPATH}** file in your repository or use the alDoc:defaultReleaseMD setting in your repository settings file (.github/AL-Go-Settings.json)\n\n{RELEASENOTES}"
         }
         "trustMicrosoftNuGetFeeds"                      = $true
+        "commitMessageSuffix"                           = ""
+        "commitOptions"                                 = [ordered]@{
+            "commitMessageSuffix"                       = ""
```
freddydk, line 666: Remove "commit" from the 3 properies

Actions/AL-Go-Helper.ps1
aholstrup1: We can add those options as well, but wouldn't you expect the input to override whatever settings we have defined? That's how the "Template Repository URL" input behaves isn't it? 

image

Actions/AL-Go-Helper.ps1
aholstrup1: Updated this to a structured setting with more options. Let me know what you think.

Actions/AL-Go-Helper.ps1
freddydk: I wouldn't "just" add these options - I would add an option called "alwaysUseDirectCommit"
If true - it would always use direct commit (also if the checkbox was left to false)
If false (or non existing) - it would use the value of the checkbox.

I wouldn't add the repository URL or the download latest to commitOptions - but I would add one called alwaysUseGhTokenWorkflow (which also is a checkbox and using the same rules as above).

aholstrup1: > If true - it would always use direct commit (also if the checkbox was left to false)

My point was... The above isn't consistent with how the "Template Repository URL" input works. If you define that input, we will override the templateUrl specified in settings. I think that's fine, but we should keep that consistent across AL-Go. IMO it should behave as if inputs are an 8th entry in this: https://github.com/microsoft/AL-Go/blob/main/Scenarios/settings.md#where-are-the-settings-located. 

I think if a user specifies one thing in the checkbox but something else happens that would be quite confusing. 

Thoughts?

freddydk: I kind of agree, but the problem with checkboxes is that they are not 3 state...
What does an empty checkbox mean? that it is defined as empty or that it is not set?
Which is why I would (for checkboxes) use the always- paradigm.
If the messagesuffix was an input field - the setting would be the default.

aholstrup1: If we want "Direct Commit" to have 3 states, it can't also be a checkbox since they only have two states. Checkboxes require the user to make a decision about whether they want the input to be true or false. You can't "leave it empty" - You're forced to make a decision. 

Maybe directCommit and useGhTokenWorkflow need to be "choice fields" that default to empty? 
image

> Which is why I would (for checkboxes) use the always- paradigm.

Always- paradigm being? If a setting has the "always" prefix it should take precedence even if a setting further down the chain is set to a different value?

freddydk: Yes - I am not talking about a generic option for all checkboxes - I am only talking about these two options:

DirectCommit is default false - if people allow direct commit (in github settings - i.e. no branch policies etc) and want to use that - it is something that would be fine to have globally - i.e. alwaysUseDirectCommit - then these workflows would always use direct commits and they wouldn't forget to check the checkbox.

Use GhTokenWorkflow is the same - if you have this secret defined and you want to use that - it would be fine to have a global setting to always use that. When people are not using GhTokenWorkflow - then the Create New App and other workflows will not spawn a CI/CD workflow - people who have setup the secrets and wants to use this don't want to forget to check the checkbox.

Actions/AL-Go-Helper.ps1
```diff
@@ -661,6 +661,11 @@ function ReadSettings {
             "defaultReleaseMD"                          = "## Release reference documentation\n\nThis is the generated reference documentation for [{REPOSITORY}](https://github.com/{REPOSITORY}).\n\nYou can use the navigation bar at the top and the table of contents to the left to navigate your documentation.\n\nYou can change this content by creating/editing the **{INDEXTEMPLATERELATIVEPATH}** file in your repository or use the alDoc:defaultReleaseMD setting in your repository settings file (.github/AL-Go-Settings.json)\n\n{RELEASENOTES}"
         }
         "trustMicrosoftNuGetFeeds"                      = $true
+        "commitOptions"                                 = [ordered]@{
+            "messageSuffix"                             = ""
+            "autoMerge"                                 = $false
```
mazhelez, line 666: `PRAutoMerge` or similar? To match the other PR-related setting.

mazhelez (dismissed):

Actions/AL-Go-Helper.ps1
aholstrup1: > then these workflows would always use direct commits and they wouldn't forget to check the checkbox.

I still think this has the potential to be quite confusing (if you specify one thing in the checkbox but something else happens). Of course, we could say it's their own fault for using the alwaysUseDirectCommit setting but the user who triggers the workflow might not know that this setting is there and what it does. Or, they could've just forgotten or misunderstood how it works.

If it's okay with you, I'll park this as a future improvement. IMO I think we first need to figure out how we generally handle inputs vs settings in both the scheduled and manually triggered scenario.

freddydk (approved):

PR #1374, opened 2024-12-29
Remove thisbuild artifacts and support preprocessor symbols

freddydk:
This is a proposed fix for https://github.com/orgs/microsoft/projects/521/views/1?pane=issue&itemId=91449371 and will greatly improve the capabilities of build modes (as requested by partners during our roundtable at Directions EMEA).
 
### New Repository Settings

- [`shortLivedArtifactsRetentionDays`](https://aka.ms/algosettings#shortLivedArtifactsRetentionDays) determines the number of days to keep short lived build artifacts (f.ex build artifacts from pull request builds, next minor or next major builds). 1 is default. 0 means use GitHub default.
- [`preProcessorSymbols`](https://aka.ms/algosettings#preProcessorSymbols) is a list of preprocessor symbols to use when building the apps. This setting can be specified in workflow specific settings files or in conditional settings.

### Change in published artifacts

When using `useProjectDependencies` in a  multi-project repository, AL-Go for GitHub used to generate short lived build artifacts called `thisBuild-<projectnaame>-<type>-...`. This is no longer the case. Instead, normal build artifacts will be published and used by depending projects. The retention period for the artifacts generated are controlled by a settings called [`shortLivedArtifactsRetentionDays`](https://aka.ms/algosettings#shortLivedArtifactsRetentionDays).

### Preprocessor symbols

It is now possible to define preprocessor symbols, which will be used when building your apps using the [`preProcessorSymbols`](https://aka.ms/algosettings#preProcessorSymbols) setting. This setting can be specified in workflow specific settings file or it can be used in conditional settings.


RELEASENOTES.md
```diff
@@ -2,10 +2,27 @@
...
 
+### New Repository Settings
+
+- [`shortLivedArtifactsRetentionDays`](https://aka.ms/algosettings#shortLivedArtifactsRetentionDays) determines the number of days to keep short lived build artifacts (f.ex build artifacts from pull request builds, next minor or next major builds). 1 is default. 0 means use GitHub default.
+- [`longLivedArtifactsRetentionDays`](https://aka.ms/algosettings#longLivedArtifactsRetentionDays) determines the number of days to keep long lived build artifacts (f.ex build artifacts from CI/CD builds). 0 is the default and means use GitHub default.
+- [`preProcessorSymbols`](https://aka.ms/algosettings#preProcessorSymbols) is a list of preprocessor symbols to use when building the apps. This setting can be specified in workflow specific settings files or in conditional settings.
+- [`<buildMode>PreProcessorSymbols`](https://aka.ms/algosettings#cleanModePreProcessorSymbols) is a list of preprocessor symbols to be used when building apps in the \<buildMode> build mode. CleanModePreProcessorSymbols is a variation of this.
```
aholstrup1, line 10: Now that this PR also adds buildmode as a condition for conditional settings, shouldn't people just create a conditional setting called "preProcessorSymbols"? Similar to what you did here https://github.com/BusinessCentralApps/buildorder/blob/8e64503f243aaf6e24e7deca8d47b80975c97c54/BO-DK/.AL-Go/settings.json#L8

RELEASENOTES.md
```diff
@@ -2,10 +2,27 @@
 
 - It is now possible to skip the modification of dependency version numbers when running the Increment Version number workflow or the Create Release workflow
 
+### New Repository Settings
+
+- [`shortLivedArtifactsRetentionDays`](https://aka.ms/algosettings#shortLivedArtifactsRetentionDays) determines the number of days to keep short lived build artifacts (f.ex build artifacts from pull request builds, next minor or next major builds). 1 is default. 0 means use GitHub default.
+- [`longLivedArtifactsRetentionDays`](https://aka.ms/algosettings#longLivedArtifactsRetentionDays) determines the number of days to keep long lived build artifacts (f.ex build artifacts from CI/CD builds). 0 is the default and means use GitHub default.
```
aholstrup1, line 8: If we use GitHubs default, why not just let people control this setting through GitHub? Is there a need for an AL-Go setting?
image

freddydk: I guess we could remove the longLivedArtifactsRetentionDays

RELEASENOTES.md
freddydk: I guess we could - but then we would have cleanModePreprocessorSymbols as a special case for compatibility?
Should we then issue a deprecation warning about that and ask people to modify this to a conditionalsetting instead?

aholstrup1: > Should we then issue a deprecation warning about that and ask people to modify this to a conditionalsetting instead?

That sounds like a good plan I think

Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -368,27 +368,29 @@ try {
...
+    }
 
-            if ($runAlPipelineParams.Keys -notcontains 'preprocessorsymbols') {
-                $runAlPipelineParams["preprocessorsymbols"] = @()
-            }
+    # REMOVE AFTER April 1st 2025 --->
+    if ($settings.ContainsKey('cleanModePreprocessorSymbols')) {
+        Write-Host "Adding Preprocessor symbols : $($settings.cleanModePreprocessorSymbols -join ',')"
+        $runAlPipelineParams["preprocessorsymbols"] += $settings.cleanModePreprocessorSymbols
+        Write-Warning -message "cleanModePreprocessorSymbols is deprecated. Use Conditional Settings instead, specifying buildModes and the `preProcessorSymbols` setting."
```
aholstrup1, line 387: Suggestion: 
Maybe we should just link to deprecations.md here? Then we can provide the details on how to migrate to the "new way" in there. E.g.: 
```suggestion
        Write-Warning -message "cleanModePreprocessorSymbols will be deprecated after April 1st 2025. Please see https://github.com/microsoft/AL-Go/blob/main/DEPRECATIONS.md for more information"
```

Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -368,27 +368,29 @@ try {
...
     }
 
-    switch($buildMode){
-        'Clean' {
-            $preprocessorsymbols = $settings.cleanModePreprocessorSymbols
+    if ($buildMode -eq 'Translated') {
+        if ($runAlPipelineParams.Keys -notcontains 'features') {
+            $runAlPipelineParams["features"] = @()
+        }
+        Write-Host "Adding translatiopnfile feature"
```
aholstrup1, line 375:
```suggestion
        Write-Host "Adding translationfile feature"
```

Actions/AL-Go-Helper.ps1
```diff
@@ -686,6 +687,7 @@ function ReadSettings {
         }
         "useGitSubmodules"                              = "false"
         "gitSubmodulesTokenSecretName"                  = "gitSubmodulesToken"
+        "shortLivedArtifactsRetentionDays"              = 1  # 0 means use GitHub default
```
aholstrup1, line 690: I would think the default should be 0. Isn't that effectively what it is today? 

IMO the default setting for this should match whatever it is today so we don't break any partner scenarios.

aholstrup1 (commented):
Couple of comments but looks good overall! 👍

Templates/AppSource App/.github/workflows/CICD.yaml
```diff
@@ -88,6 +89,7 @@ jobs:
         id: DetermineWorkflowDepth
         run: |
           Add-Content -Encoding UTF8 -Path $env:GITHUB_OUTPUT -Value "WorkflowDepth=$($env:workflowDepth)"
+          Add-Content -Encoding UTF8 -Path $env:GITHUB_OUTPUT -Value "ArtifactsRetentionDays=0"
```
mazhelez, line 92: 0 is the default anyway, so why set it explicitly?

Templates/AppSource App/.github/workflows/_BuildALGoProject.yaml
```diff
@@ -46,14 +46,10 @@ on:
...
         type: string
-      publishThisBuildArtifacts:
-        description: Flag indicating whether this build artifacts should be published
-        type: boolean
-        default: false
-      publishArtifacts:
-        description: Flag indicating whether the artifacts should be published
-        type: boolean
-        default: false
+      artifactsRetentionDays:
```
mazhelez, line 49: There's `Read settings` step in this workflow as well. Isn't it better to take the `artifactsRetentionDays` input from here?

Templates/AppSource App/.github/workflows/_BuildALGoProject.yaml
```diff
@@ -175,56 +172,32 @@ jobs:
...
 
       - name: Publish artifacts - test apps
         uses: actions/upload-artifact@6f51ac03b9356f520e9adb1b1b7802705f340c2b # v4.5.0
-        if: inputs.publishArtifacts
+        if: inputs.artifactsRetentionDays >= 0
         with:
           name: ${{ steps.calculateArtifactsNames.outputs.TestAppsArtifactsName }}
           path: '${{ inputs.project }}/.buildartifacts/TestApps/'
           if-no-files-found: ignore
+          retention-days: ${{ inputs.artifactsRetentionDays }}
```
mazhelez, line 200: From [actions/upload-artifact](https://github.com/actions/upload-artifact?tab=readme-ov-file#retention-period):
> The retention period must be between 1 and 90 inclusive

So, 0 isn't really supported.

mazhelez (commented):
Great addition! Added some comments, please address.

Actions/AL-Go-Helper.ps1
freddydk: The shortlived (thisbuild) are 1 today.
The workflow using shortlived artifacts - today they don't produce artifacts - they only produce 1 day lasting thisbuild artifacts, which I wouldn't assume that anybody are fetching themselves.

Templates/AppSource App/.github/workflows/_BuildALGoProject.yaml
freddydk: Further up in the documentaion:
{6F17A4E2-7B86-496D-A022-DBD6C38DEBC0}

Templates/AppSource App/.github/workflows/_BuildALGoProject.yaml
freddydk: It is really because we don't know whether the artifacts will be long lived (CI/CD) or short lived (other workflows) in here.

freddydk: We could deduct that I guess - but this way it is explicit.

Templates/AppSource App/.github/workflows/CICD.yaml
freddydk: Yeah - it can be removed (was here because of the longlived setting)

mazhelez (approved):

aholstrup1 (approved):

PR #1387, opened 2025-01-10
Log telemetry for uses of deprecated features

aholstrup1:
Log telemetry for uses of deprecated features

### DIFF
--- a/Actions/AL-Go-Helper.ps1
+++ b/Actions/AL-Go-Helper.ps1
@@ -196,7 +196,7 @@ function OutputDeprecationWarning {
     $telemetryModule = Join-Path $PSScriptRoot 'TelemetryHelper.psm1'
     if (Test-Path $telemetryModule) {
         Import-Module $telemetryModule
-        Trace-Information -Message "Deprecation warning: $message"
+        Trace-Warning -Message "Deprecation Warning: $message"
     }
 }
 
--- a/Actions/AL-Go-Helper.ps1
+++ b/Actions/AL-Go-Helper.ps1
@@ -185,6 +185,21 @@ function OutputWarning {
     }
 }
 
+function OutputDeprecationWarning {
+    Param(
+        [string] $message
+    )
+    # Show deprecation warning in GitHub
+    OutputWarning -message $message
+
+    # Log deprecation warning to telemetry
+    $telemetryModule = Join-Path $PSScriptRoot 'TelemetryHelper.psm1'
+    if (Test-Path $telemetryModule) {
+        Import-Module $telemetryModule
+        Trace-Information -Message "Deprecation warning: $message"
+    }
+}
+
 function OutputNotice {
     Param(
         [string] $message
--- a/Actions/AL-Go-Helper.ps1
+++ b/Actions/AL-Go-Helper.ps1
@@ -185,21 +185,6 @@ function OutputWarning {
     }
 }
 
-function OutputDeprecationWarning {
-    Param(
-        [string] $message
-    )
-    # Show deprecation warning in GitHub
-    OutputWarning -message $message
-
-    # Log deprecation warning to telemetry
-    $telemetryModule = Join-Path $PSScriptRoot 'TelemetryHelper.psm1'
-    if (Test-Path $telemetryModule) {
-        Import-Module $telemetryModule
-        Trace-Warning -Message "Deprecation Warning: $message"
-    }
-}
-
 function OutputNotice {
     Param(
         [string] $message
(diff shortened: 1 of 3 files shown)


Actions/AL-Go-Helper.ps1
```diff
@@ -185,6 +185,21 @@ function OutputWarning {
...
 
+function OutputDeprecationWarning {
+    Param(
+        [string] $message
+    )
+    # Show deprecation warning in GitHub
+    OutputWarning -message $message
+
+    # Log deprecation warning to telemetry
+    $telemetryModule = Join-Path $PSScriptRoot 'TelemetryHelper.psm1'
```
mazhelez, line 196: The telemetry module loads AL-Go-Helper.ps1.
I suppose it won't be a real problem as it's not a PS module, but I'd advise against this circular dependency.

Actions/AL-Go-Helper.ps1
```diff
@@ -185,6 +185,21 @@ function OutputWarning {
...
+    if (Test-Path $telemetryModule) {
+        Import-Module $telemetryModule
+        Trace-Information -Message "Deprecation warning: $message"
```
mazhelez, line 199: Then maybe we need `Trace-Warning`?

Actions/AL-Go-Helper.ps1
```diff
@@ -185,6 +185,21 @@ function OutputWarning {
     }
 }
 
+function OutputDeprecationWarning {
+    Param(
+        [string] $message
```
freddydk, line 190: NIT - we could have a message and a deprecationTag - and then if the deprecationTag is set we will add "See https://aka.ms/ALGoDeprecations#<deprecationTag> for more information." to the message.

freddydk (approved):

Actions/AL-Go-Helper.ps1
aholstrup1: Yeah it probably wouldn't cause issues but best to avoid it

Actions/AL-Go-Helper.ps1
mazhelez: Similarly, we can add the date as well.

E.g.
```
After <date> this warning will become an error.
```
or
```
After <date> this setting will not be used by AL-Go and it could lead to build breaks or unwanted behavior.
```
depending on what way we want to go.

Actions/TelemetryHelper.psm1
```diff
@@ -212,4 +239,35 @@ function Add-TelemetryProperty() {
...
+function Trace-DeprecationWarning {
+    param(
+        [Parameter(Mandatory = $true)]
+        [String] $Message,
+        [Parameter(Mandatory = $true)]
+        [String] $DeprecationTag
+    )
+
+    # Show deprecation warning in GitHub
+    OutputWarning -message "$Message. This warning will become an error in future versions. See https://aka.ms/ALGoDeprecations#$($DeprecationTag) for more information."
```
mazhelez, line 267: Did we agree it'll be an error in the future? I thought AL-Go will quietly stop using this functionality. @freddydk what do you think?

aholstrup1: I added based on your suggestion 😆 I can remove the sentence from the message. That information should be present in the DEPRECATIONS.md anyway. 

As for what should happen after April 1st, we might want to consider throwing an error. Quietly removing it might lead to unexpected behavior if a partner misses the deprecation message. 

That will not be part of this PR though.

Actions/TelemetryHelper.psm1
```diff
@@ -212,4 +239,41 @@ function Add-TelemetryProperty() {
...
+    Trace-DeprecationWarning -Message "This setting is deprecated. Use the new setting instead." -DeprecationTag 'SettingName'
+#>
+function Trace-DeprecationWarning {
+    param(
+        [Parameter(Mandatory = $true)]
+        [String] $Message,
+        [Parameter(Mandatory = $true)]
+        [String] $DeprecationTag,
+        [Parameter(Mandatory = $false)]
+        [System.Collections.Generic.Dictionary[[System.String], [System.String]]] $AdditionalData = @{}
```
freddydk, line 268: Why is this not a hashtable type? (system.collections.gene...)

freddydk (approved):

Actions/TelemetryHelper.psm1
freddydk: Agree with not having this added to the message here - some deprecations might become errors, others might just stop working.

Actions/TelemetryHelper.psm1
aholstrup1: Not sure I remember why but I can check up on that. I can create a follow-up PR for that. 

Right now, this type is consistent with what we use for tracing "regular" telemetry as well so keeping it in for now.

PR #1388, opened 2025-01-10
Running PSScriptAnalyzer as local action.

spetersenms:
We are currently running PSScriptAnalyzer via the action in this repo: https://github.com/microsoft/psscriptanalyzer-action, however this has more settings and options than we need.
I recreated the action locally, and simplified it to fit our needs.


Actions/PSScriptAnalyzer/README.md
```diff
@@ -0,0 +1,28 @@
+# Read Power Platform Settings
```
mazhelez, line 1: Copy-paste much? 😄

Actions/PSScriptAnalyzer/action.yml
```diff
@@ -0,0 +1,43 @@
...
+  using: "composite"
+  steps:
+    - name: Run PSScriptAnalyzer and ConvertToSARIF
+      shell: pwsh
+      run: |
+        $analyzerModule = Get-Module -ListAvailable -Name PSScriptAnalyzer
+        if ($null -eq $analyzerModule) {
+          Install-Module -Name PSScriptAnalyzer -Force
+        }
+
```
mazhelez, line 31: Missing `Import-Module`.

spetersenms: I did realize that we don't import it, however this is how it is done in the previous action we used: https://github.com/microsoft/psscriptanalyzer-action/blob/main/action.yml

Actions/PSScriptAnalyzer/README.md
spetersenms: Good catch :p

Actions/PSScriptAnalyzer/action.yml
```diff
@@ -0,0 +1,43 @@
...
+branding:
+  icon: "check"
+  color: "gray-dark"
+inputs:
+  path: 
+    description: 'Specifies the path to the scripts or module to be analyzed. Wildcard characters are supported.'
+    required: true
+    default: '.\'
+  excludeRule:
+    description: 'Omits the specified rules from the Script Analyzer test. Wildcard characters are supported.'
```
mazhelez, line 12: This should be something like "comma-separated PSScriptAnalyzer rules...". Are wildcard characters really supported? Never tried it.

Actions/PSScriptAnalyzer/action.yml
mazhelez: 😮

Actions/PSScriptAnalyzer/action.yml
spetersenms: That might be a better description. Current one just pasted from the official action. Wildcards are indeed supported: https://github.com/PowerShell/PSScriptAnalyzer/blob/main/docs/Cmdlets/Invoke-ScriptAnalyzer.md#-excluderule

Of course, it does not matter too much since our workflow uses hardcoded values.

Actions/PSScriptAnalyzer/action.yml
```diff
@@ -0,0 +1,43 @@
...
+          Install-Module -Name ConvertToSARIF -Force
+        }
+        Import-Module -Name ConvertToSARIF -Force
+
+        $htPSA = [ordered]@{ Path = '${{ inputs.path }}'; }
+        Write-Output "Modules installed, now running tests."
+        if(![string]::IsNullOrEmpty('${{ inputs.excludeRule }}')) { $htPSA.add('ExcludeRule', @(${{ inputs.excludeRule }})) }
+        if(![string]::IsNullOrEmpty('${{ inputs.recurse }}')) { $htPSA.add('Recurse', $true) }
+        $htCTS = [ordered]@{ FilePath = '${{ inputs.output }}'; }
+        Invoke-ScriptAnalyzer @htPSA -Verbose | ConvertTo-SARIF @htCTS
```
mazhelez, line 43: Wasn't the idea to have retry mechanism here to catch the "Object reference not set to an instance of an object." error? 🤔

Actions/PSScriptAnalyzer/action.yml
```diff
@@ -0,0 +1,62 @@
...
+        if(![string]::IsNullOrEmpty('${{ inputs.recurse }}')) { $htPSA.add('Recurse', $true) }
+        $htCTS = [ordered]@{ FilePath = '${{ inputs.output }}'; }
+
+        $maxRetries = 3
+        $retryCount = 0
+        $success = $false
+
+        while (-not $success -and $retryCount -lt $maxRetries) {
+          Try {
+            Invoke-ScriptAnalyzer @htPSA -Verbose | ConvertTo-SARIF @htCTS
```
aholstrup1, line 50: Took a look at one of the runs: https://github.com/microsoft/AL-Go/actions/runs/12711830687/job/35436027525?pr=1388

This seems quite verbose. IMO we can live without that level of verbosity but I'm also okay leaving it in. 

```suggestion
            Invoke-ScriptAnalyzer @htPSA | ConvertTo-SARIF @htCTS
```

Actions/PSScriptAnalyzer/action.yml
```diff
@@ -0,0 +1,62 @@
...
+            $success = $true
+          } Catch {
+            Write-Error $_
+            $retryCount++
+            Write-Output "Retrying... ($retryCount/$maxRetries)"
+          }
+        }
+
+        if (-not $success) {
+          Write-Error "Failed after $maxRetries attempts."
```
aholstrup1, line 60: Consider writing it to a [GitHub error](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions?tool=powershell#setting-an-error-message) 


E.g. `Write-Host "::Error:: Failed after..."`

Actions/PSScriptAnalyzer/action.yml
spetersenms: I agree that it's quite noisy, but I don't think we need it, at least temporary, to debug the random null reference failures.

We can collect some logs on next failure and open an issue on the PSScriptAnalyzer repo, or at least confirm if it's the same issue as here: https://github.com/PowerShell/PSScriptAnalyzer/issues/1867#issuecomment-1340941069

Actions/PSScriptAnalyzer/action.yml
spetersenms: Sorry, I had misunderstood as doing separate debugging of the issue. I added retry logic to the script.

Actions/PSScriptAnalyzer/action.yml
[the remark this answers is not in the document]
spetersenms: Thanks for the hint, that seems like a better way to do it.

aholstrup1 (approved):

Actions/PSScriptAnalyzer/action.yml
```diff
@@ -0,0 +1,62 @@
...
+    description: 'Specifies where the path for the sarif file'
+    required: true
+    default: 'results.sarif'
+runs:
+  using: "composite"
+  steps:
+    - name: Run PSScriptAnalyzer and ConvertToSARIF
+      shell: pwsh
+      run: |
+        $analyzerModule = Get-Module -ListAvailable -Name PSScriptAnalyzer
```
mazhelez, line 27: Consider extracting the script in a .ps1 file. Powershell in YAML isn't as readable and, more importantly, isn't getting analyzed.

mazhelez (dismissed):
Great first addition! Left some comments, but otherwise, let's merge it!

Actions/PSScriptAnalyzer/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+### ENV variables
+
+none
+
+### Parameters
+
+| Name | Required | Description | Default value |
+| :-- | :-: | :-- | :-- |
+| path | Yes | Specifies the path to the scripts or module to be analyzed. Wildcard characters are supported. | powershell |
```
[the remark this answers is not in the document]
spetersenms: Fixed :)

aholstrup1 (dismissed):

Actions/RunPSScriptAnalyzer/RunPSScriptAnalyzer.ps1
```diff
@@ -0,0 +1,43 @@
+param (
+    [string]$Path,
+    [string]$ExcludeRule,
+    [string]$Recurse,
```
mazhelez, line 4: Consider making this a `switch` instead.

Actions/RunPSScriptAnalyzer/RunPSScriptAnalyzer.ps1
```diff
@@ -0,0 +1,43 @@
+param (
+    [string]$Path,
+    [string]$ExcludeRule,
```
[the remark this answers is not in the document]
freddydk: The documentation states it is a comma separated list

Actions/RunPSScriptAnalyzer/RunPSScriptAnalyzer.ps1
spetersenms: It will be a bit cleaner in the script, but a bit more messy in the action I think. I would need to do something like this to ensure it's parsed correctly:
`$recurseSwitch = if ($ENV:_recurse -eq 'true') { '-recurse' } else { '' }`
`${{ github.action_path }}/../Invoke-AlGoAction.ps1 -ActionName "RunPSScriptAnalyzer" -Action {`
`    ${{ github.action_path }}/RunPSScriptAnalyzer.ps1 -path $ENV:_path -excludeRule @($ENV:_excludeRule -Split ',') $recurseSwitch -output $ENV:_output
        }`

Actions/RunPSScriptAnalyzer/RunPSScriptAnalyzer.ps1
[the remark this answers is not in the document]
spetersenms: You are both right. The action itself takes a string list, but the script should accept an array.
The confusion happens because before when the script was inline in the action, we could do this:
`if(![string]::IsNullOrEmpty('${{ inputs.excludeRule }}')) { $htPSA.add('ExcludeRule', @(${{ inputs.excludeRule }})) }`
That had to be changed since we can't directly use the inputs in the script, so instead in the action we can pass the list like this:
`-excludeRule @($ENV:_excludeRule -Split ',')`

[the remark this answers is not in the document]
spetersenms: I still run the check like this to see if the param in the script should be included:
`if (![string]::IsNullOrEmpty($ExcludeRule)) { $htPSA.add('ExcludeRule', $ExcludeRule) }`
But it still behaves as expected, even if $ExcludeRule has the string[] type instead of string.

aholstrup1 (approved):

Actions/RunPSScriptAnalyzer/RunPSScriptAnalyzer.ps1
spetersenms: Oh that's a cool syntax, did not know about it. That makes more sense to use.

Actions/RunPSScriptAnalyzer/RunPSScriptAnalyzer.ps1
```diff
@@ -0,0 +1,43 @@
...
+
+$sarifModule = Get-Module -ListAvailable -Name ConvertToSARIF
+if ($null -eq $sarifModule) {
+    Install-Module -Name ConvertToSARIF -Force
+}
+Import-Module -Name ConvertToSARIF -Force
+
+$htPSA = [ordered]@{ Path = $Path }
+Write-Output "Modules installed, now running tests."
+if (![string]::IsNullOrEmpty($ExcludeRule)) { $htPSA.add('ExcludeRule', $ExcludeRule) }
```
mazhelez, line 21:
```suggestion
if ($ExcludeRule) { $htPSA.add('ExcludeRule', $ExcludeRule) }
```

aholstrup1 (dismissed):

freddydk (dismissed):

PR #1435, opened 2025-02-06
Workshop Feedback

freddydk:
InstallApps and installTestApps settings can now contain secrets, by specifying ${{SECRETNAME}} as the url or part of the url.
In the build, the secret (which ever part of the URL it is) is masked. Example, this:

`"installApps": [ "https://www.dropbox.com/${{SECRETNAME}}&dl=1" ]`

will be shown as:

{687E3848-5B45-4539-9087-D61421F98EFD}

Also...

DeliverTo*.ps1 and DeployTo*.ps1 scripts in .github folder are now no longer cases sensitive under linux.


Actions/ReadSecrets/ReadSecrets.ps1
```diff
@@ -77,6 +77,21 @@ try {
                 }
             }
         }
+        # Look through installApps and installTestApps for secrets and add them to the collection of secrets to get
+        foreach($installSettingsKey in @('installApps','installTestApps')) {
+            if ($settings.Keys -contains $installSettingsKey) {
+                $settings."$installSettingsKey" | ForEach-Object {
+                    # If any of the installApps URLs contains '${{SECRETNAME}}' we need to get the secret
+                    $pattern = '.*(\$\{\{\s*([^}]+?)\s*\}\}).*'
```
mazhelez, line 85: Minor: not sure why you match two groups here. `'.*\$\{\{(.*)\}\}.*'` should suffice.

Also, could there be a case that, for whatever reason, two secrets are used in one entry?
Not saying AL-Go should support this, just making an observation.

mazhelez: Got it why you also match the `${{ ... }}` group: to re-use the same pattern when replacing with the actual secret.

mazhelez (approved):

Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -152,11 +152,23 @@ try {
...
+    }
 
-    $installApps += $installAppsJson | ConvertFrom-Json
-    $installTestApps += $installTestAppsJson | ConvertFrom-Json
+    # Replace secret names in install.apps and install.testApps
+    foreach($list in @('apps','testApps')) {
+        $install."$list" = @($install."$list" | ForEach-Object {
+            $pattern = '.*(\$\{\{\s*([^}]+?)\s*\}\}).*'
+            if ($_ -match $pattern) {
+                $_.Replace($matches[1],[System.Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String($secrets."$($matches[2])")))
```
aholstrup1, line 165: Would it make sense to check if the URL is valid (and the file accessible) here and fail with a useful error message? Something like this
`Invoke-WebRequest -Uri $_ -Method Head`

If your URL isn't valid for some reason (File was moved, link expired etc.) we might want to fail the run before we get too far.

freddydk: That sounds like a good idea - will add that.

Actions/ReadSecrets/ReadSecrets.ps1
freddydk: Correct - I could remove one parentheses here, but for the sake of using the same and not having to decipher.

Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -152,11 +152,30 @@ try {
...
+            }
+            else {
+                $_
+            }
+            # Check validity of URL
+            try {
+                Invoke-WebRequest -Method Head -UseBasicParsing -Uri $_ | Out-Null
+            }
+            catch {
+                throw "Setting: install$($list) contains an inaccessible URL: $($_). Error was: $($_.Exception.Message)"
```
aholstrup1, line 175: At this point we've inserted the secret into `$_` so maybe we shouldn't throw an error with the full URL in it. I know GitHub will try to mask it but still... 
Maybe it'd be safer to log the original URL without the secret in it?

aholstrup1 (dismissed):

Actions/RunPipeline/RunPipeline.ps1
freddydk: Actually we haven't - we have written it to the output stream, why this also wouldn't work...

spetersenms (approved):

mazhelez (approved):

PR #1437, opened 2025-02-07
Deploy test apps to sandbox & exclude apps based on id

### LINKED ISSUE #1309: [Question]: Deploy Test Apps to Environment
### Question

In an attempt to reduce the minutes consumed when running the CI/CD action, I have all actions using ```"runs-on": "ubuntu-latest"``` and ```"githubRunner": "ubuntu-latest"```; this working fine and save a considerable amount of minutes. I am also deploying to a sandbox environment after the Build.

However, using these settings means any test apps are not run and the apps are only deployed to the sandbox. Is there a method for automatically deploying the test apps to the sandbox? I could then run them manually using the Test Tool.

Any advice would be appreciated.

Regards,

Michael

spetersenms:
Fixes #1309

Adds functionality to deploy test apps to sandbox environments, and exclude apps based on id with 2 new properties in the DeployTo setting.

"DeployTospetersen_demo": {
    "includeTestAppsInSandboxEnvironment": true,
    "excludeAppIds": ["id1", "id2"]
  }

If a test app depends on Test-TestLibraries, it will automatically be skipped as we can't install that app in an online env. Still need to figure out if any other dependencies like that exist.


Actions/Deploy/Deploy.ps1
```diff
@@ -138,7 +187,18 @@
...
         Write-Host "project '$project'"
+
+        $allApps = @()
         $projectApps = @((Get-ChildItem -Path $artifactsFolder -Filter "$project-$refname-$($buildMode)Apps-*.*.*.*") | ForEach-Object { $_.FullName })
+        $projectTestApps = @()
+        $unknownDependencies = @()
+        if ($deploymentSettings.includeTestAppsInSandboxEnvironment) {
+            Write-Host "Including test apps for deployment"
+            $projectTestApps = @((Get-ChildItem -Path $artifactsFolder -Filter "$project-$refname-$($buildMode)TestApps-*.*.*.*") | ForEach-Object { $_.FullName })
+        }
```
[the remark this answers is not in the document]
spetersenms: Fixed.

Actions/Deploy/Deploy.ps1
```diff
@@ -148,8 +208,58 @@
...
+                            }
+                            Write-Host "App file $($_.Name) with id $($appJson.id) included in deployment"
+                        }
+                        else {
+                            Write-Host "App file $($_.Name) with id $($appJson.id) excluded from deployment"
+                        }
+                    } 
+                    else {
+                        $apps += $_.FullName
+                    }
```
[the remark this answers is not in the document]
spetersenms: Fixed.

Actions/Deploy/Deploy.ps1
```diff
@@ -148,8 +208,65 @@
...
+                                if ($unknownDependencies -notcontains $_) {
+                                    $unknownDependencies += $_
+                                }
+                            }
+                            Write-Host "App file $($app.Name) with id $($appJson.id) included in deployment"
+                        }
+                        else {
+                            Write-Host "App file $($app.Name) with id $($appJson.id) excluded from deployment"
+                        }
+                    }
```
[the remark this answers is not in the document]
spetersenms: Fixed.

spetersenms:
Still need to figure out how to best handle custom deployments. Currently the unknown dependencies list is not passed to the custom script.

Actions/Deploy/Deploy.ps1
```diff
@@ -51,53 +51,102 @@ function InstallOrUpgradeApps {
...
-            # Get AppJson (works for full .app files, symbol files and also runtime packages)
-            $appJson = Get-AppJsonFromAppFile -appFile $app
-            $isPTE = ($appjson.idRanges.from -lt 100000 -and $appjson.idRanges.from -ge 50000)
-            $installedApp = $installedApps | Where-Object { $_.id -eq $appJson.id }
-            $needsInstall, $needsUpgrade = CheckIfAppNeedsInstallOrUpgrade -appJson $appJson -installedApp $installedApp -installMode $installMode
-            if ($needsUpgrade) {
-                if (-not $isPTE -and $installedApp.publishedAs.Trim() -eq 'Dev') {
-                    Write-Host "::WARNING::Dependency AppSource App $($appJson.name) is published in Dev scoope. Cannot upgrade."
-                    $needsUpgrade = $false
+    #If regular dependencies we have the app files, otherwise in case of unknown dependencies we have the app ids
```
freddydk, line 63: I would prefer another function called InstallUnknownDependencies and leave InstallOrUpgradeApps as is

Actions/Deploy/Deploy.ps1
```diff
@@ -148,7 +208,56 @@ if (Test-Path $artifactsFolder -PathType Container) {
...
+        #Go through all .app files and exclude any with ids in the excludeAppIds list
+        if ($allApps) {
+            foreach($appFolder in $allApps) {
+                foreach($app in (Get-ChildItem -Path $appFolder)) {
+                    Write-Host "Processing app: $($app.Name)"
+                    if ($app.Name -like "*.app") {
+                        $unknownDependenciesForApp = @()
+                        Sort-AppFilesByDependencies -appFiles @($app.FullName) -unknownDependencies ([ref]$unknownDependenciesForApp) | Out-Null
+                        # Check if Test-TestLibraries is a dependency for the app - if so skip it since we can't install that dependency
+                        $testLibraryDependencyFound = $false
```
freddydk, line 230: We should determine the dependency on Test-Test_Libraries on the appid, which is available in `$TestsTestLibrariesAppId` (5d86850b-0d76-4eca-bd7b-951ad998e997) - defined in AL-Go-Helper.ps1

Actions/Deploy/Deploy.ps1
```diff
@@ -148,7 +208,56 @@ if (Test-Path $artifactsFolder -PathType Container) {
...
+            $allApps += $projectTestApps
+        }
+        #Go through all .app files and exclude any with ids in the excludeAppIds list
+        if ($allApps) {
+            foreach($appFolder in $allApps) {
+                foreach($app in (Get-ChildItem -Path $appFolder)) {
+                    Write-Host "Processing app: $($app.Name)"
+                    if ($app.Name -like "*.app") {
+                        $unknownDependenciesForApp = @()
+                        Sort-AppFilesByDependencies -appFiles @($app.FullName) -unknownDependencies ([ref]$unknownDependenciesForApp) | Out-Null
```
freddydk, line 228: Instead of throwing the return value away here - you can use that to get the AppId of the `$app`.
The return value will be `appId:appName`, so if you grab `returnvalue.split(':')[0}` you will have the appid and can determine whether it is excluded before looking at dependencies and issuing a warning before excluding it.

Actions/DetermineDeploymentEnvironments/DetermineDeploymentEnvironments.ps1
```diff
@@ -183,6 +187,10 @@ else {
             if ($deploymentSettings."shell" -ne 'pwsh' -and $deploymentSettings."shell" -ne 'powershell') {
                 throw "The shell setting in $settingsName must be either 'pwsh' or 'powershell'"
             }
+            if ($deploymentSettings."includeTestAppsInSandboxEnvironment" -and ($environmentName -like '* (PROD)' -or $environmentName -like '* (Production)' -or $environmentName -like '* (FAT)' -or $environmentName -like '* (Final Acceptance Test)')) {
```
freddydk, line 190: You already give a warning like this later during deployment (based on whether it actually is a sandbox environment) - I don't think i should be done here based on the naming algorithm

freddydk (changes requested):
Remember release notes and settings.md

Actions/Deploy/Deploy.ps1
spetersenms: Looking at the implementation of Sort-AppFilesByDependencies, it seems it only returns the dependencies on the format of `appId:appName`, while input apps are only returned as is in sorted order.
The point of looking if the app should be excluded before getting dependencies is still good though.

Actions/Deploy/Deploy.ps1
spetersenms: I don't mind either way, so I moved it to a separate function.

Actions/Deploy/Deploy.ps1
spetersenms: Thanks for making be aware of this var. Changed to use it instead of name based matching.

Actions/Deploy/Deploy.ps1
```diff
@@ -148,7 +212,51 @@
...
+                            #If app should be included, verify that it does not depend on Tests-TestLibraries
+                            $unknownDependenciesForApp = @()
+                            Sort-AppFilesByDependencies -appFiles @($app.FullName) -unknownDependencies ([ref]$unknownDependenciesForApp) | Out-Null
+                            $unknownDependenciesForApp | ForEach-Object {
+                                if ($_.Split(':')[0] -eq $TestsTestLibrariesAppId) {
+                                    Write-Host "::WARNING::Test-TestLibraries can't be installed - skipping app $($app.Name)"
+                                    continue
+                                }
+                            }
+
```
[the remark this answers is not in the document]
spetersenms: Fixed.

Actions/Deploy/Deploy.ps1
```diff
@@ -100,6 +101,58 @@ function InstallOrUpgradeApps {
...
+            $appId, $appName = $app.Split(':')
+            $appVersion = ""
+            if ($appName -like "*Tests-TestLibraries*") {
+                Write-Host "::WARNING::Tests-TestLibraries can't be installed"
+                continue
+            }
+            if ($appName -match "_(\d+\.\d+\.\d+\.\d+)\.app$") {
+                $appVersion = $matches.1
+            } else {
+                throw "Version not found for unknown dependency $app"
```
freddydk, line 127: when would we ever get in this situation?
and what should the user/partner do about that?
wouldn't it be better to just silently ignore if there is a dependency we cannot find/use - then the installation will fail later with a more explanatory error message: Dependency xxx is missing.

freddydk (changes requested):
Mostly smaller things

Actions/Deploy/Deploy.ps1
```diff
@@ -148,7 +212,51 @@ if (Test-Path $artifactsFolder -PathType Container) {
...
+                throw "There are no artifacts present in $artifactsFolder matching $project-$refname-$($buildMode)TestApps-<version>."
+            }
+        }
+        else {
+            $allApps += $projectTestApps
+        }
+        #Go through all .app files and exclude any with ids in the excludeAppIds list
+        if ($allApps) {
+            foreach($appFolder in $allApps) {
+                foreach($app in (Get-ChildItem -Path $appFolder)) {
```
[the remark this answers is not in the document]
spetersenms: Does a scenario exist where something other than a .app exist in this folder that we want to deploy? I included it together with the "catch-all" else at line 255 to add anything that might not be a .app, but if nothing like that exist, I can just filter directly in Get-ChildItem.

Actions/Deploy/Deploy.ps1
spetersenms: I agree it makes sense not to throw an error, but it might still be helpful to write $appName as either a warning or just regular message in the terminal?
If it does actually fail to find the version, it might be helpful for us for debugging if the partner can show us the app name incase something actually goes wrong.

But you know the system better than me, so if you are confident that the app name returned by Sort-AppFilesByDependencies always include the version in correct format, I am happy to remove the else part completely.

Actions/Deploy/Deploy.ps1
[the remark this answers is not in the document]
spetersenms: I can see from the other comment that this should not be a case.

Actions/Deploy/Deploy.ps1
freddydk: You are right that in debugging scenarios, it would be good to have a note specifying that this dependency from that app doesn't include the version number, ignoring dependency, so basically a Write-Host instead of the throw and then continue.

aholstrup1 (approved):

freddydk (approved):

PR #1441, opened 2025-02-10
Publish to Environment from PR

spetersenms:
Adds support to run the Publish to Environment from a PR by passing a string like "PR_x" in the artifact version input.
A PR with id x will be found, and artifacts from latest successful build in that PR will be used if it exist.


freddydk:
This one will have wait until the incremental builds is merged - and then it will have to find the artifacts slightly differently for incremental builds.

Actions/GetArtifactsForDeployment/GetArtifactsForDeployment.ps1
```diff
@@ -51,13 +52,31 @@ if ($artifactsVersion -eq "current" -or $artifactsVersion -eq "prerelease" -or $
...
+elseif ($artifactsVersion -like "PR_*") {
+    $prId = $artifactsVersion -replace "PR_", ""
+    if (!($prId -as [int])) {
+        throw "Invalid PR id: $prId"
+    }
+    $prBranch = GetBranchFromPRId -repository $ENV:GITHUB_REPOSITORY -prId $prId -token $token
+    if (!($prBranch)) {
+        throw "Unable to locate branch for PR $prId"
+    }
+    $latestSuccessfulPRRun = FindLatestSuccessfulPRRun -repository $ENV:GITHUB_REPOSITORY -branch $prBranch -token $token
```
freddydk, line 64: When you find the PR - I don't think you should go looking for a good build within the PR.
If the last build failed or a build is running - I think you should give an error.

and... - with the latest changes to incremental builds you might need to download artifacts from the last known good CI/CD build

freddydk (changes requested):

freddydk:
We also need to add this to release notes.

Actions/Github-Helper.psm1
```diff
@@ -1226,3 +1296,32 @@ function GenerateJwtForTokenRequest {
...
+        [string] $prId,
+        [Parameter(Mandatory = $true)]
+        [string] $token
+    )
+
+    $headers = GetHeaders -token $token
+
+    $pullsURI = "https://api.github.com/repos/$repository/pulls/$prId"
+    Write-Host "- $runsURI"
+    $workflowRuns = (InvokeWebRequest -Headers $headers -Uri $pullsURI).Content | ConvertFrom-Json
```
mazhelez, line 1324: Variable name isn't correct. Just `$pr` maybe?

Actions/Github-Helper.psm1
```diff
@@ -911,6 +911,76 @@ function FindLatestSuccessfulCICDRun {
...
+            }
+
+            # CICD run is considered successful if all build jobs were successful
+            $areBuildJobsSuccessful = CheckBuildJobsInWorkflowRun -workflowRunId $($PRRun.id) -token $token -repository $repository
+
+            if($areBuildJobsSuccessful) {
+                $lastSuccessfulPRRun = $PRRun.id
+                Write-Host "Found last successful PR run: $($lastSuccessfulPRRun), from $($PRRun.created_at)"
+                break
+            }
```
aholstrup1, line 964: Is this needed for PR Builds? 🤔 PR build only build and test. I think this was needed for CI/CD builds because a CI/CD can have a successful build portion but fail during deployment. Could be wrong though

Actions/GetArtifactsForDeployment/GetArtifactsForDeployment.ps1
spetersenms: That is probably a better approach. I changed the logic to now only find the latest build, and fail if it is not done or successful.

Actions/Github-Helper.psm1
spetersenms: Good catch :)

Actions/Github-Helper.psm1
```diff
@@ -911,6 +911,76 @@ function FindLatestSuccessfulCICDRun {
...
+    )
+
+    $headers = GetHeaders -token $token
+    $lastSuccessfulPRRun = 0
+    $per_page = 100
+    $page = 1
+
+    Write-Host "Finding latest successful PR run for branch $branch in repository $repository"
+
+    # Get the latest CICD workflow run
```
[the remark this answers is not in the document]
spetersenms: Fixed.

Actions/Github-Helper.psm1
spetersenms: There is still a Pull Request Status Check job which could maybe fail, even if the build is successful, but maybe that is not possible?

aholstrup1: It could but that would be due to a bug in AL-Go. Pull Request Status Check is supposed to be an indicator of whether all the other jobs succeeded. If there are any failed jobs, the status check should also fail. If all other jobs succeed, the status check should succeed.

freddydk: Agree, instead of running through jobs - just check whether the status check is succeeded - then we are good to go,
I think people would find it strange if we were to deploy something that didn't have status check = Succeeded

aholstrup1: I don't think you even need to do that. Just check whether the entire workflow succeeded. If the entire workflow succeeded, we know that the "Pull Request Status Check" also did.

spetersenms: Yeah it would be easier to just check that the workflow itself is successful. I already check for that further up, so I can just remove the build job check completely, and the rest should work as expected.

Actions/Github-Helper.psm1
```diff
@@ -911,6 +911,73 @@ function FindLatestSuccessfulCICDRun {
     return $lastSuccessfulCICDRun
 }
 
+<#
```
freddydk, line 914: Digged a little deeper into this...

Currently, you query the branch on which a PR with a specific number was created and then you enumerate runs on that branch. Problem here is that if this branch is in a fork, it might be confused with a branch in this repo. Worst case scenario is if somebody is using the main branch in a fork, you will be enumerating builds from the main branch in the target repo.

Instead, you should use the pulls API to get the PR (as you do), grab the head.sha from that (this is the latest commit SHA from the PR and the one the user wants to publish).
Now use the runs API with head_sha set to the last commit SHA and you will get all builds for THIS sha - inspect only the latest - no reason for paging.
If no builds: Error - no builds exists for the latest commit on PR_xxx (PR_xxx could be a link to the PR)
If latest build is in_progress: Error - build is still in progress. (build could be a link to the build)
If latest build is not successful: Error - build wasn't successful. (build could be a link to the build)
On all you can add: Please re-run this workflow when you have a successful build on PR_xxx (PR_xxx could be a link to the PR)

Actions/GetArtifactsForDeployment/GetArtifactsForDeployment.ps1
```diff
@@ -51,13 +52,31 @@ if ($artifactsVersion -eq "current" -or $artifactsVersion -eq "prerelease" -or $
...
+    }
+}
 else {
     $searchArtifacts = $true
 }
 
 if ($searchArtifacts) {
     $allArtifacts = @()
-    'Apps','Dependencies','PowerPlatformSolution' | ForEach-Object {
+    $artifactsToDownload | ForEach-Object {
```
freddydk, line 79: What happens here if the artifacts from the PR build has expired?

We should throw an error stating that the build artifacts have expired, stating that people would need to re-run the pull-request build (and add a link to the PR Run). You could also add a note that the retention period for Short Lived Artifacts (such as PR builds) is controlled by the AL-Go setting [shortLivedArtifactsRetentionDays](https://aka.ms/algosettings#shortLivedArtifactsRetentionDays)

freddydk (changes requested):

Actions/Github-Helper.psm1
```diff
@@ -1226,3 +1293,32 @@ function GenerateJwtForTokenRequest {
...
+        [Parameter(Mandatory = $true)]
+        [string] $prId,
+        [Parameter(Mandatory = $true)]
+        [string] $token
+    )
+
+    $headers = GetHeaders -token $token
+
+    $pullsURI = "https://api.github.com/repos/$repository/pulls/$prId"
+    Write-Host "- $runsURI"
```
freddydk, line 1320: $runsURI is undefined.

freddydk (changes requested):

Actions/Github-Helper.psm1
spetersenms: That makes sense. I agree with your solution, except I think we still need paging. 

Using the runs API with the head_sha can return other workflow runs in addition to pr builds. It might not be very likely, but it's possible that you might want to deploy from one pr to multiple environments. This means the latest run returned from the API can be something else. It's even more unlikely that a 100+ other workflows have been run since the latest pr build, but it is technically possible.

The current approach, without your suggestions uses paging, but only looks at the latest pr build. It does need better errors as your suggest though.

Actions/GetArtifactsForDeployment/GetArtifactsForDeployment.ps1
spetersenms: I added a new ref parameter to GetArtifactsFromWorkflowRun which adds any expired artifacts. We can potentially use that elsewhere too.

Actions/Github-Helper.psm1
```diff
@@ -1136,6 +1264,71 @@ function DownloadArtifact {
...
+        InvokeWebRequest -Headers $headers -Uri $artifact.archive_download_url -OutFile $filename
+        if (Test-Path $foldername) {
+            Remove-Item $foldername -Recurse -Force
+        }
+        Expand-Archive -Path $filename -DestinationPath $foldername
+        Remove-Item $filename -Force
+
+        (Get-ChildItem -Path $foldername -Filter "*_*_*.*.*.*.app") | ForEach-Object {
+            Write-Host "Debug - artifact child from PR: $($_.FullName)"
+            $appName = $_.Name.Split('_')[1]
```
freddydk, line 1294: You cannot be sure that the appName is this, the _ are just for readability. The publishers name could be `My_Publisher_Name`, then you would find Publisher as appName for all apps. You should compare on everything but the version number.
Create a regex, which matches the version number and everything in front - and copy artifacts from last known good build if is doesn't exist already.

freddydk (changes requested):
Getting there - a few logic changes and I want to try this branch on a few test repos

Actions/GetArtifactsForDeployment/GetArtifactsForDeployment.ps1
```diff
@@ -52,16 +54,59 @@ if ($artifactsVersion -eq "current" -or $artifactsVersion -eq "prerelease" -or $
...
+    }
+    else {
+        Write-Host "No last known good build found: $lastKnownGoodBuildId"
+    }
+
+    $expiredArtifacts = @()
+    $lastKnownGoodBuildArtifacts = @()
+    #Get PR artifacts
+    $artifactsToDownload | ForEach-Object {
+        $allArtifacts += GetArtifactsFromWorkflowRun -workflowRun $latestPRBuildId -token $token -api_url $ENV:GITHUB_API_URL -repository $ENV:GITHUB_REPOSITORY -mask $_ -projects $projects -expiredArtifacts ([ref]$expiredArtifacts)
```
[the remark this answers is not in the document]
freddydk: Actually allArtifacts is because it is used to download later - maybe downloadArtifacts instead

Actions/Github-Helper.psm1
```diff
@@ -914,6 +914,129 @@ function FindLatestSuccessfulCICDRun {
         return 0, ''
     }
 }
+function FindPRRunAnnotationForIncrementalBuilds {
```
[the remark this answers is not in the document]
spetersenms: Yeah I guess that is better.

Actions/Github-Helper.psm1
spetersenms: I came up with an even simpler solution, I can just index with -1 to get the version and file ending like this `$_.Name.Split('_')[-1]` and then use replace on the full name to be left with publisher and app name.

Actions/GetArtifactsForDeployment/GetArtifactsForDeployment.ps1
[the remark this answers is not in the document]
spetersenms: I ended up going with prArtifacts, but had to change the way allArtifacts was used too, due to other changes.

Actions/GetArtifactsForDeployment/GetArtifactsForDeployment.psm1
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    #If the initialization annotation exist, check if a message pointing to last good build exist.
+    $annotationMessage = ''
+    if($annotationsURI) {
+        Write-Host "- $annotationsURI"
+        $annotations = (InvokeWebRequest -Headers $headers -Uri $annotationsURI).Content | ConvertFrom-Json
+        if($annotations -and $annotations.count -gt 0) {
+            foreach($annotation in $annotations) {
+                if($annotation.message -like "Last known good build: https://github.com/$repository/actions/runs/*") {
```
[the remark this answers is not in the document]
spetersenms: Good idea, this makes it a bit cleaner to look at.

Actions/GetArtifactsForDeployment/GetArtifactsForDeployment.ps1
```diff
@@ -21,6 +26,7 @@
     New-Item $artifactsFolder -ItemType Directory | Out-Null
 }
 $searchArtifacts = $false
+$downloadArtifacts = $false
```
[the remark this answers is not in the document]
spetersenms: Removed, as this is no longer needed.

Actions/GetArtifactsForDeployment/GetArtifactsForDeployment.psm1
```diff
@@ -0,0 +1,277 @@
...
+        [Parameter(Mandatory = $true)]
+        [string]$repository,
+        [Parameter(Mandatory = $true)]
+        [string]$checkSuiteId,
+        [Parameter(Mandatory = $true)]
+        [string] $token
+    )
+
+    $headers = GetHeaders -token $token
+    $lastKnownGoodBuildId = 0
```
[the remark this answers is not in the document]
spetersenms: Fixed.

freddydk (changes requested):
Small things really - I think the PR is almost ready.

Actions/GetArtifactsForDeployment/GetArtifactsForDeployment.psm1
```diff
@@ -0,0 +1,340 @@
...
+        [string]$token,
+        [Parameter(Mandatory = $true)]
+        [string]$path,
+        [Parameter(Mandatory = $true)]
+        $prArtifacts,
+        [Parameter(Mandatory = $false)]
+        $lastKnownGoodBuildArtifacts,
+        [Parameter(Mandatory = $true)]
+        [string]$prRunId,
+        [Parameter(Mandatory = $false)]
```
aholstrup1, line 203:
```suggestion
        [Parameter(Mandatory = $true)]
        [string]$lastKnownGoodBuildRunId
```
Looks like it's mandatory when you call GetHeadRefFromRunId

Actions/GetArtifactsForDeployment/GetArtifactsForDeployment.psm1
```diff
@@ -0,0 +1,340 @@
...
+    .PARAMETER lastKnownGoodBuildArtifacts
+    List of artifacts published by the last known good build, linked from the PR build
+    .PARAMETER prRunId
+    Run id of the PR build
+    .PARAMETER lastKnownGoodBuildRunId
+    Run id of the last known good build
+#>
+function DownloadPRArtifacts {
+    Param(
+        [string]$token,
```
aholstrup1, line 194:
```suggestion
        [Parameter(Mandatory = $true)]
        [string]$token,
```

Actions/GetArtifactsForDeployment/GetArtifactsForDeployment.psm1
```diff
@@ -0,0 +1,340 @@
...
+        }
+    }
+
+    # If the initialization annotation exist, check if a message pointing to last good build exist.
+    if($annotationsURI) {
+        Write-Host "- $annotationsURI"
+        $annotations = (InvokeWebRequest -Headers $headers -Uri $annotationsURI).Content | ConvertFrom-Json
+        if($annotations -and $annotations.count -gt 0) {
+            foreach($annotation in $annotations) {
+                if($annotation.message -match "https://github.com/$repository/actions/runs/([0-9]{1,11})") {
```
aholstrup1, line 50: This feels like something we could easily break unknowingly. 

Did you look into whether we could publish a "lastknowngoodbuild" artifact to the PR? Alternatively, maybe we could match on the full message rather than just a URL? 

Another idea could be to just find the latest good CI/CD build at the time of running the deployment. If the latest good artifact is incompatible with your PR Build you probably need to run a new PR Build anyway + possible fix something in your code to make your PR compatible with main again

Actions/Deploy/Deploy.ps1
```diff
@@ -7,7 +9,9 @@ Param(
     [ValidateSet('CD','Publish')]
     [string] $type = "CD",
     [Parameter(HelpMessage = "The settings for all Deployment Environments", Mandatory = $true)]
-    [string] $deploymentEnvironmentsJson
+    [string] $deploymentEnvironmentsJson,
+    [Parameter(HelpMessage = "Artifacts version. Used to check if this is a deployment from a PR", Mandatory = $false)]
+    [string]$artifactsVersion = ''
 )
```
freddydk, line 15:
```suggestion
    [string] $artifactsVersion = ''
)
```

Actions/GetArtifactsForDeployment/GetArtifactsForDeployment.ps1
```diff
@@ -52,6 +53,53 @@ if ($artifactsVersion -eq "current" -or $artifactsVersion -eq "prerelease" -or $
         }
     }
 }
+elseif ($artifactsVersion -like "PR_*") {
+    $prId = $artifactsVersion.Substring(3)
+    if (!($prId -as [int])) {
+        throw "Invalid PR id: $prId"
+    }
```
freddydk, line 60:
```suggestion
    $prId = $artifactsVersion.Substring(3)
    $i = 0
    if (!([Int]::TryParse($prId, [ref] $i))) {
        throw "Invalid PR id: $prId"
    }
```

Actions/GetArtifactsForDeployment/GetArtifactsForDeployment.psm1
freddydk: I would match the full message - we need the last known good build used when building this PR.

Actions/Deploy/Deploy.ps1
```diff
@@ -192,21 +197,28 @@ $artifactsFolder = Join-Path $ENV:GITHUB_WORKSPACE $artifactsFolder
 if (Test-Path $artifactsFolder -PathType Container) {
     $deploymentSettings.Projects.Split(',') | ForEach-Object {
         $project = $_.Replace('\','_').Replace('/','_')
+        $artifactVersionFilter = '*.*.*.*'
         $refname = "$ENV:GITHUB_REF_NAME".Replace('/','_')
+        #Artifacts from PRs are named differently - project-ref-Apps-PRx-date
+        if ($artifactsVersion -like "PR_*") {
+            $prId = $artifactsVersion -replace "PR_", ""
```
freddydk, line 204:
```suggestion
            $prId = $artifactsVersion.SubString(3)
```

Actions/GetArtifactsForDeployment/GetArtifactsForDeployment.ps1
freddydk, line 60:
```suggestion
    $i = 0
    if (!([Int]::TryParse($prId, [ref] $i))) {
        throw "Invalid PR id: $prId"
    }
```

Actions/GetArtifactsForDeployment/GetArtifactsForDeployment.ps1
```diff
@@ -52,6 +53,53 @@ if ($artifactsVersion -eq "current" -or $artifactsVersion -eq "prerelease" -or $
...
+        $prBuildLink = "https://github.com/$($ENV:GITHUB_REPOSITORY)/actions/runs/$latestPRBuildId"
+        $shortLivedRetentionSettingLink = "https://aka.ms/algosettings#shortLivedArtifactsRetentionDays"
+        throw "Build artifacts are expired, please re-run the pull request build ($prBuildLink). Note that you can control the retention days of short-lived artifacts using the AL-Go setting ShortLivedArtifactsRetentionDays. See $shortLivedRetentionSettingLink"
+    }
+
+    if ($prArtifacts) {
+        OutputMessageAndArray "Artifacts from pr build" $prArtifacts.Name
+    }
+    if ($lastKnownGoodBuildArtifacts) {
+        OutputMessageAndArray "Artifacts from last known good build" $lastKnownGoodBuildArtifacts.Name
```
freddydk, line 99:
```suggestion
        OutputMessageAndArray -message "Artifacts from last known good build" -arrayOfStrings $lastKnownGoodBuildArtifacts.Name
```

Actions/Deploy/Deploy.ps1
```diff
@@ -192,21 +197,28 @@ $artifactsFolder = Join-Path $ENV:GITHUB_WORKSPACE $artifactsFolder
 if (Test-Path $artifactsFolder -PathType Container) {
     $deploymentSettings.Projects.Split(',') | ForEach-Object {
         $project = $_.Replace('\','_').Replace('/','_')
+        $artifactVersionFilter = '*.*.*.*'
         $refname = "$ENV:GITHUB_REF_NAME".Replace('/','_')
+        #Artifacts from PRs are named differently - project-ref-Apps-PRx-date
```
freddydk, line 202:
```suggestion
        # Artifacts from PRs are named differently - project-ref-Apps-PRx-date
```

Actions/GetArtifactsForDeployment/GetArtifactsForDeployment.psm1
freddydk: Please also add a space between the type and the variable name - we do that all over.

freddydk:
Only minor things before we can get this one in and in preview

Actions/GetArtifactsForDeployment/GetArtifactsForDeployment.psm1
spetersenms: Yes, you are right.

Actions/GetArtifactsForDeployment/GetArtifactsForDeployment.psm1
spetersenms: For now, I went with just matching the full message.

It should not really be that fragile though. The only main points of failure is renaming the Initialization job, or changing the message, but that is probably not something we will just do randomly.

RELEASENOTES.md
```diff
@@ -64,6 +64,12 @@ If your installApps or installTestApps are secure URL, containing a secret token
 
 Which would hide the secret part of your URL instead of exposing it in clear text.
 
+### Support for deploying from PRs
+
+AL-Go for GitHub now supports deploying from a PR. When using the 'Publish To Environment' workflow, it is now possible to input 'PR_X' as the App version, where 'X' is the PR Id. This will deploy the artifacts from the latest PR build to the chosen environment, if that build is completed and successful.
+
+All apps, which were not built by the PR build will be deployed from the last known good build. You can find a notification on the PR build explaining which build is used as the last known good build.
```
freddydk, line 71:
```suggestion
### Support for deploying to sandbox environments from a pull request

AL-Go for GitHub now supports deploying from a PR. When using the 'Publish To Environment' workflow, it is now possible to input 'PR_X' as the App version, where 'X' is the PR Id. This will deploy the artifacts from the latest PR build to the chosen environment, if that build is completed and successful.

All apps, which were not built by the PR build will be deployed from the last known good build. You can find a notification on the PR build explaining which build is used as the last known good build.

> [!NOTE]
> When deploying a PR build to a sandbox environment, the app will get a special version number, which is: major.minor.maxint.run-number. This means that the sandbox environment likely needs to be deleted after the testing has ended.

```

freddydk: ALSO - please move the release note section up above v6.4 - this part is not in v6.4

freddydk (approved):
I think this one is ready for AL-Go preview:-)

nice job

Actions/ReadSettings/ReadSettings.ps1
```diff
@@ -43,9 +43,9 @@ if ($settings.versioningstrategy -ne -1) {
             $settings.appBuild = -1
             $settings.appRevision = $settings.runNumberOffset + [Int32]($ENV:GITHUB_RUN_NUMBER)
         }
-        15 { # Use maxValue
+        15 { # Use maxValue and RUN_NUMBER
             $settings.appBuild = [Int32]::MaxValue
-            $settings.appRevision = 0
+            $settings.appRevision = $settings.runNumberOffset + [Int32]($ENV:GITHUB_RUN_NUMBER)
```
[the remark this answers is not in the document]
freddydk: No, that code only runs when creating a development environment and publishing the apps

freddydk (approved):

PR #1445, opened 2025-02-10
Create Release fails due to recent changes to the AL-Go

### LINKED ISSUE #1440: [Bug]: Create Release fails due to recent changes to the AL-GO
### AL-Go version

6.3

### Describe the issue

Hey @freddydk, 

We just noticed that our "Create Release" fails with a strange error:

"No artifacts found for version 20250210"

After short investigation we found that it's looking at the artifacts created by "Test Current" workflow

Image

If I check older Test Current worklow run, I see that it did not generate any artifacts besides BuildOutput, and TestResults

Image

My "Create Release" settings looks like this

Image

### Expected behavior

"Test Current" should not generate app files. If this is new intended behavior, those artifacts should be ignored during release. 

### Steps to reproduce

1. Run CICD workflow
2. Run Test Current worklow
3. Create release with option latest. You should see that it takes artifacts from step 2, instead of step 1

_No response_

freddydk:
Fixes #1440

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -1,3 +1,7 @@
+### Issues
+
+- Issue 1440 Create Release fails due to recent changes to the AL-Go
+
 ### Support for GitHub App authentication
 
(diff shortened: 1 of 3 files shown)


spetersenms (approved):

HazardMK:
Are these changes already in 6.3? Still getting the same error for PTE. No AL-Go update available.

freddydk:
> Are these changes already in 6.3? Still getting the same error for PTE. No AL-Go update available.

This fix will be included in v6.4, which will ship next week.
It is currently included in preview.

PR #1450, opened 2025-02-12
CompilerFolder doesn't transfer installed Apps to NuGet resolution

### LINKED ISSUE #1268: [Bug]: Build pipeline sees base apps as missing NuGet dependencies
### AL-Go version

5bb4f65ffe64f5328a70a288d04640786aa467ec (modified)

### Describe the issue

This issue only occurred after updating my AL-Go files to the equivalent of 5bb4f65ffe64f5328a70a288d04640786aa467ec.
A repo building a BC14 app on a self-hosted Windows runner fails during the build step due to wrongfully registering the Microsoft Base application as a missing NuGet dependency. After failing to download the package, the workflow fails.

[logs_29671715468.zip](https://github.com/user-attachments/files/17412169/logs_29671715468.zip)

### Expected behavior

This is the log of the "Build" job from the last successful run, performed before updating the AL Go system files.

[logs_28303561294.zip](https://github.com/user-attachments/files/17412174/logs_28303561294.zip)

### Steps to reproduce

The workflow runs on a self-hosted Windows runner, and the BC version of the container is 14.12. There are no dependencies beyond the default Microsoft apps, neither locally nor to other GitHub projects.

_No response_

### LINKED ISSUE #1330: [Bug]: correct external appsource dependency not used when skipping docker (useCompilerFolder = true, doNotPublishApps = true)
### AL-Go version

6.0

### Describe the issue

We have an AL-Go appsource repository that depends on a third party appsource app not published by us. The publisher for that app hasn't provisioned a private nuget feed for their apps yet, so we're using the installApps property for now. 

our [main] branch in our repository currently targets BC24. as such, the installApps points to an appsource app URL (not a symbols-only package, and not a runtime package) targeted for runtime 13.0.

the build pipeline fails because AL-Go seems to be fetching the BC25 version (runtime 14) of that third party app, not the BC24 version (runtime 13), nor does the pipeline to use the BC24 version appended to installApps. 

when we use docker for compilation and publishing (useCompilerFolder = false, doNotPublishApps = false), the nuget packags is ignored and the installApps version is used. 

### Expected behavior

when no private nuget feeds nor installApps settings are used, nuget should download the correct nuget symbols package based on the BC version our app is targeting. When there is an installApps URL, use that app and not the nuget package. 

this should behave like this both when docker is used and when docker is skipped.  

### Steps to reproduce

sceenario 1) 

create an AL-Go repo (either appsource or PTE), add a new app to it. target BC24. skip docker (useCompilerFolder = true, doNotPublishApps = true)

add a dependency to a third party appsource app. the appsource app should have versions for both BC24 and BC25 available.

the build pipeline will fail since it will try to build your BC24 app with a dependency that targets BC25 and newer

scenario 2)

create an AL-Go repo (either appsource or PTE), add a new app to it. target BC24. skip docker (useCompilerFolder = true, doNotPublishApps = true)

add a dependency to a third party appsource app. the appsource app should have versions for both BC24 and BC25 available.

get a full app or runtime package of the third party app (not a symbols package) and add it to installApps. 

the build pipeline will fail since it will try to build your BC24 app with a dependency that targets BC25 and newer, and the installApps version is ignored. 

_No response_

freddydk:
Fixes #1330
Fixes #1268

This requires a new version of BcContainerHelper with https://github.com/microsoft/navcontainerhelper/pull/3832 for the bugs to be fixed. It will still work like before with the old version.

This PR makes sure that the installed apps, the platform and the country are transferred to NuGet resolution when locating apps for downloading - making sure that you don't download a package with dependencies to BC which is newer than what you are building for.

### DIFF
--- a/Actions/AL-Go-Helper.ps1
+++ b/Actions/AL-Go-Helper.ps1
@@ -18,7 +18,7 @@ $defaultCICDPushBranches = @( 'main', 'release/*', 'feature/*' )
 [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSUseDeclaredVarsMoreThanAssignments', 'defaultCICDPullRequestBranches', Justification = 'False positive.')]
 $defaultCICDPullRequestBranches = @( 'main' )
 $runningLocal = $local.IsPresent
-$defaultBcContainerHelperVersion = "preview" # Must be double quotes. Will be replaced by BcContainerHelperVersion if necessary in the deploy step - ex. "https://github.com/organization/navcontainerhelper/archive/refs/heads/branch.zip"
+$defaultBcContainerHelperVersion = "https://github.com/microsoft/navcontainerhelper/archive/refs/heads/freddydk/algoissue1330.zip" # Must be double quotes. Will be replaced by BcContainerHelperVersion if necessary in the deploy step - ex. "https://github.com/organization/navcontainerhelper/archive/refs/heads/branch.zip"
 $notSecretProperties = @("Scopes","TenantId","BlobName","ContainerName","StorageAccountName","ServerUrl","ppUserName","GitHubAppClientId")
 
 $runAlPipelineOverrides = @(
--- a/Actions/AL-Go-Helper.ps1
+++ b/Actions/AL-Go-Helper.ps1
@@ -41,6 +41,7 @@ $runAlPipelineOverrides = @(
 )
 
 # Well known AppIds
+$platformAppId = "8874ed3a-0643-4247-9ced-7a7002f7135d"
 $systemAppId = "63ca2fa4-4f03-4f2b-a480-172fef340d3f"
 $baseAppId = "437dbf0e-84ff-417a-965d-ed2bb9650972"
 $applicationAppId = "c1335042-3002-4257-bf8a-75c898ccb1b8"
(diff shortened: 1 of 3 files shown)


Actions/AL-Go-Helper.ps1
```diff
@@ -18,7 +18,7 @@ $defaultCICDPushBranches = @( 'main', 'release/*', 'feature/*' )
 [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSUseDeclaredVarsMoreThanAssignments', 'defaultCICDPullRequestBranches', Justification = 'False positive.')]
 $defaultCICDPullRequestBranches = @( 'main' )
 $runningLocal = $local.IsPresent
-$defaultBcContainerHelperVersion = "preview" # Must be double quotes. Will be replaced by BcContainerHelperVersion if necessary in the deploy step - ex. "https://github.com/organization/navcontainerhelper/archive/refs/heads/branch.zip"
+$defaultBcContainerHelperVersion = "https://github.com/microsoft/navcontainerhelper/archive/refs/heads/freddydk/algoissue1330.zip" # Must be double quotes. Will be replaced by BcContainerHelperVersion if necessary in the deploy step - ex. "https://github.com/organization/navcontainerhelper/archive/refs/heads/branch.zip"
```
aholstrup1, line 21: To be reverted

aholstrup1 (approved):

Issue #1449, opened 2025-02-12
[Bug]: Since last week we get this error when running on self-hosted runner `Error Message: Could not find a part of the path 'C:\..\..\.nuget\packages\Microsoft.ApplicationInsights\2.20.0\package\services\metadata\core-properties'`


nemanja899:
### AL-Go version

6.2

### Describe the issue

Full Error message 
[...]

PR #1451, opened 2025-02-12
Use temp folder for App Insights DLLs

aholstrup1:
Related to #1449


freddydk:
Didn't know about RUNNER_TEMP - looks like a good approach for temp folders in github workflows.

mazhelez (approved):

PR #1462, opened 2025-02-19
Fix link in RELEASENOTES.md

mazhelez:
Link is currently broken. E.g. https://github.com/microsoft/BCApps/pull/2995

image


freddydk (dismissed):

aholstrup1 (approved):

.github/workflows/CI.yaml
```diff
@@ -44,3 +44,19 @@ jobs:
             Write-Host "::Error::Error when running tests. The Error was $($_.Exception.Message)"
             $host.SetShouldExit(1)
           }
+
+      - name: Test aka.ms links in markdown files
```
freddydk, line 48: Why doesn't the AkaMSLinks.Test.ps1 not just follow all the other tests?
Why does it have to have its own step?

freddydk: I do think it is a detour if we start to have many steps like this in the CI.yaml.
We can easily parallelize the tests by adding that in tests/runtests.ps1 - but there has been no reason to do that.
Just hurts my eye to have test steps in CI.yaml which are 95% code copy of the step above - and using a lot of PowerShell in yaml, which we kind of agreed not to have.

My $0.02 - for what it's worth.

I will approve and leave you to do what you think is right.

freddydk (dismissed):

.github/workflows/CI.yaml
mazhelez: Thanks for the feedback. Made runtests.ps1 more generic so that it can reused in the other steps.

freddydk (approved):

aholstrup1 (approved):

PR #1473, opened 2025-02-26
runs-on type is ambiguous

### LINKED ISSUE #1182: Runs-on setting type is ambiguous - string or array
### Question

After upgrade AL-GO to 5.3 we do received these warnings:

`The property runs-on in DeployToPROD is expected to be of type Object[]`
image

Is it a code issue on your end or do we have to change something in our configs? My config looks like this:

```
"DeployToQA": {
    "EnvironmentType": "SaaS",
    "EnvironmentName": "QA",
    "Projects": "SRS.nl",
    "Branches": [
      "main"
    ],
    "ContinuousDeployment": true,
    "runs-on": "ubuntu-latest"
  },
```
Deployments works as expected

Best Regards, 
Gintautas

freddydk:
Today, the runs-on type in settings must be a string (can be a comma separated number of runner tags) and the runs-on property under environment needs to be an array of strings = runner tags.

This PR makes the runs-on under Environments become a string as well, but support arrays, to not break old settings files.

Fixes #1182
FIxes https://www.yammer.com/dynamicsnavdev/threads/3172600716926976

This means that this setting:

image

Yields identical os settings in environment matrix json:

```
EnvironmentsMatrixJson={"matrix":{"include":[{"environment":"TEST","os":"[\"windows-latest\",\"windows-2022\"]","shell":"powershell"},{"environment":"TEST2","os":"[\"windows-latest\",\"windows-2022\"]","shell":"powershell"}]},"fail-fast":false}
```

The PR also ensures that shell becomes pwsh if runs-on is `*ubuntu-*`

### DIFF
--- a/Actions/DetermineDeploymentEnvironments/DetermineDeploymentEnvironments.ps1
+++ b/Actions/DetermineDeploymentEnvironments/DetermineDeploymentEnvironments.ps1
@@ -268,7 +268,7 @@ else {
 $json = @{"matrix" = @{ "include" = @() }; "fail-fast" = $false }
 $deploymentEnvironments.Keys | Sort-Object | ForEach-Object {
     $deploymentEnvironment = $deploymentEnvironments."$_"
-    $json.matrix.include += @{ "environment" = $_; "os" = "$(ConvertTo-Json -InputObject @($deploymentEnvironment."runs-on".Split(',')) -compress)"; "shell" = $deploymentEnvironment."shell" }
+    $json.matrix.include += @{ "environment" = $_; "os" = "$(ConvertTo-Json -InputObject @($deploymentEnvironment."runs-on".Split(',').Trim()) -compress)"; "shell" = $deploymentEnvironment."shell" }
 }
 $environmentsMatrixJson = $json | ConvertTo-Json -Depth 99 -compress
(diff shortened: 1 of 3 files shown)


Actions/DetermineDeploymentEnvironments/DetermineDeploymentEnvironments.ps1
```diff
@@ -173,7 +173,14 @@ else {
             foreach($key in $keys) {
                 if ($deploymentSettings.ContainsKey($key)) {
                     if ($null -ne $deploymentSettings."$key" -and $null -ne $deployTo."$key" -and $deploymentSettings."$key".GetType().Name -ne $deployTo."$key".GetType().Name) {
-                        Write-Host "::WARNING::The property $key in $settingsName is expected to be of type $($deploymentSettings."$key".GetType().Name)"
+                        if ($key -eq "runs-on" -and $deployTo."$key" -is [Object[]]) {
+                            # Support setting runs-on as an array in settings to not break old settings
+                            # See https://github.com/microsoft/AL-Go/issues/1182
+                            $deployTo."$key" = $deployTo."$key" -join ','
```
mazhelez, line 179: What about SETTINGS.md?

freddydk: In settings.md it isn't described as a specific type and people tend to use a string (like the original runs-on)

aholstrup1 (approved):

PR #1477, opened 2025-02-27
fix internal template problem

### LINKED ISSUE #1475: [Bug]: Step "Check for updates to AL-Go system files" fails when having internal AL-Go repositories
### AL-Go version

preview

### Describe the issue

As I mentioned in the Office Hour call yesterday, I'm struggling with AL-Go when using in an organization that do not allow public repos.

The step "Check for updates to AL-Go system files" fails with a 404 error in the CICD workflow (and others).

### Expected behavior

All workflows should work, even if all AL-Go repositories are internal.

### Steps to reproduce

- Fork AL-Go to a repo in our org, set as internal
- Run the deploy workflow to create Actions and templates repositories, all set to internal
  - Set action access to the AL-Go and AL-Go-Actions repos, so that the actions can be access by other repos in the org
- Create a new repository from the PTE template
- Run the "Update AL-Go System Files" in the new repo
- Now the CICD workflow got started by the push from the "Update AL-Go System Files" workflow, and the step "Check for updates to AL-Go system files" fails with a 404 error when trying to get the latest SHA.

- This is probably because we are not using the token parameter of the `CheckForUpdates` action in other workflows than UpdateGitHubGoSystemFiles.yaml.
- The `github.token` seems to not have enough access to our internal template repo. It is configured the Workflow Permissions with "Read repository contents and packages permissions". (Also tested with "Read and write permissions", but didn't help)
- I tested to add `token: ${{ fromJson(steps.ReadSecrets.outputs.Secrets).ghTokenWorkflow }}` as a parameter (and a `ReadSecrets` step that fetches `ghTokenWorkflow`), but that token failed with 401 error. Probably due to that the `GetAccessToken` function is only called when `update` = 'Y'.
- The same applies to other workflows that uses `CheckForUpdates` without a token, as the `CreateRelease` workflow.

Oh, yes - I'm using GitHub App authentication, but I'm not sure if this matter in this scenario.

Just let me know if you need more information or ping me on Teams if you want to chat and see what we've got.

PS.
We are evaluating AL-Go, with the goal to migrate all our repos to GitHub and abandon our custom pipelines on DevOps. But first we are trying to set everything up as we want it to work, with security, policies etc.

freddydk:
Fixes #1475

By making sure that a GhTokenWorkflow is used (either as PAT or as App Auth) when reading the template.
If the Token doesn't have access to the remove tepo - it will retry with anonymous

### DIFF
--- a/Actions/CheckForUpdates/CheckForUpdates.ps1
+++ b/Actions/CheckForUpdates/CheckForUpdates.ps1
@@ -38,7 +38,10 @@ if ($update -eq 'Y') {
     if (-not $token) {
         throw "The GhTokenWorkflow secret is needed. Read https://github.com/microsoft/AL-Go/blob/main/Scenarios/GhTokenWorkflow.md for more information."
     }
+}
 
+$readToken = $token
+if ($token) {
     # token comes from a secret, base 64 encoded
     $token = [System.Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String($token))
 
@@ -48,10 +51,6 @@ if ($update -eq 'Y') {
         $repositories += $templateUrl.Split('@')[0]
     }
     $readToken = GetAccessToken -token $token -permissions @{"actions"="read";"contents"="read";"metadata"="read"} -repositories $repositories
-}
-else {
-    # use token directly
-    $readToken = $token
 }
 
 # Use Authenticated API request if possible to avoid the 60 API calls per hour limit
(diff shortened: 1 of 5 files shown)


jwikman (approved):
Looks good to me.
Confirmed to work with PAT and GitHub App authentication with internal repositories.

freddydk:
I am also running the full end 2 end test suite on this to see whether this breaks anything - doesn't look like it

freddydk:
End 2 end run: https://github.com/freddydk/AL-Go/actions/runs/13570931970

mazhelez (approved):

aholstrup1 (approved):

PR #1481, opened 2025-03-03
Downloaded apps no longer published in PR builds

spetersenms:
As part of the incremental builds feature, we download any apps that have not been updated. In single project repos, these downloaded apps are incorrectly published as artifacts in PR builds.

This PR adds a small snippet to remove any downloaded apps after the build, to ensure only compiled apps are published.

### DIFF
--- a/Actions/RunPipeline/RunPipeline.ps1
+++ b/Actions/RunPipeline/RunPipeline.ps1
@@ -141,7 +141,7 @@ try {
         }
         if (!$buildAll) {
             Write-Host "Get unmodified apps from baseline workflow run"
-            $downloadAppFolders, $downloadTestFolders, $downloadBcptTestFolders = Get-UnmodifiedAppsFromBaselineWorkflowRun `
+            $downloadedAppsByType = Get-UnmodifiedAppsFromBaselineWorkflowRun `
                 -token $token `
                 -settings $settings `
                 -baseFolder $baseFolder `
@@ -508,12 +508,16 @@ try {
         -appBuild $appBuild -appRevision $appRevision `
         -uninstallRemovedApps
 
-    $downloadAppFolders, $downloadTestFolders, $downloadBcptTestFolders | ForEach-Object {
-        Write-Host "Debug: removing apps from $_"
-        foreach($downloadedApp in $_) {
-            if (Test-Path $downloadedApp) {
-                Write-Host "Debug: Removing downloaded app $downloadedApp"
-                Remove-Item $downloadedApp
+    $downloadedAppsByType | ForEach-Object {
+        $type = $_.type
+        $mask = $_.mask
+        $thisArtifactFolder = Join-Path $buildArtifactFolder $mask
+        Write-Host "Debug: removing $type apps with mask: $mask from $thisArtifactFolder"
+        foreach($downloadedApp in $_.downloadedApps) {
+            $thisApp = Join-Path $thisArtifactFolder $downloadedApp
+            Write-Host "Debug: app: $thisApp"
+            if (Test-Path $thisApp) {
+                Remove-Item $thisApp
             }
         }
     }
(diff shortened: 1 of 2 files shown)


Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -507,6 +508,22 @@ try {
         -appBuild $appBuild -appRevision $appRevision `
         -uninstallRemovedApps
 
+    # If any apps were downloaded as part of incremental builds, we should remove them again after the build
```
freddydk, line 511: The logic is fine, but this should only happen if this is a PR build - if this is a normal build we will publish downloaded apps and downloaded projects on the build.
You could check ENV:GITHUB_EVENT_NAME -like 'pull_request*' before running this section.

freddydk (changes requested):

Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -140,7 +141,7 @@ try {
         }
         if (!$buildAll) {
             Write-Host "Get unmodified apps from baseline workflow run"
-            Get-UnmodifiedAppsFromBaselineWorkflowRun `
+            $downloadedAppsByType = Get-UnmodifiedAppsFromBaselineWorkflowRun `
```
mazhelez, line 144: Can you also please leave a comment with your findings: that because the app files are downloaded in the build artifacts folders, the apps won't be built when calling `Run-AlPipeline`?

Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -507,6 +508,28 @@ try {
         -appBuild $appBuild -appRevision $appRevision `
         -uninstallRemovedApps
 
+    # If any apps were downloaded as part of incremental builds in a pr, we should remove them again after the build
+    try {
```
mazhelez, line 512: Minor: Calling `Remove-Item` with `-ErrorAction Ignore` will let you continue trying to delete files.

Alternatively, the try-catch could be narrower.

mazhelez (approved):

Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -507,6 +508,28 @@ try {
...
+                        $thisApp = Join-Path $thisArtifactFolder $downloadedApp
+                        Write-Host "Removing: $thisApp"
+                        if (Test-Path $thisApp) {
+                            Remove-Item $thisApp
+                        }
+                    }
+                }
+            }
+        }
+    } catch {
```
freddydk, line 529: Please move the try/catch to the Remove-Item and explain which app couldn't be deleted. This will help in troubleshooting scenarios.

freddydk (changes requested):

Actions/RunPipeline/RunPipeline.ps1
spetersenms: Going with the narrower try-catch so we can write better warnings to the logs.

aholstrup1 (dismissed):

mazhelez (approved):

Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -507,6 +509,28 @@ try {
...
+                Write-Host "Removing downloaded apps from $thisArtifactFolder"
+                foreach($downloadedApp in $_.downloadedApps) {
+                    $thisApp = Join-Path $thisArtifactFolder $downloadedApp
+                    try {
+                        if (Test-Path $thisApp) {
+                            Remove-Item $thisApp
+                        }
+                        Write-Host "Removed app: $thisApp"
+                    } catch {
+                        Write-Host "::warning::Failed to remove app: $thisApp"
```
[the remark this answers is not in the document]
freddydk: I agree - please make the warning a normal write-host instead.
The user won't be able to do anything about it and it doesn't really matter if it fails.
We can use the log to answer questions from partners who wonder

aholstrup1 (approved):

freddydk (approved):

mazhelez (approved):

PR #1488, opened 2025-03-07
Fix incremental build problems

### LINKED ISSUE #1487: [Question]: Setting incrementalBuilds not working as expected
### Question

Hi,

I'm working with the preview version of AL-Go and added the setting incrementalBuilds in my `\AL-Go\settings.json`.

```json
{
  "incrementalBuilds": {
    "onPush": true,
    "onSchedule": true,
    "mode": "modifiedApps"
  }
}
```

I expected that this setting would just copy the already built apps from a previous CI/CD run, but instead, it compiles every app again.

```log
Compiling...
  .\alc.exe /project:"c:\sources\Assembly Lines" ...
  Microsoft (R) AL Compiler version 14.3.20.17127
  Copyright (C) Microsoft Corporation. All rights reserved
  Compilation started for project 'Assembly Lines' containing '1' files at '17:23:05.069'.
  Compilation ended at '17:23:12.741'.
```

And this for the 13 apps in my repository. After this, every app needs to be redeployed to the UAT of the customer. This process takes a long time to run the CI/CD.

Image

**Question**: Am I missing a setting or did I misinterpret the new feature?

freddydk:
It can be hard to discover the reason why a full build was triggered deep inside the log.
This PR adds a notice to the build when a full build is triggered, with the reason behind it.

Fixes #1487

Additionally, a Pull Request, which should trigger a full build (due to files changed) wouldn't trigger the change due to a parameter error.

Thirdly - in some cases, we would attempt to do a git diff with an empty sha.

Lastly - skip the build cleanup step if no build step was ever run (saves ~10 seconds)

### DIFF
--- a/Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.Action.ps1
+++ b/Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.Action.ps1
@@ -35,7 +35,7 @@ if(-not $buildAllProjects) {
 
     Write-Host "::group::Get Modified Files"
     try {
-        $modifiedFiles = @(Get-ModifiedFiles -baselineSHA $baselineWorkflowSHA)
+        $buildAllProjects, $modifiedFiles = Get-ModifiedFiles -baselineSHA $baselineWorkflowSHA
         OutputMessageAndArray -message "Modified files" -arrayOfStrings $modifiedFiles
     }
     catch {
--- a/Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.Action.ps1
+++ b/Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.Action.ps1
@@ -21,37 +21,39 @@ if ($ENV:GITHUB_EVENT_NAME -eq 'pull_request') {
     $targetBranch = $env:GITHUB_BASE_REF
 }
 
+#region Action: Determine projects to build
 Write-Host "$($ENV:GITHUB_EVENT_NAME) on $targetBranch"
 $buildAllProjects, $publishSkippedProjects = Get-BuildAllProjectsBasedOnEventAndSettings -ghEventName $ENV:GITHUB_EVENT_NAME -settings $settings
 
+$modifiedFiles = @()
 $baselineWorkflowRunId = 0 #default to 0, which means no baseline workflow run ID is set
 $baselineWorkflowSHA = ''
 if(-not $buildAllProjects) {
     Write-Host "::group::Determine Baseline Workflow ID"
     $baselineWorkflowRunId,$baselineWorkflowSHA = FindLatestSuccessfulCICDRun -repository $env:GITHUB_REPOSITORY -branch $targetBranch -token $token -retention $settings.incrementalBuilds.retentionDays
     Write-Host "::endgroup::"
+
+    Write-Host "::group::Get Modified Files"
+    try {
+        $modifiedFiles = @(Get-ModifiedFiles -baselineSHA $baselineWorkflowSHA)
+        OutputMessageAndArray -message "Modified files" -arrayOfStrings $modifiedFiles
+    }
+    catch {
+        OutputNotice -message "Failed to calculate modified files since $baselineWorkflowSHA, building all projects"
+        $buildAllProjects = $true
+    }
+    Write-Host "::endgroup::"
 }
 
-#region Action: Determine projects to build
-Write-Host "::group::Get Modified Files"
-try {
-    $modifiedFiles = @(Get-ModifiedFiles -baselineSHA $baselineWorkflowSHA)
-    OutputMessageAndArray -message "Modified files" -arrayOfStrings $modifiedFiles
+if (-not $buildAllProjects) {
+    Write-Host "::group::Determine Incremental Build"
+    $buildAllProjects = Get-BuildAllProjects -modifiedFiles $modifiedFiles -baseFolder $baseFolder
+    Write-Host "::endgroup::"
 }
-catch {
-    OutputNotice -message "Failed to calculate modified files since $baselineWorkflowSHA, building all projects"
-    $buildAllProjects = $true
-    $modifiedFiles = @()
-}
-Write-Host "::endgroup::"
-
-Write-Host "::group::Determine Incremental Build"
-$buildAllProjects = $buildAllProjects -or (Get-BuildAllProjects -modifiedFiles $modifiedFiles -baseFolder $baseFolder)
-Write-Host "::endgroup::"
 
 # If we are to publish artifacts for skipped projects later, we include the full project list and in the build step, just avoid building the skipped projects
 Write-Host "::group::Get Projects To Build"
-$allProjects, $modifiedProjects, $projectsToBuild, $projectDependencies, $buildOrder = Get-ProjectsToBuild -baseFolder $baseFolder -buildAllProjects $publishSkippedProjects -modifiedFiles $modifiedFiles -maxBuildDepth $maxBuildDepth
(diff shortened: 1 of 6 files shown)


Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -135,7 +135,7 @@ try {
             $buildAll = Get-BuildAllApps -baseFolder $baseFolder -project $project -modifiedFiles $modifiedFiles
         }
         catch {
-            OutputWarning -message "Failed to calculate modified files since $baselineWorkflowSHA, building all apps"
+            OutputNotice -message "Failed to calculate modified files since $baselineWorkflowSHA, building all apps"
```
mazhelez, line 138: This should stay a warning. 

Maybe even an error?
At this point modified files should be known, so there should be no reason for failure (either an instability or a bug). So, one might want to re-run the build instead of building all apps.

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.psm1
```diff
@@ -298,7 +298,7 @@ function Get-BuildAllProjects {
     $settings = $env:Settings | ConvertFrom-Json
 
     if (!$modifiedFiles) {
-        Write-Host "No files modified, building everything"
+        Write-Host "::notice::No files modified, building everything"
```
mazhelez, line 301: Why, if there are no modified files, everything is being built?

I recall with the previous design, no modified files meant CICD is running (not PR Build).

With the new design, wouldn't it be a problem? E.g. there's a CICD that runs on a schedule with incremental builds: if there have been no changes since yesterday, suddenly everything will be built.

Actions/RunPipeline/RunPipeline.ps1
freddydk: The primary reason for this situation is a. A rebase or b. Too many modified files. In both cases what you want is not an error or a warning, instead you want a full rebuild

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.psm1
freddydk: Even if you haven't modified files, bc version might be changed or your dependencies. This is by design and should remain so imo.

Actions/RunPipeline/RunPipeline.ps1
mazhelez: But this should all be determined during _Initialization_ job, not here.

I also don't see how `Get-ModifiedFiles` handles rebases or too many files. 🤔

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.psm1
mazhelez: So, if CICD runs on a schedule:
If one day you change a single app, then only that app is built.
But if the next day you change no apps, then all of them are built?

That cannot be the design...

freddydk: I didn't change the behavior in this PR - I only change the notifications to write why it selects to do a full build due to a partner not being able to understand why it did a full build.

If CI/CD runs on a schedule - it will (by default) build all apps every time (incremental builds by default is off on a schedule). I get your point if you have selected to add incremental builds on a schedule - will make a fix for this.

Actions/RunPipeline/RunPipeline.ps1
freddydk: This PR doesn't change the logic, it just changes the logs.

Get-ModifiedFiles doesn't handle rebases or too many files, but...
If you are trying to do a git diff with a SHA, which isn't in the history, it will fail.
If you are trying to do a git diff with a changelist which exceeds memory usage, it will fail.
In these cases - a full build is better than an error, which the user doesn't have a clue what to do about.

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.Action.ps1
```diff
@@ -21,37 +21,41 @@ if ($ENV:GITHUB_EVENT_NAME -eq 'pull_request') {
...
-    Write-Host "Failed to calculate modified files, building all projects"
-    $buildAllProjects = $true
-    $modifiedFiles = @()
+    Write-Host "::group::Get Modified Files"
+    try {
+        $buildAllProjects, $modifiedFiles = Get-ModifiedFiles -baselineSHA $baselineWorkflowSHA
+        OutputMessageAndArray -message "Modified files" -arrayOfStrings $modifiedFiles
+    }
+    catch {
+        OutputNotice -message "Failed to calculate modified files since $baselineWorkflowSHA, building all projects"
```
aholstrup1, line 42: Can we write the actual error message in the log as well? Would make it easier to debug why we couldn't calculate modified files

aholstrup1 (dismissed):

Actions/RunPipeline/RunPipeline.ps1
mazhelez: > a full build is better than an error, which the user doesn't have a clue what to do about.

I'm not sure about that. A full build due to an instability might have unwanted consequences. The user might prefer to simply re-run the build.

Anyway, let's take this discussion offline.

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.Action.ps1
mazhelez: Also, I still believe a warning is better suited here.

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.psm1
```diff
@@ -26,15 +26,18 @@ function Get-ModifiedFiles {
...
         }
-        else {
+        elseif ($baselineSHA) {
             $headSHA = git rev-parse HEAD
             Write-Host "Current HEAD is $headSHA"
             Invoke-CommandWithRetry -ScriptBlock { RunAndCheck git fetch origin $baselineSHA | Out-Host }
             Write-Host "Not a pull request, using baseline SHA $baselineSHA and current HEAD $headSHA"
         }
+        else {
+            return $true, @()
```
mazhelez, line 36: When is it in this case? `$baselineSHA` is a mandatory parameter.

Also, doc should be updated.

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.Action.ps1
aholstrup1: > Also, I still believe a warning is better suited here.

Agree 👍

Tests/DetermineProjectsToBuild.Test.ps1
```diff
@@ -762,13 +762,13 @@ Describe "Get-BuildAllProjects" {
         $baseFolder = (New-Item -ItemType Directory -Path (Join-Path $([System.IO.Path]::GetTempPath()) $([System.IO.Path]::GetRandomFileName()))).FullName
     }
 
-    It ('returns true if there are no modified filed') {
+    It ('returns false if there are no modified filed') {
```
mazhelez, line 765:
```suggestion
    It ('returns false if there are no modified files') {
```

Typo

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.psm1
```diff
@@ -333,7 +360,7 @@ function Get-BuildAllProjects {
 .Description
     Determines whether a full build is required.
```
mazhelez, line 361:
```suggestion
    Determines whether all apps in a project should be built.
```

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.psm1
freddydk: Removed this. It came from DetermineProjectsToBuild.Action.ps1, where the function was called even if no baseline SHA was found. Probably this lead to the error you have seen - now I added a notice and do not call get-modifiedFiles

aholstrup1 (approved):

mazhelez (approved):

PR #1499, opened 2025-03-11
FIx issue with removal of pre-built apps.

spetersenms:
$downloadedAppsByType was incorrectly initialized as an object instead of a list.
This failed because checking if the object exist in an if statement, will return true for an empty object, but false for a list.

### DIFF
--- a/Actions/RunPipeline/RunPipeline.ps1
+++ b/Actions/RunPipeline/RunPipeline.ps1
@@ -127,7 +127,7 @@ try {
     $buildArtifactFolder = Join-Path $projectPath ".buildartifacts"
     New-Item $buildArtifactFolder -ItemType Directory | Out-Null
 
-    $downloadedAppsByType = @{}
+    $downloadedAppsByType = @()
     if ($baselineWorkflowSHA -and $baselineWorkflowRunId -ne '0' -and $settings.incrementalBuilds.mode -eq 'modifiedApps') {
         # Incremental builds are enabled and we are only building modified apps
         try {


Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -127,7 +127,7 @@ try {
     $buildArtifactFolder = Join-Path $projectPath ".buildartifacts"
     New-Item $buildArtifactFolder -ItemType Directory | Out-Null
 
-    $downloadedAppsByType = @{}
+    $downloadedAppsByType = @()
```
aholstrup1, line 130: We saw failures on line 515. Don't we need to handle that line differently as well? 
https://github.com/microsoft/AL-Go/blob/d5c69121c87019232b4aa31a92827a16a3986471/Actions/RunPipeline/RunPipeline.ps1#L515

Something like:
`$_.PSObject.Properties.Name -eq "downloadedApps"`

I assumed the error was due to strictmode in powershell. But I could be wrong

aholstrup1 (approved):

Actions/RunPipeline/RunPipeline.ps1
spetersenms: We can do the additional check, but the reason it happed is because the if statement on 513 succeeds,
An empty object will return true in an if-statement `if (@{}) == true`, however an empty list wont `if (@()) == false` .

`ForEach-Object` will also run once when inputting an empty object, but not when inputting an empty list.

So we can still check if the property exist on line 515, but since I changed the type to list, it will now only have actual content if `Get-UnmodifiedAppsFromBaselineWorkflowRun` on line 145 runs.

mazhelez (approved):

PR #1503, opened 2025-03-12
fix: Use ghTokenWorkflow for CheckForUpdates in CreateRelease.yaml

### LINKED ISSUE #1475: [Bug]: Step "Check for updates to AL-Go system files" fails when having internal AL-Go repositories

jwikman:
Fixes #1475 (hopefully completely this time)

My apologies Freddy, but I think I messed up when I tested #1477...

CI/CD worked just as intended, but "Create Release" did not work. It ended up with a 404 error. It seemed as if the token sent to the CreateRelase action was empty:
```
    _actor: jwikman
    _token:                                            <----------
    _templateUrl: https://github.com/[redacted]/AL-Go-PTE@preview
    _downloadLatest: true
    _update: N
    _updateBranch: main
    _directCommit: false
```

After the proposed change, the token is passed as expected
```
    _actor: jwikman
    _token: ***                                     <---------------------
    _templateUrl: https://github.com/[redacted]/AL-Go-PTE@preview
    _downloadLatest: true
    _update: N
    _updateBranch: main
    _directCommit: false
```

I do not fully follow the logic with the `TokenForPush` secret, but that token did not seem to work in this context?

### DIFF
--- a/Templates/AppSource App/.github/workflows/CreateRelease.yaml
+++ b/Templates/AppSource App/.github/workflows/CreateRelease.yaml
@@ -104,7 +104,7 @@ jobs:
         with:
           shell: powershell
           gitHubSecrets: ${{ toJson(secrets) }}
-          getSecrets: 'TokenForPush'
+          getSecrets: 'TokenForPush,ghTokenWorkflow'
           useGhTokenWorkflowForPush: '${{ github.event.inputs.useGhTokenWorkflow }}'
 
       - name: Determine Projects
@@ -118,7 +118,7 @@ jobs:
         with:
           shell: powershell
           templateUrl: ${{ env.templateUrl }}
-          token: ${{ fromJson(steps.ReadSecrets.outputs.Secrets).TokenForPush }}
+          token: ${{ fromJson(steps.ReadSecrets.outputs.Secrets).ghTokenWorkflow }}
           downloadLatest: true
 
       - name: Analyze Artifacts
--- a/Templates/Per Tenant Extension/.github/workflows/CreateRelease.yaml
+++ b/Templates/Per Tenant Extension/.github/workflows/CreateRelease.yaml
@@ -104,7 +104,7 @@ jobs:
         with:
           shell: powershell
           gitHubSecrets: ${{ toJson(secrets) }}
-          getSecrets: 'TokenForPush'
+          getSecrets: 'TokenForPush,ghTokenWorkflow'
           useGhTokenWorkflowForPush: '${{ github.event.inputs.useGhTokenWorkflow }}'
 
       - name: Determine Projects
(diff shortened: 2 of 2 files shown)


freddydk:
Not sure that this fix is correct - will check

jwikman:
> When you ran this - did you check this field:
> 
> 
> 
> image
> 
> 
> 
> or did you leave it empty?

I ticked that one.

Just let me know if I can do anything, like test other approaches (and test correctly this time 🫣).

jwikman:
Btw, what's the reason why the "Use GhTokenWorkflow" input parameter isn't replaced with a setting?
Either you have configured your repo with GhTokenWorkflow or not, I do not see when I want my colleagues to sometimes tick that and sometimes not. But I am probably missing something here...? 😅

freddydk:
The useGhTokenWorkflow should be a setting instead.
Could you send me the log of the create release workflow run, where the token ended up empty?

freddydk:
After looking at the logs, discussing this with @jwikman, I think the fix is the right one.
This is also the way we do in CI/CD - and there is no reason to re-use the TokenForPush for a purpose it wasn't intended for.
TokenForPush is a token for push - not a token for reading the template repository and hence - it doesn't have the necessary permissions for the template repo.

freddydk (approved):

jonasblunck (approved):

PR #1515, opened 2025-03-18
Add doc for PreCompileApp and PostCompileApp

mazhelez:
Add doc for `PreCompileApp` and `PostCompileApp` scripts.


freddydk (approved):
Way more documentation than in the other overrides - maybe we should add the same for the others?

aholstrup1 (approved):

PR #1545, opened 2025-03-26
Support using an online environment for testing when building apps

freddydk:
This PR adds support for using an online environment for running tests when building apps. If you are building a PTE with dependencies to 3rd party appsource apps, you might not have access to runtime packages for these AppSource apps and hence, you might not be able to run tests for your PTE.

Using an online environment for CI/CD or potentially a scheduled Test Current workflow, allows you to run tests for your PTE.

### New Workflow specific settings

- `buildEnvironmentName` - should be the Business Central environment name (from the Admin Center) for the environment you want to use if you want to run tests using online environments. Note that you also need to create a buildAuthContext secret, which gives access to this environment.

### Use online environment during build for running tests

By creating a buildAuthContext secret and setting the buildEnvironmentName for a specific branch and a specific workflow, will cause AL-Go for GitHub to use this environment for running tests. An example of how to define these settings could be:

```json
  "useCompilerFolder": true,
  "doNotPublishApps": true,
  "conditionalSettings": [
    {
      "workflows": [ "CICD" ],
      "branches": [ "main" ],
      "settings": {
        "doNotPublishApps": false,
        "buildEnvironmentName": "cicd",
        "buildAuthContextSecretName": "cicdAuthContext",
        "workflowConcurrency": [ "group: ${{ github.workflow }}-${{ github.ref }}", "cancel-in-progress: false" ]
      }
    }
  ]
```


RELEASENOTES.md
```diff
@@ -8,6 +8,30 @@
...
+
+- `buildEnvironmentName` - should be the Business Central environment name (from the Admin Center) for the environment you want to use if you want to run tests using online environments. Note that you also need to create a [buildAuthContext](https://aka.ms/algosecrets#buildAuthContext) secret, which gives access to this environment.
+
+### Use online environment during build for running tests
+
+By creating a [build authentication context](https://aka.ms/algosecrets#buildAuthContext) secret and setting the [buildEnvironmentName](https://aka.ms/algosettings#buildEnvironmentName) for a specific branch and a specific workflow, will cause AL-Go for GitHub to use this environment for running tests. An example of how to define these settings could be:
+
+```json
+  "useCompilerFolder": true,
+  "doNotPublishApps": true,
```
aholstrup1, line 21: Would it make sense to recommend using the workflowConcurrency setting in this example as well? I'm guessing it would fail if you have two CI/CDs running concurrently.

Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -101,6 +101,18 @@ try {
...
 
+    $authContext = $null
+    $environmentName = ""
+    if ($buildAuthContext -and $settings.buildEnvironmentName) {
+        $authContextHT = $buildAuthContext | ConvertFrom-Json | ConvertTo-HashTable
+        $environmentName = $settings.buildEnvironmentName
+        if ($environmentName -notlike 'https://*') {
+            $authContext = New-BcAuthContext @authContextHT
+            # When using online environment for builds, use CompilerFolder to avoid creating a container
+            $settings.useCompilerFolder = $true
```
spetersenms, line 112: Should we write a log about this? Might be confusing for users if they expect to use a container.

Actions/AL-Go-Helper.ps1
```diff
@@ -18,7 +18,7 @@ $defaultCICDPushBranches = @( 'main', 'release/*', 'feature/*' )
 [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSUseDeclaredVarsMoreThanAssignments', 'defaultCICDPullRequestBranches', Justification = 'False positive.')]
 $defaultCICDPullRequestBranches = @( 'main' )
 $runningLocal = $local.IsPresent
-$defaultBcContainerHelperVersion = "preview" # Must be double quotes. Will be replaced by BcContainerHelperVersion if necessary in the deploy step - ex. "https://github.com/organization/navcontainerhelper/archive/refs/heads/branch.zip"
+$defaultBcContainerHelperVersion = "https://github.com/microsoft/navcontainerhelper/archive/refs/heads/freddydk/includeauthcontext.zip" # Must be double quotes. Will be replaced by BcContainerHelperVersion if necessary in the deploy step - ex. "https://github.com/organization/navcontainerhelper/archive/refs/heads/branch.zip"
```
aholstrup1, line 21: To be reverted

RELEASENOTES.md
```diff
@@ -30,7 +30,8 @@ By creating a [build authentication context](https://aka.ms/algosecrets#buildAut
       "settings": {
         "doNotPublishApps": false,
```
aholstrup1, line 31:
```suggestion
        "doNotPublishApps": false,
        "useCompilerFolder": true,        
```
Include useCompilerFolder?

Actions/ReadSettings/ReadSettings.ps1
```diff
@@ -54,6 +54,11 @@ if ($settings.versioningstrategy -ne -1) {
     }
 }
 
+if ($settings.buildEnvironmentName -and -not $settings.useCompilerFolder) {
+    Write-Host "Using an online environment for testing, use CompilerFolder to build the app"
```
aholstrup1, line 58: Suggestion; Make it a warning? This indicates to me that they repo owner has misconfigured something in their AL-Go settings. We catch it so this workflow doesn't fail but they should fix their settings.

freddydk: Actually, the idea here is that people shouldn't care - if they specify that they want to use an online environment for testing, then we automatically find out that using compilerfolder is the best option.

RELEASENOTES.md
freddydk: This sample is different - the global settings are using compilerfolder and have set doNotPublishApps to true, because they might have dependencies, which cannot be resolved. When setting the buildAuthContext, you then set doNotPublishApps to false - meaning: Yes, publish the apps when using an online environment.

Actions/AL-Go-Helper.ps1
freddydk: yes

RELEASENOTES.md
freddydk: Added a little extra text to the description

Actions/ReadSettings/ReadSettings.ps1
```diff
@@ -54,6 +54,11 @@ if ($settings.versioningstrategy -ne -1) {
...
+    $settings.useCompilerFolder = $true
```
mazhelez, line 59: Why defaulting to compiler folder? The BC online environment is used for running the tests, right? Compiling the apps can still happen in a BC container.

freddydk: Why would you?
Containers are much slower than CompilerFolders and since you are running tests on an online environment, you will not need the container ever.
Also - if you switch to use an online environment because of dependencies or functionalities, which cannot be tested on a container - it will fail if you try to.
I cannot see any reason for building using a container and running tests in an online environment.

mazhelez: For one, there could be overrides that do some additional setup before building the apps and test apps.

Switching like that, under the hood, is a breaking change. It must be documented at least.

freddydk: IMO - compiling with containers when using an online environment for testing will at best just be 4-5 times slower and at worst will fail slowly. But I agree that this should be more clearly documented under the setting - when you set this - you also do...

BTW - In the future (after getting rid of BcContainerHelper) - I don't think we should support building in containers at all.

freddydk:
This will be done in a different way

PR #1553, opened 2025-03-28
Support deploying AL-Go without OrgPAT

aholstrup1:
Rather than using a PAT to deploy AL-Go we should support deploying using a GitHub app.

### DIFF
--- a/Internal/Deploy.ps1
+++ b/Internal/Deploy.ps1
@@ -2,8 +2,6 @@
 Param(
     [Parameter(Mandatory=$true)]
     [Hashtable] $config,
-    [Parameter(Mandatory=$true)]
-    [string] $token,
     [Parameter(Mandatory=$false)]
     [bool] $directCommit
 )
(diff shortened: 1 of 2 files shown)


mazhelez (approved):

freddydk (approved):

freddydk:
We should probably see how far we can get on avoiding a PAT for end 2 end testing as well

Internal/Deploy.ps1
```diff
@@ -162,12 +150,13 @@ try {
...
 
         try {
-            $serverUrl = "https://$($config.githubOwner):$token@github.com/$($config.githubOwner)/$repo.git"
+            $serverUrl = "https://$($config.githubOwner)@github.com/$($config.githubOwner)/$repo.git"
             if (Test-Path $repo) {
                 Remove-Item $repo -Recurse -Force
             }
             invoke-git clone --quiet $serverUrl
             Set-Location $repo
+            invoke-gh auth setup-git
```
jwikman, line 159: @aholstrup1 I tested this out today, to get rid of my OrgPAT.

But I believe that `invoke-gh auth setup-git` should be moved a couple of lines up, above `invoke-git clone`.
One can even argue for having this line above the try block, since we need this authentication both in the try and catch block :)

PR #1639, opened 2025-04-07
Create $artifactsFolder only if it doesn't exist

### LINKED ISSUE #1630: [Bug]: Path to artifacts already exists error during CICD run of first Release, from Release branch
### AL-Go version

6.4

### Describe the issue

Hi all,
I keep getting `Error trying to locate previous release. Error was An item with the specified name D:\a\d365-advanced-landed-cost\d365-advanced-landed-cost\artifacts already exists.` error when i try to run PR build or CICD on a release branch (very first one in the repo) that i created from main branch.

Same PR build or any given CICD run on main works without issues.
I am not sure if its a bug or if im doing something horribly wrong but i cannot understand. Any help is appreciated!

### Expected behavior

Build run to work on release branch.

### Steps to reproduce

Create very first release with Create Release workflow, use MAIN branch, give it whatever name/tag desired.
Create Release Branch must be selected. Release being Prerelease or Release doesn't change the outcome, tried both.
Try to run a CICD manually on release branch, or create a PR to release branch to trigger a PR build flow run.

[logs_36486275053.zip](https://github.com/user-attachments/files/19550616/logs_36486275053.zip)

mazhelez:
When downloading the previous release, the folder is always created.

However, in the case of issue #1630 the path already exists.

Fixes #1630

### DIFF
--- a/Actions/RunPipeline/RunPipeline.ps1
+++ b/Actions/RunPipeline/RunPipeline.ps1
@@ -251,7 +251,9 @@ try {
             if ($latestRelease) {
                 Write-Host "Using $($latestRelease.name) (tag $($latestRelease.tag_name)) as previous release"
                 $artifactsFolder = Join-Path $baseFolder "artifacts"
-                New-Item $artifactsFolder -ItemType Directory | Out-Null
+                if(-not (Test-Path $artifactsFolder)) {
+                    New-Item $artifactsFolder -ItemType Directory | Out-Null
+                }
                 DownloadRelease -token $token -projects $project -api_url $ENV:GITHUB_API_URL -repository $ENV:GITHUB_REPOSITORY -release $latestRelease -path $artifactsFolder -mask "Apps"
                 $previousApps += @(Get-ChildItem -Path $artifactsFolder | ForEach-Object { $_.FullName })
             }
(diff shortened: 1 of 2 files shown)


Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -251,7 +251,9 @@ try {
             if ($latestRelease) {
                 Write-Host "Using $($latestRelease.name) (tag $($latestRelease.tag_name)) as previous release"
                 $artifactsFolder = Join-Path $baseFolder "artifacts"
```
aholstrup1, line 253: Another option could be to create the artifacts folder in a temp folder (e.g. $ENV:RUNNER_TEMP). If we use an existing folder we don't really know what's in it. Idk if that could lead to other issues

aholstrup1 (approved):

Actions/RunPipeline/RunPipeline.ps1
mazhelez: I actually thought about it this scenario as well. Generally, having other apps there, shouldn't matter.
And I'm not sure if there'll be side effects of using a temp folder.

freddydk (approved):

PR #1652, opened 2025-04-11
Unknown dependencies only installed for test apps.

spetersenms:
Installation of unknown apps was originally added for the deploy test apps feature, however it was incorrectly made to find unknown dependencies for all apps, not just test apps. This fix changes the logic to only get unknown dependencies for test apps.

### DIFF
--- a/Actions/Deploy/Deploy.ps1
+++ b/Actions/Deploy/Deploy.ps1
@@ -230,7 +230,7 @@ if (Test-Path $artifactsFolder -PathType Container) {
             }
         }
         else {
-            $allApps += $projectApps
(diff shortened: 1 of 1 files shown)


freddydk:
I don't think this is the right fix.
Unknown dependencies should be installed for both apps and test apps.

I think the problem here is that Get-BcInstalledExtensions doesn't return all the apps called `Microsoft__EXCLUDE_*` - these apps are also excluded from Extension Management in Business Central.

I think this is the right fix: https://github.com/microsoft/AL-Go/pull/1653

freddydk:
Please close this PR - we should not do this.

PR #1656, opened 2025-04-14
Check release tag when creating a release

### LINKED ISSUE #1627: [Bug]: Release changelog uses wrong tag to compare with
### AL-Go version

6.3 (customized)

### Describe the issue

A while ago, we had a typo in the tag parameter for the Create new release action, writing `32.9.0` instead of `23.9.0`. From that point on forward, all future releases (whose major version number is still well below 32) have been using the tag `32.9.0` to compare with.
I've since deleted the faulty tag and its release, but it could be a problem for certain release strategies.

### Expected behavior

When creating the changelog, AL Go should either use the latest tag or tag of the latest release. Alternatively, the workflow could look for the highest version below the version in the tag of the new release.
If it's necessary to use the highest tag version, the workflow should run into an error and inform the user.

### Steps to reproduce

- Enter a repository with AL Go
- Create a new release with tag `32.9.0`
- Introduce some changes
- Create a new release with tag `25.2.0`
- Introduce more changes
- Create a new release with tag `25.2.1`

At this point, both releases should use `32.9.0` to compare with.

The customized AL Go repository can be found [here](https://github.com/CBS-BC-AT/AL-Go).

Additionally, I've attached the log of the Create new release workflow.

freddydk:
This PR will check whether the latest release tag is lower than the release tag specified.

If this is not the case, you will get an error or a warning depending on whether you are creating a release based on the latest build (the default).

Fixes #1627

### DIFF
--- a/Actions/CreateReleaseNotes/README.md
+++ b/Actions/CreateReleaseNotes/README.md
@@ -14,6 +14,7 @@ none
 | :-- | :-: | :-- | :-- |
 | shell | | The shell (powershell or pwsh) in which the PowerShell script in this action should run | powershell |
 | token | | The GitHub token running the action | github.token |
+| appVersion | Yes | App version | |
 | tag_name | Yes | This release tag name | |
 | target_commitish | | Last commit to include in release notes | Latest |
 
--- a/Actions/Github-Helper.psm1
+++ b/Actions/Github-Helper.psm1
@@ -496,6 +496,40 @@ function SemVerStrToSemVerObj {
     }
 }
 
+# Compare SemVerStr1 and SemVerStr2
+# Returns -1 if SemVerStr1 < SemVerStrj2
+# Returns 1 if SemVerStr1 > SemVerStr2
+# Returns 0 if SemVerStr1 = SemVerStr2
+function CompareSemVerStrs {
+    Param(
+        [Parameter(Mandatory = $true)]
+        [string] $semVerStr1,
+        [Parameter(Mandatory = $true)]
+        [string] $semVerStr2
+    )
+
+    Process {
+        $semVerObj1 = $semVerStr1 | SemVerStrToSemVerObj
+        $semVerObj2 = $semVerStr2 | SemVerStrToSemVerObj
+        $result = $semVerObj1.Major.CompareTo($semVerObj2.Major)
+        if ($result -eq 0) {
+            $result = $semVerObj1.Minor.CompareTo($semVerObj2.Minor)
+            if ($result -eq 0) {
+                $result = $semVerObj1.Patch.CompareTo($semVerObj2.Patch)
+                if ($result -eq 0) {
+                    for ($i=0; $i -lt 5; $i++) {
+                        $result = ("$($semVerObj1."Addt$i")".CompareTo("$($semVerObj2."Addt$i")"))
+                        if ($result -ne 0) {
+                            return $result
+                        }
+                    }
+                }
+            }
+        }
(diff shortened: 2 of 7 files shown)


Actions/CreateReleaseNotes/README.md
```diff
@@ -14,6 +14,7 @@ none
 | :-- | :-: | :-- | :-- |
 | shell | | The shell (powershell or pwsh) in which the PowerShell script in this action should run | powershell |
 | token | | The GitHub token running the action | github.token |
+| appVersion | Yes | App version | |
```
mazhelez, line 17: Minor: I'd call this simply _version_ or _releaseVersion_ as it's not necessarily related to the app version (there could even be apps with different versions in the same release, right?)

Actions/Github-Helper.psm1
```diff
@@ -496,6 +496,40 @@ function SemVerStrToSemVerObj {
     }
 }
 
+# Compare SemVerStr1 and SemVerStr2
+# Returns -1 if SemVerStr1 < SemVerStrj2
+# Returns 1 if SemVerStr1 > SemVerStr2
+# Returns 0 if SemVerStr1 = SemVerStr2
+function CompareSemVerStrs {
```
mazhelez, line 503: Wouldn't comparison with `[Sytem.Version]` work instead?

Actions/CreateReleaseNotes/README.md
freddydk: It is called appVersion because that is the name in CreateRelease.yaml as well - build version would be better for both

Actions/Github-Helper.psm1
mazhelez: Saw the tests, nevermind 😄

mazhelez (approved):

Actions/CreateReleaseNotes/README.md
freddydk: renamed to build version

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.psm1
```diff
@@ -54,8 +54,8 @@ function ShouldBuildProject {
         $project,
         [Parameter(HelpMessage = "The base folder", Mandatory = $true)]
         $baseFolder,
-        [Parameter(HelpMessage = "A list of modified files", Mandatory = $true)]
-        $modifiedFiles
+        [Parameter(HelpMessage = "A list of modified files", Mandatory = $false)]
+        $modifiedFiles = @()
```
mazhelez, line 58: There could be an early exit if there are no modified files.

Actions/DetermineProjectsToBuild/DetermineProjectsToBuild.psm1
```diff
@@ -54,8 +54,8 @@ function ShouldBuildProject {
         $project,
         [Parameter(HelpMessage = "The base folder", Mandatory = $true)]
         $baseFolder,
-        [Parameter(HelpMessage = "A list of modified files", Mandatory = $true)]
-        $modifiedFiles
+        [Parameter(HelpMessage = "A list of modified files", Mandatory = $false)]
```
mazhelez, line 57: The doc above (lines 44-50) is inaccurate, by the way.

mazhelez (approved):

aholstrup1 (approved):

mazhelez (approved):

PR #1687, opened 2025-04-28
Enable blocking pull requests that add new warnings

jonasblunck:
Things to consider for feedback:

1. Invoking the check for warnings as the last thing in RunPipeline.ps1. Could move to a step in _BuildALGoProject.yaml, but not sure how value it adds?
2. Should the setting be named 'checkForNewWarnings', or 'blockNewWarnings', etc?
3. We could optimize runtime by only comparing warnings when the number of warnings from baseline has changed, with the risk of ignoring new warnings if introduced at the same time as other warnings are removed.
4. Warnings are only stored in the log if the setting 'enableCodeCop' is set to true. Should we throw an error if checkForNewWarnings is $true but enableCodeCop is $false?
5. A new compiler version could add new warnings. In this case, a new 'official' build has to run before a PR can be completed.


Actions/AL-Go-Helper.ps1
```diff
@@ -634,6 +634,7 @@ function ReadSettings {
         "installTestLibraries"                          = $false
         "installPerformanceToolkit"                     = $false
         "enableCodeCop"                                 = $false
+        "checkForNewWarnings"                           = $false
```
aholstrup1, line 637: blockNewWarnings or failOnNewWarnings is better IMO

Actions/CheckForNewWarnings/CheckForNewWarnings.ps1
```diff
@@ -0,0 +1,44 @@
...
+
+$baselineWorkflowRunId,$baselineWorkflowSHA = FindLatestSuccessfulCICDRun -repository $env:GITHUB_REPOSITORY -branch $targetBranch -token $token -retention $settings.incrementalBuilds.retentionDays
+if ($project) { $projectName = $project } else { $projectName = $env:GITHUB_REPOSITORY -replace '.+/' }
+    
+$mask = Get-Item $prBuildOutputFile | Select-Object -ExpandProperty BaseName 
+$artifact = GetArtifactsFromWorkflowRun -workflowRun $baselineWorkflowRunId -token $token -api_url $env:GITHUB_API_URL -repository $env:GITHUB_REPOSITORY -mask $mask -projects $projectName | Select-Object -First 1
+
+Write-Host "Downloading build logs from previous good build."
+
+$artifactsFolder = Join-Path $ENV:GITHUB_WORKSPACE ".warnings"
```
aholstrup1, line 34:
```suggestion
$artifactsFolder = Join-Path $ENV:RUNNER_TEMP ".warnings"
```

Actions/CheckForNewWarnings/CheckForWarningsUtils.psm1
```diff
@@ -0,0 +1,74 @@
...
+
+    if ($delta)
+    {
+        throw "New warnings were introduced in this PR."
+    }
+
+}
+
+Export-ModuleMember -Function Initialize-Directory
+Export-ModuleMember -Function Get-Warnings
```
aholstrup1, line 73: Is Get-Warnings used outside of this module? If not, we could probably just remove it from here

Actions/CheckForNewWarnings/CheckForNewWarnings.ps1
```diff
@@ -0,0 +1,44 @@
...
+    return
+}
+
+if (-not $settings.checkForNewWarnings)
+{
+    Write-Host "Checking for new warnings is disabled in the settings."
+    return
+}
+
+Write-Host "Analyzing build for new warnings..."
```
aholstrup1, line 21:
```suggestion
Write-Host "::group::Analyzing build for new warnings..."
```
Consider doing something like this (and add an Write-Host "::endgroup::" at the bottom of the script). It'll help organize the GitHub log 
image

Actions/AL-Go-Helper.ps1
freddydk: failOnNewWarnings is my favorite

Actions/AL-Go-Helper.ps1
```diff
@@ -634,6 +634,7 @@ function ReadSettings {
         "installTestLibraries"                          = $false
         "installPerformanceToolkit"                     = $false
         "enableCodeCop"                                 = $false
+        "failOnNewWarnings"                             = $false
```
mazhelez, line 637: Add doc for the new setting at Scenarios\settings.md

Actions/CheckForNewWarnings/CheckForNewWarnings.ps1
```diff
@@ -0,0 +1,49 @@
...
+}
+
+try
+{
+    Write-Host "::group::Analyzing build for new warnings..."
+
+    Import-Module (Join-Path $PSScriptRoot "..\Github-Helper.psm1" -Resolve) -DisableNameChecking
+    Import-Module (Join-Path $PSScriptRoot ".\CheckForWarningsUtils.psm1" -Resolve) -DisableNameChecking
+
+    $baselineWorkflowRunId,$baselineWorkflowSHA = FindLatestSuccessfulCICDRun -repository $env:GITHUB_REPOSITORY -branch $targetBranch -token $token -retention $settings.incrementalBuilds.retentionDays
```
mazhelez, line 28: RunPipeline.ps1 already has parameters `$baselineWorkflowRunId` and `$baselineWorkflowSHA`, so no need to query for the latest successful CICD here.

Actions/CheckForNewWarnings/CheckForNewWarnings.ps1
```diff
@@ -0,0 +1,49 @@
...
+    if ($project) { $projectName = $project } else { $projectName = $env:GITHUB_REPOSITORY -replace '.+/' }
+
+    $mask = Get-Item $prBuildOutputFile | Select-Object -ExpandProperty BaseName
+    $artifact = GetArtifactsFromWorkflowRun -workflowRun $baselineWorkflowRunId -token $token -api_url $env:GITHUB_API_URL -repository $env:GITHUB_REPOSITORY -mask $mask -projects $projectName | Select-Object -First 1
```
mazhelez, line 32: What if the PR adds a new project, so there isn't anything in the baseline? Defensive check might be needed.

Actions/CheckForNewWarnings/CheckForWarningsUtils.psm1
```diff
@@ -0,0 +1,90 @@
+<#
+    .SYNOPSIS
+    Ensures that a specified directory exists. If the directory does not exist, it creates it.
+    .DESCRIPTION
+    Ensures that a specified directory exists. If the directory does not exist, it creates it.
+#>
+function Initialize-Directory
```
mazhelez, line 7: Let's move this one to AL-Go-Helper.ps1 for now.

Actions/CheckForNewWarnings/CheckForNewWarnings.ps1
```diff
@@ -0,0 +1,49 @@
...
+
+try
+{
+    Write-Host "::group::Analyzing build for new warnings..."
+
+    Import-Module (Join-Path $PSScriptRoot "..\Github-Helper.psm1" -Resolve) -DisableNameChecking
+    Import-Module (Join-Path $PSScriptRoot ".\CheckForWarningsUtils.psm1" -Resolve) -DisableNameChecking
+
+    $baselineWorkflowRunId,$baselineWorkflowSHA = FindLatestSuccessfulCICDRun -repository $env:GITHUB_REPOSITORY -branch $targetBranch -token $token -retention $settings.incrementalBuilds.retentionDays
+    if ($project) { $projectName = $project } else { $projectName = $env:GITHUB_REPOSITORY -replace '.+/' }
```
mazhelez, line 29: Is there a case when `$project` isn't set?

`RunPipeline` always runs in the context of a project. Default is `'.'`, which means the repo root is the AL-Go project, being built.

Actions/CheckForNewWarnings/CheckForNewWarnings.ps1
```diff
@@ -0,0 +1,49 @@
...
+
+    $mask = Get-Item $prBuildOutputFile | Select-Object -ExpandProperty BaseName
```
mazhelez, line 31: The _buildMode_ should be included in the mask as well.

There's actually an action, "Calculate Artifact names" that calculates the all the artifact names for a project build.
In this case it is a coincidence that it's called the same way as the build output file itself. The proper way to get the build output artifact name is from that action (which currently runs after _RunPipeline_, so it's not trivial).

[Here](https://github.com/microsoft/AL-Go/blob/7d7d46b728ac7131f031d0765efba6a1ad025383/Actions/CalculateArtifactNames/CalculateArtifactNames.ps1#L52) you can check how artifact names are constructed.

mazhelez (commented):
Great addition to guarantee certain code quality level.
Left some comments, that need addressing.

Some other things to consider:
- one of the requests was actually to compare build outputs between two CICD runs. This is something we can support later on, but it'll be great if the helper functions cover such scenario.
- What is the expectation when refactoring? E.g. moving one AL project to another AL-Go project or renaming AL-Go projects. Please include a few lines in RELEASENOTES.md to describe the feature and the known limitations.

mazhelez:
My 2 cents:

> 1. Invoking the check for warnings as the last thing in RunPipeline.ps1. Could move to a step in _BuildALGoProject.yaml, but not sure how value it adds?

One pro is to leverage the "Calculate Artifact names" action (more info in one of my comments). But either way is fine.

> 2. Should the setting be named 'checkForNewWarnings', or 'blockNewWarnings', etc?

Naming is always hard. The most voted one is `failOnNewWarnings` as far as I can tell. 

> 3. We could optimize runtime by only comparing warnings when the number of warnings from baseline has changed, with the risk of ignoring new warnings if introduced at the same time as other warnings are removed.

My opinion is that this isn't needed. Bring the pain forward: if comparing the warning takes too much time, that's yet another reason to fix them 😉 

> 4. Warnings are only stored in the log if the setting 'enableCodeCop' is set to true. Should we throw an error if checkForNewWarnings is $true but enableCodeCop is $false?

No, no need for error. There are plenty of settings that "overwrite" other settings.

> 5. A new compiler version could add new warnings. In this case, a new 'official' build has to run before a PR can be completed.

Yes... As long as it's documented what to do in case of new warnings due to "outside" factors (new compiler, new rulesets, new baseline, etc.).

Actions/CheckForNewWarnings/CheckForWarningsUtils.psm1
jonasblunck: Skipping for now, waiting for refactoring.

Actions/CheckForNewWarnings/CheckForNewWarnings.ps1
```diff
@@ -0,0 +1,90 @@
+Param(
+    [string] $token,
+    [string] $project,
+    [object] $settings,
+    [string] $buildMode,
+    [string] $prBuildOutputFile,
+    [string] $baselineWorkflowRunId
+)
+
+function GetArtifactMask
```
freddydk, line 10: Everywhere else in AL-Go for GitHub we use a different formatting for { }

Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -536,6 +536,15 @@ try {
         Copy-Item -Path $buildOutputFile -Destination $destFolder -Force -ErrorAction SilentlyContinue
         Copy-Item -Path $containerEventLogFile -Destination $destFolder -Force -ErrorAction SilentlyContinue
     }
+
+    & $PSScriptRoot\..\CheckForNewWarnings\CheckForNewWarnings.ps1 `
```
freddydk, line 540: The CheckForNewWarnings is placed in the Actions area even though it isn't an action.
We also typically don't call functions by calling a script this way - we normally declare and call functions (I am aware we do it the other way in actions)

I think the code in CheckForNewWarnings.ps1 should be moved into the psm1 module - the module should be moved to the RunPipeline action folder, loaded and then the module imported and called here if we are to check for new warnings.

freddydk (changes requested):

aholstrup1 (approved):

freddydk (dismissed):

Actions/AL-Go-Helper.ps1
```diff
@@ -634,6 +634,7 @@ function ReadSettings {
...
         "enableUICop"                                   = $false
```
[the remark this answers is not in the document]
aholstrup1: We also pass failOn to Run-AlPipeline (bccontainerhelper)
https://github.com/microsoft/AL-Go/blob/4c8898a0e88a45102be6a6e78568f5be8f5f93d2/Actions/RunPipeline/RunPipeline.ps1#L515

Run-AlPipeline only supports 'none','error' and 'warning' at the moment so either we'll need a change in bccontainerhelper to support newWarning or we'll need to do something like this within RunPipeline.ps1

```
$failOn = $settings.failOn
if ($settings.failOn -eq 'newWarning') { 
    $failOn = "error"'
}
```

And then pass $failOn to Run-AlPipeline

Scenarios/settings.md
```diff
@@ -99,7 +99,7 @@ The repository settings are only read from the repository settings file (.github
 | <a id="enableUICop"></a>enableUICop | If enableUICop is set to true, the CI/CD workflow will enable the UICop analyzer when building. | false |
 | <a id="customCodeCops"></a>customCodeCops | CustomCodeCops is an array of paths or URLs to custom Code Cop DLLs you want to enable when building. | [ ] |
 | <a id="enableCodeAnalyzersOnTestApps"></a>enableCodeAnalyzersOnTestApps | If enableCodeAnalyzersOnTestApps is set to true, the code analyzers will be enabled when building test apps as well. | false |
-| <a id="failOn"></a>failOn | Specifies what the pipeline will fail on. Allowed values are none, warning and error | error |
+| <a id="failOn"></a>failOn | Specifies what the pipeline will fail on. Allowed values are none, warning, newWarning and error | error |
```
aholstrup1, line 102: Should we mention here that the "failing on new warning" mechanism is only enabled on PRs and that on any other workflow failOn=newWarning is effectively failOn=Error?

Actions/AL-Go-Helper.ps1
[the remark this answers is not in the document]
jonasblunck: Freddy fixed Run-AlPipeline to support new value.

aholstrup1 (approved):

freddydk (approved):

PR #1688, opened 2025-04-28
Use App in E2E tests during setup

aholstrup1:
This PR switches the E2E testing to use a GitHub app for setup and removal of the temp repositories. This is the first part of the work to remove our dependency on the E2EPAT wherever possible. Once this PR is merged, I will start working on using the same GitHub app for running the E2E tests. 

There are some places we will not be able to remove our dependency on the E2EPAT right now, but the goal (for now) is just to remove our dependency on the E2EPAT everywhere we can. 

Related to AB#573954


mazhelez (approved):

freddydk (approved):
Looks good

PR #1691, opened 2025-04-29
Fix for duplicated icons in test result visualizer.

spetersenms:


spetersenms:
There are 2 fixes in this PR which both contributed to the issue.
1) The _BuildTestMarkdownTable_ function edited the input rows array instead of creating a new one which was not intended in the original implementation.

2) The _GetTestResultSummaryMD_ function which calls the markdown helper calls it twice instead of once as the call is placed in a wrong location.

mazhelez (approved):
Release notes?

Actions/AnalyzeTests/TestResultAnalyzer.ps1
```diff
@@ -23,27 +23,32 @@ function BuildTestMarkdownTable {
...
-                    $row[$i] = $null
+                    $formattedRow += $null
                 }
+            } else {
+                $formattedRow += $row[$i]
             }
         }
-        Write-Host $row
+        $mdTableRows.Add($formattedRow) | Out-Null
+        Write-Host $formattedRow
```
freddydk, line 46: Maybe `"|$($formattedRow -join '|')|" | Out-Host` instead?
to balance the number of lines in the log?
(And the above Write-Host $row - could be `$row -join ',' | Out-Host`

freddydk (changes requested):
Please add a line to release notes and a 

`Fixes #<bugnumber>`

to the PR description to link the bug.

aholstrup1 (approved):

PR #1702, opened 2025-05-05
DeployTo setting supported as environment variable

spetersenms:
With this change, it is now possible to define your DeployTo setting as an environment variable in your GitHub environments.

### DIFF
--- a/Actions/AL-Go-Helper.ps1
+++ b/Actions/AL-Go-Helper.ps1
@@ -552,6 +552,8 @@ function ReadSettings {
         [string] $branchName = "$ENV:GITHUB_REF_NAME",
         [string] $orgSettingsVariableValue = "$ENV:ALGoOrgSettings",
         [string] $repoSettingsVariableValue = "$ENV:ALGoRepoSettings",
+        [string] $environmentName = "",
+        [string] $environmentDeployToVariableValue = "",
         [switch] $silent
     )
 
@@ -745,6 +747,12 @@ function ReadSettings {
         $projectFolder = Join-Path $baseFolder $project -Resolve
         $projectSettingsObject = GetSettingsObject -Path (Join-Path $projectFolder $ALGoSettingsFile)
         $settingsObjects += @($projectSettingsObject)
+    }
+    if ($environmentDeployToVariableValue) {
+        # Read settings from environment variable (parameter)
+        $environmentVariableObject = [pscustomobject]@{"DeployTo$environmentName" = ($environmentDeployToVariableValue | ConvertFrom-Json) } 
+        Write-Host "Environment variable settings: $environmentVariableObject" 
+        $settingsObjects += @($environmentVariableObject)
     }
     if ($workflowName) {
         # Read settings from workflow settings file
--- a/Actions/AL-Go-Helper.ps1
+++ b/Actions/AL-Go-Helper.ps1
@@ -750,8 +750,8 @@ function ReadSettings {
     }
     if ($environmentDeployToVariableValue) {
         # Read settings from environment variable (parameter)
-        $environmentVariableObject = [pscustomobject]@{"DeployTo$environmentName" = ($environmentDeployToVariableValue | ConvertFrom-Json) } 
-        Write-Host "Environment variable settings: $environmentVariableObject" 
+        $environmentVariableObject = [pscustomobject]@{"DeployTo$environmentName" = ($environmentDeployToVariableValue | ConvertFrom-Json) }
+        Write-Host "Environment variable settings: $environmentVariableObject"
         $settingsObjects += @($environmentVariableObject)
     }
     if ($workflowName) {
--- a/Templates/AppSource App/.github/workflows/CICD.yaml
+++ b/Templates/AppSource App/.github/workflows/CICD.yaml
@@ -269,6 +269,8 @@ jobs:
         with:
           shell: ${{ matrix.shell }}
           get: type,powerPlatformSolutionFolder
+          environmentName: ${{ matrix.environment }}
+          environmentDeployToVariableValue: ${{ vars.DeployTo }}
 
       - name: EnvName
         id: envName
--- a/Templates/Per Tenant Extension/.github/workflows/CICD.yaml
+++ b/Templates/Per Tenant Extension/.github/workflows/CICD.yaml
@@ -283,6 +283,8 @@ jobs:
         with:
           shell: ${{ matrix.shell }}
           get: type,powerPlatformSolutionFolder
+          environmentName: ${{ matrix.environment }}
+          environmentDeployToVariableValue: ${{ vars.DeployTo }}
 
       - name: EnvName
         id: envName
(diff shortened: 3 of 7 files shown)


Templates/AppSource App/.github/workflows/CICD.yaml
```diff
@@ -269,6 +269,8 @@ jobs:
         with:
           shell: ${{ matrix.shell }}
           get: type,powerPlatformSolutionFolder
+          environmentName: ${{ matrix.environment }}
```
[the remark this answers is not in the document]
spetersenms: Right that make sense, I'll take a look at that.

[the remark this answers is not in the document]
spetersenms: We should be good in private repos as well. The documentation for the vars context says that any configuration variable that has not been will return an empty string. Since org, repo and environment variables all exist in the vars context, there should be no problem, but just to be sure, I'll run a test in a private repo to verify:
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#example-usage-of-the-vars-context

RELEASENOTES.md
```diff
@@ -6,6 +6,24 @@
 
 - Issue 1697 Error in CheckForUpdates: "Internet Explorer engine is not available" when using self-hosted runners
 
+### DeployTo setting can now be defined in GitHub deployment environment variables.
```
mazhelez: Make sure to move this above v7.2

spetersenms: Done.

Actions/AL-Go-Helper.ps1
```diff
@@ -700,6 +701,8 @@ function ReadSettings {
         [string] $branchName = "$ENV:GITHUB_REF_NAME",
         [string] $orgSettingsVariableValue = "$ENV:ALGoOrgSettings",
         [string] $repoSettingsVariableValue = "$ENV:ALGoRepoSettings",
+        [string] $environmentName = "",
+        [string] $environmentDeployToVariableValue = "",
```
mazhelez: This is oddly specific here. I imagine this pattern will be replicated for another setting as well.

How about a more generic approach? E.g. allowing a `ALGoEnvironmentSettings` environment setting (simalar to `ALGoRepoSettings`, but on environment level).

spetersenms: Updated the naming to match the repo and org settings.

Actions/AL-Go-Helper.ps1
```diff
@@ -784,6 +786,28 @@ function ReadSettings {
...
+        # Warn user that 'DeployTo' setting needs to include environment name
+        if ($environmentVariableObject.PSObject.Properties.Name -contains "DeployTo") {
+            Write-Host "::warning::The environment settings variable contains the property 'DeployTo'. Did you intend to use 'DeployTo$environmentName' instead? The 'DeployTo' property without a specific environment name is not supported."
+        }
+        # Warn user if 'runs-on', 'shell' or 'ContinuousDeployment' is defined in the environment settings variable, as these are not supported when defined there.
+        if ($environmentVariableObject.PSObject.Properties.Name -contains "DeployTo$environmentName") {
+            @('runs-on', 'shell', 'ContinuousDeployment') | ForEach-Object {
+                if ($environmentVariableObject."DeployTo$environmentName".PSObject.Properties.Name -contains $_) {
+                    Write-Host "::warning::The property $_ in the DeployTo setting is not supported when defined within a GitHub deployment environment variable. Please define this property within the AL-Go repo settings file instead."
```
mazhelez: `OutputWarning` here to cover the case when this is run locally.

mazhelez: Also, the setting can be defined in multiple other places than "AL-Go repo settings file"

Actions/AL-Go-Helper.ps1
```diff
@@ -784,6 +786,28 @@ function ReadSettings {
...
+                }
+            }
+        }
+        else {
+            $sampleJson = "{`n`t`"DeployTo$environmentName`": {`n`t`t...`n`t}`n}"
+            Write-Host "::warning::The environment settings variable does not contain the property 'DeployTo$environmentName'. Did you define your environment settings at the top JSON level? `nDeployment settings should be defined within the 'DeployTo$environmentName' property, as in other settings like: `n$sampleJson"
```
mazhelez: That warning shouldn't be needed, I believe.

Templates/Per Tenant Extension/.github/workflows/CICD.yaml
```diff
@@ -279,6 +279,14 @@ jobs:
         with:
           path: '.artifacts'
 
+      - name: RepoEnvVariable
```
mazhelez: Why not define it as an env variable on the job directly?

RELEASENOTES.md
```diff
@@ -2,6 +2,26 @@
 
 AL-Go now offers a dataexplorer dashboard to get started with AL-Go telemetry. Additionally, we've updated the documentation to include a couple of kusto queries if you would rather build your own reports.
 
+### DeployTo setting can now be defined in GitHub deployment environment variables.
```
mazhelez: This could be broader. E.g. "Support for AL-Go settings as GitHub environment variable: ALGoEnvSettings" with `DeployTo` as a prime example.

mazhelez (commented):
Looks okay, but more attention is needed.

Templates/Per Tenant Extension/.github/workflows/CICD.yaml
spetersenms: Because GitHub for some reason evaluates the env: block before the environment context is set, meaning the vars object does not contain the env variable at that point in time.

Actions/AL-Go-Helper.ps1
spetersenms: While we don't really need it, the only settings we do support in this env variable at this point is DeployTo, which is why I added the warning. But if we want to keep it more generic, it can be removed.

RELEASENOTES.md
spetersenms: Yep good catch.

Templates/Per Tenant Extension/.github/workflows/CICD.yaml
mazhelez: This seems odd.

I just tried the following and the environment variable was read as expected:
```
on:
  workflow_dispatch:
          
jobs:
  display-variables:
    name: Display Env vars
    runs-on: ubuntu-latest
    environment: TestEnvironment
    env:
      environmentVariable: ${{ vars.TestVar }}
    steps:
    - name: Use variables
      shell: pwsh
      run: |
        Write-Host "Test var is $($env:environmentVariable)"
```

Actions/AL-Go-Helper.ps1
mazhelez: What your PR is supporting is another level of defining AL-Go settings. Namely, as an environment variable.

What is only supported currently is running the `Deploy` job on a GitHub environment. But even that isn't quite right, as both "Update AL-Go System Files" job (from the corresponding workflow) and "Deploy Reference Documentation" (from CICD) also run (or can be made to run) in an environment context.

Templates/Per Tenant Extension/.github/workflows/CICD.yaml
spetersenms: Just tested the same and see you are right. Not sure what went wrong in my own testing, because I was sure I tried this, but maybe did a typo in the name?

ankosiorek (approved):

PR #1711, opened 2025-05-06
Test result job summaries not displayed if no tests found.

spetersenms:
Currently if no tests are found, we display a job summary saying no tests were found. This takes up an unnecessary amount of space, so this PR fixes this by not displaying the summary at all, if no tests were found.


freddydk:
The way we avoid normal test results from being displayed is by not even calling the action:

```
      - name: Analyze Test Results
        id: analyzeTestResults
        if: (success() || failure()) && env.doNotRunTests == 'False' && ((hashFiles(format('{0}/.buildartifacts/TestResults.xml',inputs.project)) != '') || (hashFiles(format('{0}/.buildartifacts/bcptTestResults.json',inputs.project)) != ''))
        uses: microsoft/AL-Go-Actions/AnalyzeTests@main
        with:
          shell: ${{ inputs.shell }}
          testType: "normal"
```

why wouldn't we do the same with the two other?

freddydk (changes requested):
The check should be in the yaml file - not in code IMO

spetersenms:
I did see the check in the yaml, but went with the code approach for now. The reason being that we need the test file path in the code, so if something changes, we would need to change that path both in yaml and code which seems a bit less clean. The downside is the action will run, but it will basically just run a few if-statements and return nothing.
The code way is also a bit easier if we introduce new test types at some point.
If non of these are real concerns we can go with the yaml approach.

Actions/AnalyzeTests/AnalyzeTests.ps1
```diff
@@ -42,16 +42,19 @@ switch ($testType) {
...
-}
+# If no tests are found, don't add a job summary at all.
+if ($testResultsSummaryMD) {
+    # If summary fits, we will display it in the GitHub summary
+    if ($testResultsSummaryMD.Length -gt (1MB - 1)) {
+        # If Test results summary is too long, we will not display it in the GitHub summary, instead we will display a message to download the test results
+        $testResultsSummaryMD = "<i>Test results summary size exceeds GitHub summary capacity. Download **TestResults** artifact to see details.</i>"
+    }
+    if ($testResultsSummaryMD.Length + $testResultsfailuresMD.Length -gt (1MB - 1)) {
+        # If Combined Test Results and failures exceeds GitHub summary capacity, we will not display the failures details, only the failures summary
```
freddydk, line 53: Couldn't this
($testResultsSummaryMD.Length + $testResultsfailuresMD.Length -gt (1MB - 1))

still be true even though you just grab the failure summary?

And... you subtract 1 from 1Mb - but you add test title and newlines and more.

freddydk: You might want to compare with 1Mb - length(testtitle) - 16

freddydk:
> I did see the check in the yaml, but went with the code approach for now. The reason being that we need the test file path in the code, so if something changes, we would need to change that path both in yaml and code which seems a bit less clean. The downside is the action will run, but it will basically just run a few if-statements and return nothing. The code way is also a bit easier if we introduce new test types at some point. If non of these are real concerns we can go with the yaml approach.

Code is fine and you are right, maybe easier going forward - but then the if statement in the yaml should be removed I think

freddydk (changes requested):
The check should be in the yaml file - not in code IMO

Actions/AnalyzeTests/AnalyzeTests.ps1
spetersenms: Size check has been updated to take everything into account.

freddydk (approved):

aholstrup1 (approved):

PR #1735, opened 2025-05-15
Freddydk/avoidallsecrets

freddydk:
Avoid transfering all secrets available to the repository to ReadSecrets.

Instead Determine the secrets needed and read/transfer them.

Also refactor ReadSecrets to only attempt to read the missing secrets from Azure KeyVault as secrets from GitHub has already been resolved.

TODO
- [x] Refactor ReadSecrets to only get missing secrets from Azure DevOps
- [x] Check for max. number of secrets (32)


Actions/DetermineSecrets/action.yaml
```diff
@@ -0,0 +1,130 @@
...
+    default: powershell
+  getSecrets:
+    description: Comma-separated list of Secrets to get. Secrets preceded by an asterisk are returned encrypted
+    required: true
+  useGhTokenWorkflowForPush:
+    description: Determines whether you want to use the GhTokenWorkflow secret for TokenForPush
+    required: false
+    default: 'false'
+outputs:
+  FORMATSTR:
```
mazhelez, line 16: I still cannot understand why the action doesn't simply output a string that represents comma-separated secret names, that need to be read.
Why does the formatting have to be on the action caller instead of here?

freddydk: The challenge is to resolve all GitHub secrets in yaml.
Before this, we transferred all GitHub Secrets to PowerShell and had PowerShell do the working (i.e. ToJson(Secrets))
I tried a lot of different things in a prototype (here: https://github.com/freddydk/test/blob/main/.github/workflows/Secrets.yaml) and ended up with this construct as the only way I could make that work.
If the determineSecrets returns a list of secret names - how would you then get the secrets in YAML in a generic matter?

Actions/DetermineSecrets/DetermineSecrets.ps1
```diff
@@ -0,0 +1,107 @@
...
+# and one environment variable per secret called S0, S1, S2 with the name of the GitHub Secret (or Azure DevOps secret) to look for
+if ($script:secretsCollection.Count -gt 32) {
+    throw "Maximum number of secrets exceeded."
+}
+
+$cnt = 0
+$formatArr = @()
+foreach($secret in $script:secretsCollection) {
+    $formatArr += @("""$Secret"":{$cnt}")
+    Add-Content -Encoding UTF8 -Path $ENV:GITHUB_ENV -Value "S$cnt=$($script:secretNames[$secret])"
```
mazhelez, line 102: I think it's better to add these in the output as well.

The action does define S0, S1, etc. outputs.

Actions/DetermineSecrets/action.yaml
mazhelez: I finally get the limitation. Your approach works, so let's go with it.

mazhelez (commented):
Don't forget release notes.

Actions/DetermineSecrets/DetermineSecrets.ps1
freddydk: The action lists them as output because I had them there in the beginning.
I moved them to env in order to shorten the length of the FORMAT_JSON line - and to make sure that all instances of that line are equal.
If you want me to put them back as OUTPUT - I can do that - or I can remove the OUTPUT definitions.
(I think the latter makes most sense)

freddydk: The line would go from:

${{ format(steps.DetermineSecrets.outputs.formatStr,toJSON(secrets[env.S0]),toJSON(secrets[env.S1]),toJSON(secrets[env.S2]),toJSON(secrets[env.S3]),toJSON(secrets[env.S4]),toJSON(secrets[env...

to

```
${{ format(steps.DetermineSecrets.outputs.formatStr,toJSON(secrets[steps.DetermineSecrets.outputs.S0]),toJSON(secrets[steps.DetermineSecrets.outputs.S1]),toJSON(secrets[steps.DetermineSecrets.outputs.S2]),toJSON(secrets[steps.DetermineSecrets.outputs.S3]),toJSON(secrets[steps.DetermineSecrets.outputs.S4]),toJSON(secrets[steps.DetermineSecrets.outputs...
```

Neither is perfect, but I would prefer 1 over 2 - your call?

mazhelez: 2 is indeed longer, but 1 pollutes the env variables unnecessarily. Screenshot from https://github.com/freddydk/test/actions/runs/15132554080/job/42536747440

image


People don't look at yaml as often as the look at the logs. So, my vote goes for 2 if there are no other limitations.

freddydk: 2 it is - your wish is my command :-)

Actions/ReadSecrets/ReadSecrets.ps1
```diff
@@ -22,151 +22,92 @@ try {
...
-    $keyVaultCredentials = GetKeyVaultCredentials
-    $getAppDependencySecrets = $false
+    $keyVaultCredentials = GetKeyVaultCredentials -jsonStr $ghSecrets.AZURE_CREDENTIALS
     $getTokenForPush = $false
-    [System.Collections.ArrayList]$secretsCollection = @()
-    foreach($secret in ($getSecrets.Split(',') | Select-Object -Unique)) {
+    $secrets = @($ghSecrets.PSObject.Properties.Name)
+    for ($sno = 0; $sno -lt $secrets.Count; $sno++) {
+        $secret = $secrets[$sno]
+        $secretName = (get-item "ENV:S$sno").value
```
mazhelez, line 35: That part shouldn't work if secret names are not env variables.

freddydk: Good catch - forgot about this one - we actually have to have the mapping in an ENV variable

freddydk: reverted back to use ENV - will run the tests to see if I got all

PR #1739, opened 2025-05-15
Fix issue with transfering $schema propery when destination file does't exist

### LINKED ISSUE #1738: [Bug]: AL-Go settings schema brings empty settings files
### AL-Go version

preview

### Describe the issue

"Update AL-Go System Files" creates settings files with no settings

### Expected behavior

Settings files are either not created, or at least contain proper settings.

### Steps to reproduce

See https://github.com/microsoft/ALAppExtensions/pull/28627/files

Image

_No response_
Internal work item: [AB#578453](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/578453)

mazhelez:
Fixes #1738

### DIFF
--- a/Actions/CheckForUpdates/CheckForUpdates.HelperFunctions.ps1
+++ b/Actions/CheckForUpdates/CheckForUpdates.HelperFunctions.ps1
@@ -474,21 +474,23 @@ function GetModifiedSettingsContent {
 
     $srcSettings = Get-ContentLF -Path $srcSettingsFile | ConvertFrom-Json
 
-    if(Test-Path -Path $dstFile -PathType Leaf) {
+    if(Test-Path -Path $dstSettingsFile -PathType Leaf) {
         $dstSettings = Get-ContentLF -Path $dstSettingsFile | ConvertFrom-Json
+
+        # Change the $schema property to be the same as the source settings file (add it if it doesn't exist)
+        $schemaKey = '$schema'
+        $schemaValue = $srcSettings."$schemaKey"
+
+        $dstSettings | Add-Member -MemberType NoteProperty -Name "$schemaKey" -Value $schemaValue -Force
+
+        # Make sure the $schema property is the first property in the object
+        $dstSettings = $dstSettings | Select-Object @{ Name = '$schema'; Expression = { $_.'$schema' } }, * -ExcludeProperty '$schema'
     }
     else {
-        # If the destination settings file does not exist, create an empty object
-        $dstSettings = [PSCustomObject]@{}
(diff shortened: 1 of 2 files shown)


Actions/CheckForUpdates/CheckForUpdates.HelperFunctions.ps1
```diff
@@ -474,21 +474,23 @@ function GetModifiedSettingsContent {
...
+        $dstSettings | Add-Member -MemberType NoteProperty -Name "$schemaKey" -Value $schemaValue -Force
+
+        # Make sure the $schema property is the first property in the object
+        $dstSettings = $dstSettings | Select-Object @{ Name = '$schema'; Expression = { $_.'$schema' } }, * -ExcludeProperty '$schema'
     }
     else {
-        # If the destination settings file does not exist, create an empty object
-        $dstSettings = [PSCustomObject]@{}
+        # If the destination settings file does not exist, create an new settings object with default values from the source settings (which includes the $schema property already)
+        $dstSettings = $srcSettings
```
aholstrup1, line 491: I'm still getting the same issue 🤷‍♂️ 
https://github.com/aholstrup1/ALAppExtensions/pull/264/files

aholstrup1 (changes requested):

Actions/CheckForUpdates/CheckForUpdates.HelperFunctions.ps1
mazhelez: That doesn't look like the same issue.
If the settings file is not there (e.g. it was deleted w/o adding it to `unusedALGoSystemFiles`), AL-Go will bring it back.
That's what's happening with any other file from the templates.

Or are you refering to something else I am missing?

aholstrup1: Mmh fair but that didn't used to be the case, did it? 🤔 

What about "Test Current.settings.json". How do you avoid that getting deleted?

mazhelez: Remove it from `unusedALGoSystemFiles`.

I think I did mention at some point that the logic around adding the schema property is also fixing an issue with AL-Go settings files: new settings files are not transferred from the template and settings files in `unusedALGoSystemFiles` are not removed.

With the latest changes, settings files in the templates are treated just like the other files.

aholstrup1 (approved):

PR #1761, opened 2025-06-03
Add workflow to mark stale issues

mazhelez:
Utilize https://github.com/actions/stale/

### DIFF
--- a/.github/workflows/StaleIssues.yml
+++ b/.github/workflows/StaleIssues.yml
@@ -14,7 +14,6 @@ jobs:
     steps:
     - uses: actions/stale@v9.1.0
       with:
-        stale-issue-message: 'Stale issue message'
         stale-issue-label: 'stale'
         days-before-issue-stale: 180
         days-before-issue-close: 14
--- a/.github/workflows/StaleIssues.yml
+++ b/.github/workflows/StaleIssues.yml
@@ -3,6 +3,7 @@ name: Mark stale issues
 on:
   schedule:
   - cron: '44 5 * * 2' # Tuesday, 5:44
+  workflow_dispatch:
 
 jobs:
   stale:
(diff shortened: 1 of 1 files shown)


.github/workflows/StaleIssues.yml
```diff
@@ -0,0 +1,20 @@
...
+    permissions:
+      issues: write
+
+    steps:
+    - uses: actions/stale@v9.1.0
+      with:
+        stale-issue-label: 'stale'
+        days-before-issue-stale: 180
+        days-before-issue-close: 14
+        close-issue-message: 'There has been no recent activity on this issue. Please re-open or create a new issue, if needed.'
```
aholstrup1, line 20: Can we add value for the `stale-issue-message` parameter as well? Something like

"There has been no recent activity on this issue. If this is still an issue, please add a new comment."

.github/workflows/StaleIssues.yml
```diff
@@ -0,0 +1,20 @@
...
+  workflow_dispatch:
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+
+    steps:
+    - uses: actions/stale@v9.1.0
```
aholstrup1:
```suggestion
    - uses: actions/stale@5bef64f19d7facfb25b37b414482c7164d639639 # v9.1.0
```

mazhelez: Oh, yeah, good point!

.github/workflows/StaleIssues.yml
mazhelez: I intentionally omitted that part as I don't think it's necessary.
If the issue doesn't need to be closed, they can still comment, and we'll reopen it. Worst case, new issue is created.
It's a bit more aggressive approach, I know.

aholstrup1 (approved):

PR #1786, opened 2025-06-16
Fixed access issue when Publishing from PR in private repo.

spetersenms:
When using the Pubish to Environment workflow, targeting a PR in a private repo, the flow will fail with a permission issue, as the Github token does not have access to read the PR and PR annotations.
I have updated the permissions of the workflow to allow the Github token to access these.

Issue #1787

### DIFF
--- a/Templates/AppSource App/.github/workflows/PublishToEnvironment.yaml
+++ b/Templates/AppSource App/.github/workflows/PublishToEnvironment.yaml
@@ -15,6 +15,8 @@ permissions:
   actions: read
   contents: read
   id-token: write
+  pull-requests: read
+  checks: read
 
 defaults:
   run:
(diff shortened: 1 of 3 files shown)


Templates/AppSource App/.github/workflows/PublishToEnvironment.yaml
```diff
@@ -15,6 +15,8 @@ permissions:
   actions: read
   contents: read
   id-token: write
+  pull-requests: read
```
mazhelez, line 18: Release notes?

spetersenms: Created an Issue for proper tracking and linked it in PR as well as in release notes.

aholstrup1 (approved):

Issue #1512, opened 2025-03-17
[Bug]: Publish App Error on Pull Request Build with dependency on Continia Expenses


nickgoddard777:
### AL-Go version

preview

### Describe the issue

I have just created a PR for some functionality that is depedant on Contina Expenses. I can see in the logs that the script has worked out all the Continia dependencies, published them to my self-hosted docker container, compiled my app but when it tries to publish my app to the docker container I get errors saying it can't find various Continia obects I am referencing.
In the logs when I look at the last Intalled Apps entries I can see all the Continia Apps are installed.
Over the weekend I have also just moved us over to the PReview build and implemented Incremental Builds as this project has several apps and test apps. 

### Expected behavior

I would expect the app to be published

### Steps to reproduce

1. Create an App with a dependency to Continia Expenses
2. Setup the app to use Incremental Builds from the AL-GO Preview
3. Add a permissionSet Extension which extends CEM-APPROVE
4. Commit the App and create a PR.

### Additional context (logs, screenshots, etc.)

[2_Build Infopro Haynes (Default) _ Infopro Haynes (Default).txt](https://github.com/user-attachments/files/19290671/2_Build.Infopro.Haynes.Default._.Infopro.Haynes.Default.txt)

freddydk:
The nuGet packages you get from the MSSymbols feed are only symbols - they are not runtime packages, and as such they cannot be used to deploy apps to a container.
You can use them to build the app, but not to deploy and test.
We should fine a better error message for this.
We are also working on finalizing NuGet support to better allow partners to share NuGet runtime packages with partners and also allow you to deploy and run tests on online environments if you cannot get the runtime package.

aholstrup1:
We'll add some better error/warning for this. Will track it as part of another issue though

PR #1795, opened 2025-06-19
[Draft] Give a warning before trying to publish symbols-only packages 

aholstrup1:
### ❔What, Why & How

We've seen a couple of issues where builds are failing because one of the installApps is a symbols-only package. We could improve a bit here by displaying a warning when people try to do this. Eventually, this should probably be an error but showing a warning is a good first step IMO.

Related to issue: #1512

### ✅ Checklist

- [ ] Update RELEASENOTES.md

### DIFF
--- a/Actions/RunPipeline/RunPipeline.ps1
+++ b/Actions/RunPipeline/RunPipeline.ps1
@@ -235,7 +235,7 @@ try {
     $codeanalysisdll = Get-ChildItem -Path $tempFolder -Filter "Microsoft.Dynamics.Nav.CodeAnalysis.dll" -Recurse | Select-Object -First 1
     LoadDLL -path $codeanalysisdll.FullName
 
-    foreach($app in $allInstallApps) {
+    <#foreach($app in $allInstallApps) {
         # If the app is a URL, skip it
         if ($app -like 'http*://*') {
             continue
@@ -249,7 +249,7 @@ try {
             # This is likely a symbols package. Write a waning that this app shouldn't be published to a container
             OutputWarning -message "App $app is a symbols package and should not be published to a container. Skipping."
         }
-    }
+    }#>
 
     # Check if codeSignCertificateUrl+Password is used (and defined)
     if (!$settings.doNotSignApps -and $codeSignCertificateUrl -and $codeSignCertificatePassword -and !$settings.keyVaultCodesignCertificateName) {
--- a/Actions/RunPipeline/RunPipeline.ps1
+++ b/Actions/RunPipeline/RunPipeline.ps1
@@ -246,6 +246,8 @@ try {
         Write-Host "Analyzing app file $appFile"
         $packageStream = [System.IO.File]::OpenRead($appFile)
         $package = [Microsoft.Dynamics.Nav.CodeAnalysis.Packaging.NavAppPackageReader]::Create($PackageStream, $true)
+        Write-Host "IsRuntimePackage: $($package.IsRuntimePackage)"
+        Write-Host "SourceCodeFilePaths: $($package.ReadSourceCodeFilePaths())"
         if (($null -eq $package.ReadSourceCodeFilePaths()) -and (-not $package.IsRuntimePackage)) {
             # This is likely a symbols package. Write a waning that this app shouldn't be published to a container
             OutputWarning -message "App $app is a symbols package and should not be published to a container. Skipping."
--- a/Actions/RunPipeline/RunPipeline.ps1
+++ b/Actions/RunPipeline/RunPipeline.ps1
@@ -240,9 +240,12 @@ try {
         foreach($app in $allInstallApps) {
             # If the app is a URL, skip it for now
             if ($app -like 'http*://*') {
+                $appFile = "something.app" # TODO
+                Invoke-WebRequest -Uri $app -Method GET -OutFile $appFile
                 continue
+            } else {
+                $appFile = Join-Path $projectPath $app -Resolve -ErrorAction SilentlyContinue
             }
-            $appFile = Join-Path $projectPath $app -Resolve -ErrorAction SilentlyContinue
             if ($appFile) {
                 Write-Host "Analyzing app file $appFile"
                 $package = [Microsoft.Dynamics.Nav.CodeAnalysis.Packaging.NavAppPackageReader]::Create([System.IO.File]::OpenRead($appFile), $true)
--- a/Actions/RunPipeline/RunPipeline.ps1
+++ b/Actions/RunPipeline/RunPipeline.ps1
@@ -232,7 +232,8 @@ try {
     dotnet tool install Microsoft.Dynamics.BusinessCentral.Development.Tools --version "16.0.23.34358-beta" --tool-path $tempFolder | Out-Host
 
     # Find the Microsoft.Dynamics.Nav.CodeAnalysis.dll
-    $codeanalysisdll = Get-ChildItem -Path $tempFolder -Filter "Microsoft.Dynamics.Nav.CodeAnalysis.dll" -Recurse | Select-Object -First 1
+    $codeanalysisdll = Get-ChildItem -Path $tempFolder -Recurse | Where-Object { $_.FullName -like "*Microsoft.Dynamics.Nav.CodeAnalysis.dll" } | Select -Last 1
+    Write-Host "Found $codeanalysisdll - Loading dll"
     LoadDLL -path $codeanalysisdll.FullName
 
     foreach($app in $allInstallApps) {
--- a/Actions/RunPipeline/RunPipeline.ps1
+++ b/Actions/RunPipeline/RunPipeline.ps1
@@ -235,13 +235,13 @@ try {
     $codeanalysisdll = Get-ChildItem -Path $tempFolder -Filter "Microsoft.Dynamics.Nav.CodeAnalysis.dll" -Recurse | Select-Object -First 1
     LoadDLL -path $codeanalysisdll.FullName
 
-    <#foreach($app in $allInstallApps) {
+    foreach($app in $allInstallApps) {
         # If the app is a URL, skip it
         if ($app -like 'http*://*') {
             continue
         }
-        Write-Host "Analyzing app $app for version dependencies"
-        $appFile = Get-ChildItem $app
+        Write-Host "Analyzing app $app for version dependencies in $projectPath"
+        $appFile = Join-Path $projectPath $app -Resolve
         Write-Host "Analyzing app file $appFile"
         $packageStream = [System.IO.File]::OpenRead($appFile)
         $package = [Microsoft.Dynamics.Nav.CodeAnalysis.Packaging.NavAppPackageReader]::Create($PackageStream, $true)
@@ -249,7 +249,7 @@ try {
             # This is likely a symbols package. Write a waning that this app shouldn't be published to a container
             OutputWarning -message "App $app is a symbols package and should not be published to a container. Skipping."
         }
-    }#>
+    }
 
     # Check if codeSignCertificateUrl+Password is used (and defined)
     if (!$settings.doNotSignApps -and $codeSignCertificateUrl -and $codeSignCertificatePassword -and !$settings.keyVaultCodesignCertificateName) {
--- a/Actions/RunPipeline/RunPipeline.ps1
+++ b/Actions/RunPipeline/RunPipeline.ps1
@@ -248,7 +248,7 @@ try {
         $package = [Microsoft.Dynamics.Nav.CodeAnalysis.Packaging.NavAppPackageReader]::Create($PackageStream, $true)
         Write-Host "IsRuntimePackage: $($package.IsRuntimePackage)"
         Write-Host "SourceCodeFilePaths: $($package.ReadSourceCodeFilePaths())"
-        if (($null -eq $package.ReadSourceCodeFilePaths()) -and (-not $package.IsRuntimePackage)) {
+        if ((($null -eq $package.ReadSourceCodeFilePaths()) -or ($package.ReadSourceCodeFilePaths() -eq "")) -and (-not $package.IsRuntimePackage)) {
             # This is likely a symbols package. Write a waning that this app shouldn't be published to a container
             OutputWarning -message "App $app is a symbols package and should not be published to a container. Skipping."
         }
--- a/Actions/RunPipeline/RunPipeline.ps1
+++ b/Actions/RunPipeline/RunPipeline.ps1
@@ -219,6 +219,34 @@ try {
     # Analyze app.json version dependencies before launching pipeline
 
     # Analyze InstallApps and InstallTestApps before launching pipeline
+    function LoadDLL {
+    Param(
+        [string] $path
+    )
+        $bytes = [System.IO.File]::ReadAllBytes($path)
+        [System.Reflection.Assembly]::Load($bytes) | Out-Null
+    }
+    $allInstallApps = $install.Apps + $install.TestApps
+    # Create folder in temp directory with a unique name
+    $tempFolder = Join-Path $ENV:RUNNER_TEMP "DevelopmentTools-$(Get-Random)"
+    dotnet tool install Microsoft.Dynamics.BusinessCentral.Development.Tools --version "16.0.23.34358-beta" --tool-path $tempFolder | Out-Host
+
+    # Find the Microsoft.Dynamics.Nav.CodeAnalysis.dll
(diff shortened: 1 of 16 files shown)


Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -203,22 +205,31 @@ try {
...
+            return $appFile
         })
     }
 
     # Analyze app.json version dependencies before launching pipeline
 
     # Analyze InstallApps and InstallTestApps before launching pipeline
+    if ((-not $settings.useCompilerFolder) -and (-not $settings.doNotPublishApps)) {
+        # Test that InstallApps are not symbols packages
+        # Skip this check if we are using a compiler folder or doNotPublishApps is set
```
mazhelez: The compiler folder is used for building the apps, but if `doNotPublishApps` is _false_, a container will still be created to publish the apps.

So, isn't only the check for `doNotPublishApps` needed here?

Actions/RunPipeline/RunPipeline.psm1
```diff
@@ -0,0 +1,69 @@
...
+
+        foreach ($app in $AllInstallApps) {
+            $appFilePath = Join-Path $ProjectPath $app -Resolve -ErrorAction SilentlyContinue
+            if ($appFilePath) {
+                $appFile = Get-Item -Path $appFilePath
+                $appFileName = $appFile.Name
+                Write-Host "Analyzing app file $appFileName"
+                $package = [Microsoft.Dynamics.Nav.CodeAnalysis.Packaging.NavAppPackageReader]::Create([System.IO.File]::OpenRead($appFile), $true)
+                if ((($null -eq $package.ReadSourceCodeFilePaths()) -or ("" -eq $package.ReadSourceCodeFilePaths())) -and (-not $package.IsRuntimePackage)) {
+                    # If package is not a runtime package and has no source code files, it is likely a symbols package
```
mazhelez: "Likely"? 😄 
Isn't there definitive check?

Tests/RunPipeline.Action.Test.ps1
```diff
@@ -23,6 +23,27 @@ Describe "RunPipeline Action Tests" {
...
+
+        # Invoke the function with TestApp1 (a symbols package) and TestApp2 (a full app package)
+        $tempFolder = [System.IO.Path]::GetTempPath()
+        Test-InstallApps -AllInstallApps @(".\TestApps\EssentialBusinessHeadlinesFull.app", ".\TestApps\EssentialBusinessHeadlinesSymbols.app") -ProjectPath $PSScriptRoot -RunnerTempFolder $tempFolder
+
+        # Assert that the warning was output
+        Should -Invoke -CommandName 'OutputWarning' -Times 1 -ModuleName RunPipeline
+        Should -Invoke -CommandName 'OutputWarning' -Times 1 -ModuleName RunPipeline -ParameterFilter { $Message -like "*App EssentialBusinessHeadlinesSymbols.app is a symbols package and should not be published. The workflow may fail if you try to publish it." }
+
+        # Assert that Trace-Information was not called
```
mazhelez:
```suggestion
        # Assert that Trace-Warning was not called
```

mismatch

RELEASENOTES.md
```diff
@@ -48,6 +48,7 @@ Please note, that due to certain security limitations, the properties `runs-on`,
 - Issue 1837 Deployment from PR builds fail if PR branch name includes forward slashes (e.g., `feature/branch-name`).
 - Issue 1852 Page Scripting Tests are not added to build summary
 - Issue 1829 Added custom jobs cannot be removed
+- Issue 1512: Throw a warning before trying to publish symbols packages
```

Actions/Environment.Packages.proj
```diff
@@ -12,6 +12,7 @@
     <PackageReference Include="Az.Accounts" Version="2.15.1" GeneratePathProperty="true" ExcludeAssets="All" />
     <PackageReference Include="Az.Storage" Version="6.1.1" GeneratePathProperty="true" ExcludeAssets="All" />
     <PackageReference Include="Az.KeyVault" Version="5.2.0" GeneratePathProperty="true" ExcludeAssets="All" />
+    <PackageReference Include="Microsoft.Dynamics.BusinessCentral.Development.Tools" Version="16.0.24.41895-beta" GeneratePathProperty="true" ExcludeAssets="All" />
```
spetersenms: This version used to exist in our internal feed, but was deleted at some point, so this would fail the component governance flow. We can restore it in our feed, but not sure why it was deleted in the first place. This version does seem to be the latest one available in NuGet Gallery though, so probably can't use a newer version.

Actions/RunPipeline/RunPipeline.psm1
```diff
@@ -0,0 +1,81 @@
...
+                    # Symbols packages are not meant to be published to a BC Environment
+                    OutputWarning -Message "App $appFileName is a symbols package and should not be published. The workflow may fail if you try to publish it."
+                }
+            } else {
+                Write-Host "App file path for $app could not be resolved ($appFilePath). Skipping symbols check."
+            }
+        }
+    }
+    catch {
+        Trace-Warning -Message "Something went wrong while analyzing install apps."
```
spetersenms, line 76: Maybe we should write a debug message with the error as well?
```suggestion
        Trace-Warning -Message "Something went wrong while analyzing install apps."
        OutputDebug -message "Error: $_"
```

Scenarios/settings.md
```diff
@@ -100,7 +100,7 @@ The repository settings are only read from the repository settings file (.github
...
+| <a id="installApps"></a>installApps | An array of 3rd party dependency apps, which you do not have access to through the appDependencyProbingPaths. The setting should be an array of either secure URLs or paths to folders or files relative to the project, where the CI/CD workflow can find and download the apps. The apps in installApps are downloaded and installed before compiling and installing the apps. Before adding an app to the installApps property, please ensure it is not a symbols-only package. Symbols-only packages cannot be published and installed during CI/CD and will throw an error. <br/>**Note:** If you specify ${{SECRETNAME}} as part of a URL, it will be replaced by the value of the secret SECRETNAME. | [ ] |
```
mazhelez: Minor: wording.

```suggestion
| <a id="installApps"></a>installApps | An array of 3rd party dependency apps, which you do not have access to through the appDependencyProbingPaths. The setting should be an array of either secure URLs or paths to folders or files relative to the project, where the build workflow can find and download the apps. The apps in installApps are downloaded and installed before compiling and installing the apps in the current project. Before adding an app to the installApps property, please ensure it is not a symbols-only package. Symbols-only packages are not intended for publishing and installing and using them via installApps might lead to compilation or runtime errors.<br/>**Note:** If you specify ${{SECRETNAME}} as part of a URL, it will be replaced by the value of the secret SECRETNAME. | [ ] |
```

Actions/RunPipeline/RunPipeline.psm1
```diff
@@ -0,0 +1,81 @@
...
+                $appFileName = $appFile.Name
+                Write-Host "Analyzing app file $appFileName"
+                $package = [Microsoft.Dynamics.Nav.CodeAnalysis.Packaging.NavAppPackageReader]::Create([System.IO.File]::OpenRead($appFile), $true)
+                if ((($null -eq $package.ReadSourceCodeFilePaths()) -or ("" -eq $package.ReadSourceCodeFilePaths())) -and (-not $package.IsRuntimePackage)) {
+                    # If package is not a runtime package and has no source code files, it is a symbols package
+                    # Symbols packages are not meant to be published to a BC Environment
+                    OutputWarning -Message "App $appFileName is a symbols package and should not be published. The workflow may fail if you try to publish it."
+                }
+            } else {
+                Write-Host "App file path for $app could not be resolved ($appFilePath). Skipping symbols check."
```
mazhelez: `$appFilePath` is null or empty here. No need to add it to the message.

Actions/RunPipeline/RunPipeline.psm1
mazhelez, line 76: Maybe output the error here?

Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -203,22 +205,59 @@ try {
...
+                [array]$AllInstallApps,
+                [string]$ProjectPath,
+                [string]$RunnerTempFolder
+            )
+
+            Import-Module (Join-Path -Path $ScriptRoot -ChildPath ".\RunPipeline.psm1" -Resolve)
+            Test-InstallApps -AllInstallApps $AllInstallApps -ProjectPath $ProjectPath -RunnerTempFolder $RunnerTempFolder
+        }
+        # Filter out URLs and non-existing files
+        $allInstallApps = ($install.Apps + $install.TestApps) | Where-Object { $_ -notlike 'http*://*' }
```
mazhelez: What's `$install.Apps` here? Isn't it `$installApps`?

aholstrup1: No, that's in the Run-ALPipeline function. We pass $install.apps as the installApps parameter to Run-ALPipeline 
-installApps $install.apps

Actions/RunPipeline/RunPipeline.psm1
aholstrup1: I added an OutputDebug here instead

Actions/Environment.Packages.proj
aholstrup1: Nuget.org is an upstream source to our internal feed. Shouldn't the pipeline be able to pull it from the upstream source into our feed?

Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -203,22 +205,59 @@ try {
...
+                    $urlWithoutQuery = $finalUrl.Split('?')[0].TrimEnd('/')
+                    $rawFileName = [System.IO.Path]::GetFileName($urlWithoutQuery)
+                    $decodedFileName = [Uri]::UnescapeDataString($rawFileName)
+                    $appFile = Join-Path $tempDependenciesLocation $decodedFileName
+                    Invoke-WebRequest -Method GET -UseBasicParsing -Uri $finalUrl -OutFile $appFile | Out-Null
                 }
                 catch {
-                    throw "Setting: install$($list) contains an inaccessible URL: $($url). Error was: $($_.Exception.Message)"
+                    # If the app file fails to be downloaded we keep it as a URL and let Run-ALPipeline download it
+                    Write-Host "Failed to download app from $finalUrl. Error was: $($_.Exception.Message)"
```
mazhelez: Would be weird to see this "error" in the logs.
Why not fail here directly? In what scenario `Run-ALPipeline` succeeds in downloading the file if it fails here?

Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -25,6 +25,7 @@ try {
     Import-Module (Join-Path $PSScriptRoot '..\TelemetryHelper.psm1' -Resolve)
     DownloadAndImportBcContainerHelper
     Import-Module (Join-Path -Path $PSScriptRoot -ChildPath "..\DetermineProjectsToBuild\DetermineProjectsToBuild.psm1" -Resolve) -DisableNameChecking
+    Import-Module (Join-Path -Path $PSScriptRoot -ChildPath ".\RunPipeline.psm1" -Resolve)
```
mazhelez: This import doesn't seem needed here.

Actions/RunPipeline/RunPipeline.psm1
```diff
@@ -0,0 +1,95 @@
...
+    )
+
+    if ($AllInstallApps.Count -eq 0) {
+        Write-Host "No install apps to analyze."
+        return
+    }
+
+    try {
+        # Create folder in temp directory with a unique name
+        $tempFolder = Join-Path $RunnerTempFolder "DevelopmentTools-$(Get-Random)"
```
mazhelez: Why not installing the tool locally? 

For better experience in the self-hosted runner case, consider reusing the folder and checking if the tool is already installed.

aholstrup1: My concern with doing that was mainly around what will happen when we update the package version in AL-Go. We'd need some logic that can update the package version on-the-fly. Seemed easier and less risky just to install it in a temp folder that is only used for that job.

mazhelez: `dotnet tool update`?

You can also still use `--tool-path` and install the higher version when needed (a folder per version?).

Tests/RunPipeline.Action.Test.ps1
```diff
@@ -23,6 +23,27 @@ Describe "RunPipeline Action Tests" {
...
+        # Invoke the function with TestApp1 (a symbols package) and TestApp2 (a full app package)
+        $tempFolder = [System.IO.Path]::GetTempPath()
+        Test-InstallApps -AllInstallApps @(".\TestApps\EssentialBusinessHeadlinesFull.app", ".\TestApps\EssentialBusinessHeadlinesSymbols.app") -ProjectPath $PSScriptRoot -RunnerTempFolder $tempFolder
+
+        # Assert that the warning was output
+        Should -Invoke -CommandName 'OutputWarning' -Times 1 -ModuleName RunPipeline
+        Should -Invoke -CommandName 'OutputWarning' -Times 1 -ModuleName RunPipeline -ParameterFilter { $Message -like "*App EssentialBusinessHeadlinesSymbols.app is a symbols package and should not be published. The workflow may fail if you try to publish it." }
+
+        # Assert that Trace-Warning was not called
+        Assert-MockCalled Trace-Warning -Exactly 0 -ModuleName RunPipeline
```
mazhelez: Why not trace the warning actually?

Actions/RunPipeline/RunPipeline.psm1
mazhelez: Also, I recommend separating the installation of `Microsoft.Dynamics.BusinessCentral.Development.Tools` in another function.

Tests/RunPipeline.Action.Test.ps1
aholstrup1: I only trace a warning to telemetry if there is some internal error in the check. This asserts that it didn't fail during the analysis.

Actions/RunPipeline/RunPipeline.psm1
aholstrup1: > dotnet tool update?
> 
>  You can also still use --tool-path and install the higher version when needed (a folder per version?).

I tried to avoid installing the tool globally (i.e. `dotnet tool install --global`) as I'm not totally sure how it would work on self-hosted runners and whether there could be some caveats. 

Using a specific tool-path on the runner that we keep around from build-to-build could be an option but again could have some caveats: 

- What happens if/when the file is corrupted? We've seen that happen with BCArtifacts for example. 
- Could we run into file locks if multiple runners are requesting it at the same time? 
- Once we update to a new version, should AL-Go go and clean up the old folder? 

Installing it in a temp folder unique to each runner seemed like the less risky approach to me. 

> Also, I recommend separating the installation of Microsoft.Dynamics.BusinessCentral.Development.Tools in another function.

Will do. I think we have similar logic for installing the sign tool actually so I can refactor these to use the same function.

Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -203,22 +204,44 @@ try {
             else {
                 $finalUrl = $url
             }
-            # Check validity of URL
+
             if ($finalUrl -like 'http*://*') {
+                # Check validity of URL
                 try {
                     Invoke-WebRequest -Method Head -UseBasicParsing -Uri $finalUrl | Out-Null
```
mazhelez: This check was added to fail early, if case the URL isn't accessible.
Now that apps are downloaded a couple of lines below, this check isn't necessary.

Tests/RunPipeline.Action.Test.ps1
mazhelez: That's what I mean: why not also trace a warning also when symbols are about to be published?

mazhelez (approved):

Tests/RunPipeline.Action.Test.ps1
aholstrup1: Added a trace at the end of the function and updated the test

mazhelez:
### ⚠️ Release Notes Update Reminder

Thank you for updating the release notes! 

Please ensure that your changes are placed **above the new version section** (currently `## v8.1`) in the RELEASENOTES.md file.

This helps maintain a clear changelog structure where new changes are grouped under the latest unreleased version.

spetersenms (approved):

RELEASENOTES.md
```diff
@@ -1,5 +1,6 @@
 ### Issues
```
aholstrup1, line 1: A new version of AL-Go (v8.1) has been released.

Please move your release notes changes to above the `## v8.1` section in the RELEASENOTES.md file.

PR #1796, opened 2025-06-19
Improve E2E test Stability

aholstrup1:
### ❔What, Why & How

* Refresh E2E access token more aggressively
* Update how app signing is validated

### DIFF
--- a/e2eTests/e2eTestHelper.psm1
+++ b/e2eTests/e2eTestHelper.psm1
@@ -67,7 +67,8 @@ function RefreshToken {
             throw "Token not set."
         }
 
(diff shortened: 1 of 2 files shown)


e2eTests/e2eTestHelper.psm1
```diff
@@ -68,7 +68,7 @@ function RefreshToken {
         }
 
         # Check if the last token refresh was more than 30 minutes ago
```
mazhelez:
```suggestion
        # Check if the last token refresh was more than 10 minutes ago

```

mazhelez (approved):

PR #1828, opened 2025-07-17
Stop AL-Go from purging non AL-Go BC containers.

### LINKED ISSUE #1675: [Bug]: AL-Go docker pruning too aggressive(?)
### AL-Go version

6.3

### Describe the issue

We have a private Azure VM we've been using for the past year to unit test development tickets (for both saas and onprem apps). The docker instances on that VM are created via bccontainerhelper with the usetraefik option. 

last week we installed self-hosted github runners on the same VM (because our standard github-hosted runner build pipelines were failing for apps with large third party dependencies)

My team alereted me this morning that ALL BC containers on that VM are GONE (both saas and onprem containers).

Are the AL-GO self-hosted runners performing any container cleanup/housekeeping? If so, it appears the current housekeeping that is implemented in algo/bccontainerhelper is far too agresssive (why its pruning containers NOT created by AL-Go itself is beyond me). 

Please consider adding in settings to whitelist by wildcard which containers AL-GO should and shouldnt touch. 

Please advise. 

The self-hosted runner documention BTW doesn't mention any risks of running a self-hosted runner on a server where bccontainerhelper is actiively used locally. 

### Expected behavior

al-go/bccontainerhelper shouldn't touch/purge BC containers not spun up by AL-Go

### Steps to reproduce

create a docker container (saas or onprem) on a machine via bccontainerhelper. next, install a self-hosted runner on the same machine. I don't know the details beyond that, but eventually the self-hosted runner will kill off the BC Container previously created (by docker container creation date?)

_No response_

spetersenms:
During the CICD workflow, AL-Go calls https://github.com/microsoft/navcontainerhelper/blob/main/ContainerHandling/Flush-ContainerHelperCache.ps1, which kills any BC container that has been off for 3 days or more, regardless if that container was created by AL-Go.
This scenario can happen if you use self-hosted GitHub runners on the same machine/VM that you use for other BC development.

Related to issue: #1675


spetersenms:
This PR on the navcontainerhelper side, adds the functionaltiy to check for the AL-Go creator label: https://github.com/microsoft/navcontainerhelper/pull/3983

Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -515,7 +515,8 @@ try {
         -pageScriptingTestResultsFolder (Join-Path $buildArtifactFolder 'PageScriptingTestResultDetails') `
         -CreateRuntimePackages:$CreateRuntimePackages `
         -appBuild $appBuild -appRevision $appRevision `
-        -uninstallRemovedApps
+        -uninstallRemovedApps `
+        -NewBcContainer { Param([Hashtable]$parameters) $parameters.additionalParameters += @("--label creator=AL-Go"); New-BcContainer @parameters; Invoke-ScriptInBcContainer $parameters.ContainerName -scriptblock { $progressPreference = 'SilentlyContinue' } }
```
mazhelez: This won't work when there is an override for `NewBcContainer`.

spetersenms: Changed the logic to only add this NewBcContainer param if no override is used.

PR #1851, opened 2025-08-06
Using Governed template for Component Governance pipeline

spetersenms:
Updated CG pipeline to run on 1ES governed template.


tomasmsft (approved):

Actions/Environment.Packages.proj
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <TargetFramework>net48</TargetFramework>
+    <OutputPath>./.out</OutputPath>
+    <RestorePackagesPath>$(OutputPath)/packages</RestorePackagesPath>
+    <RestoreProjectStyle>PackageReference</RestoreProjectStyle>
+    <OutputType>Library</OutputType>
+    <RestoreSources>https://dynamicssmb2.pkgs.visualstudio.com/_packaging/DynamicsSMBPackages/nuget/v3/index.json</RestoreSources>
```
aholstrup1: We typically keep this in a Nuget.config in the root folder of the repo. I seem to remember having to do some work in the past to ensure this config is always present in the root folder. 

Could we do that?

Actions/Environment.Packages.proj
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <TargetFramework>net48</TargetFramework>
```
aholstrup1: Let's go with something slightly newer? I think we use .NET 8 internally. I guess it doesn't matter too much what we set this isn't an actual .NET project.

Actions/Environment.Packages.proj
spetersenms: I don't really mind using a NuGet.config, just figured it looked a bit cleaner to keep it all in one file, since it's "faked" and only used for the CG task anyway. 
Is it an internal compliance requirement to have that file in the root?

Actions/Environment.Packages.proj
spetersenms: Actually just realized we don't need a TargetFramework at all. The restore works fine without.

aholstrup1 (dismissed):

aholstrup1 (approved):

PR #1863, opened 2025-08-13
AL code analysis tracking in GitHub.

spetersenms:
Adding new feature to track AL code analysis results in GitHub.

We already have existing CodeCop features for AL in AL-Go, but this builds on top of that and integrates any code scan issues with GitHub security.

This allows alerts to appear in the Security tab of GitHub, and as automated comments on PRs if new alerts are detected.

### DIFF
--- a/Actions/CalculateArtifactNames/action.yaml
+++ b/Actions/CalculateArtifactNames/action.yaml
@@ -46,6 +46,9 @@ outputs:
   ContainerEventLogArtifactsName:
     description: Artifacts name for ContainerEventLog
     value: ${{ steps.calculateartifactnames.outputs.ContainerEventLogArtifactsName }}
+  ErrorLogsArtifactsName:
+    description: Artifacts name for ErrorLogs
+    value: ${{ steps.calculateartifactnames.outputs.ErrorLogsArtifactsName }}
   BuildMode:
     description: Build mode used when building the artifacts
     value: ${{ steps.calculateartifactnames.outputs.BuildMode }}
--- a/Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
+++ b/Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
@@ -1,0 +1,76 @@
+$errorLogsFolder = Join-Path $ENV:GITHUB_WORKSPACE "ErrorLogs"
+$errorLogFiles = Get-ChildItem -Path $errorLogsFolder -Filter "*.errorLog.json" -File -Recurse
+
+# Base SARIF structure
+$sarif = @{
+    version = "2.1.0"
+    '$schema' = "https://json.schemastore.org/sarif-2.1.0.json"
+    runs = @(@{
+        tool = @{
+            driver = @{
+                name = 'ALCodeAnalysis'
+                informationUri = 'https://aka.ms/AL-Go'
+                rules = @()
+            }
+        }     
+        results = @()
+    })
+}
+
+
+function GenerateSARIFJson {
+    param(
+        [PSCustomObject] $errorLogContent
+    )
+
+    foreach ($issue in $errorLogContent.issues) {
+        # Add rule if not already added
+        if (-not ($sarif.runs[0].tool.driver.rules | Where-Object { $_.id -eq $issue.ruleId })) {
+            $sarif.runs[0].tool.driver.rules += @{
+                id = $issue.ruleId
+                shortDescription = @{ text = $issue.fullMessage }
+                fullDescription = @{ text = $issue.fullMessage }
+                helpUri = $issue.properties.helpLink
+                properties = @{
+                    category = $issue.properties.category
+                    severity = $issue.properties.severity
+                }
+            }
+        }
+
+        # Convert absolute path to relative path from repository root
+        $absolutePath = $issue.locations[0].analysisTarget[0].uri
+        $workspacePath = $ENV:GITHUB_WORKSPACE
+        $relativePath = $absolutePath.Replace($workspacePath, '').TrimStart('\').Replace('\', '/')
+
+        # Add result
+        $sarif.runs[0].results += @{
+            ruleId = $issue.ruleId
+            message = @{ text = $issue.shortMessage }
+            locations = @(@{
+                physicalLocation = @{
+                    artifactLocation = @{ uri = $relativePath }
+                    region = $issue.locations[0].analysisTarget[0].region
+                }
+            })
+            level = ($issue.properties.severity).ToLower()
+        }
+    }
+}
+
+$errorLogFiles | ForEach-Object {
+    OutputDebug -message "Found error log file: $($_.FullName)"
+    $fileName = $_.Name
+    try {
+        $errorLogContent = Get-Content -Path $_.FullName -Raw | ConvertFrom-Json
+        GenerateSARIFJson -errorLogContent $errorLogContent
+    }
+    catch {
+        Write-Host "Failed to process $fileName"
+        OutputDebug -message "Failed to read error log file: $_"
+    }
+}
+
+$sarifJson = $sarif | ConvertTo-Json -Depth 10 -Compress
+Write-Host ($sarifJson)
+Set-Content -Path "$errorLogsFolder/output.sarif.json" -Value $sarifJson
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -1,3 +1,9 @@
+### AL Code Analysis tracked in GitHub
+
+AL-Go already supports AL code analysis, but up until now this was not tracked in GitHub. It is now possible to track code analysis issues automatically in the GitHub security tab, as well as having any new issues posted as a comment in Pull Requests. 
+
+Please note that some automated features are premium and require the use of [GitHub Code Security](https://docs.github.com/en/get-started/learning-about-github/about-github-advanced-security)
+
 ### AL-Go Telemetry
 
 AL-Go now offers a dataexplorer dashboard to get started with AL-Go telemetry. Additionally, we've updated the documentation to include a couple of kusto queries if you would rather build your own reports.
--- a/Templates/AppSource App/.github/workflows/CICD.yaml
+++ b/Templates/AppSource App/.github/workflows/CICD.yaml
@@ -18,6 +18,7 @@ permissions:
   contents: read
   id-token: write
   pages: read
+  security-events: write
 
 env:
   workflowDepth: 1
@@ -196,6 +197,43 @@ jobs:
       signArtifacts: true
       useArtifactCache: true
       needsContext:  ${{ toJson(needs) }}
+
+  CodeAnalysisUpload:
+    needs: [ Initialization, Build ]
+    if: (!cancelled())
+    runs-on: [ windows-latest ]
+    name: Code Analysis Processing
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+          
+      - name: Read settings
+        uses: microsoft/AL-Go-Actions/ReadSettings@main
+        with:
+          shell: powershell
+          get: trackCodeCopIssuesInGitHub
+
+      - name: Download artifacts - ErrorLogs
+        uses: actions/download-artifact@634f93cb2916e3fdff6788551b99b062d0335ce0 # v5.0.0
+        if: (success() || failure()) && env.trackCodeCopIssuesInGitHub == 'True'
+        with:
+          pattern: '*-ErrorLogs-*'
+          path: '${{ github.workspace }}/ErrorLogs/'
+          merge_multiple: true
+
+      - name: Process AL Code Analysis Logs
+        id: ProcessALCodeAnalysisLogs
+        if: (success() || failure()) && env.trackCodeCopIssuesInGitHub == 'True'
+        uses: microsoft/AL-Go-Actions/ProcessALCodeAnalysisLogs@main
+        with:
+          shell: ${{ inputs.shell }}
+
+      - name: Upload SARIF file to GitHub
+        uses: github/codeql-action/upload-sarif@v3
+        if: always() && env.trackCodeCopIssuesInGitHub == 'True'
+        with:
+          sarif_file: '${{ github.workspace }}/ErrorLogs/output.sarif.json'
+          category: "ALCodeAnalysis"
 
   DeployALDoc:
     needs: [ Initialization, Build ]
--- a/Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
+++ b/Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
@@ -17,6 +17,7 @@ permissions:
   contents: read
   id-token: write
   pull-requests: read
+  security-events: write
 
 env:
   workflowDepth: 1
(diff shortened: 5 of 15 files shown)


Templates/AppSource App/.github/workflows/CICD.yaml
```diff
@@ -197,6 +198,43 @@ jobs:
       useArtifactCache: true
       needsContext:  ${{ toJson(needs) }}
 
+  CodeAnalysisUpload:
+    needs: [ Initialization, Build ]
+    if: (!cancelled())
```
aholstrup1: Could we output something in the Initialization job that indicates whether the CodeAnalysisUpload job should be run? Then we can skip it entirely if the setting isn't enabled.

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
```diff
@@ -0,0 +1,81 @@
+$errorLogsFolder = Join-Path $ENV:GITHUB_WORKSPACE "ErrorLogs"
```
aholstrup1: Minor: Make this a parameter so it's easier to run this script locally

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
```diff
@@ -0,0 +1,81 @@
...
+                    artifactLocation = @{ uri = $relativePath }
+                    region = $issue.locations[0].analysisTarget[0].region
+                }
+            })
+            level = ($issue.properties.severity).ToLower()
+        }
+    }
+}
+
+if (Test-Path $errorLogsFolder -PathType Container){
```
aholstrup1: I'm considering if we should wrap a try-catch around this whole block? I guess you wouldn't want you whole CICD to fail because this analysis fails

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
```diff
@@ -0,0 +1,81 @@
+$errorLogsFolder = Join-Path $ENV:GITHUB_WORKSPACE "ErrorLogs"
+
+# Base SARIF structure
+$sarif = @{
```
aholstrup1: Minor: Put this into a separate .json file and load it during the script.

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
spetersenms: Good idea!

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
spetersenms: Agreed, that makes it a bit cleaner.

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
spetersenms: Yep that's better. I added a try-catch block, with a warning in the catch block to manually inspect any potential alerts.

Templates/AppSource App/.github/workflows/CICD.yaml
spetersenms: Good idea. I added a new output in the Init job with the new setting and used that in the job condition.

Actions/CalculateArtifactNames/action.yaml
```diff
@@ -46,6 +46,9 @@ outputs:
   ContainerEventLogArtifactsName:
     description: Artifacts name for ContainerEventLog
     value: ${{ steps.calculateartifactnames.outputs.ContainerEventLogArtifactsName }}
+  ErrorLogsArtifactsName:
```
mazhelez, line 49: The README.md for the action needs to be updated as well.

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
```diff
@@ -0,0 +1,84 @@
...
+            }
+        }
+
+        # Convert absolute path to relative path from repository root
+        $absolutePath = $issue.locations[0].analysisTarget[0].uri
+        $workspacePath = $ENV:GITHUB_WORKSPACE
+        $relativePath = $absolutePath.Replace($workspacePath, '').TrimStart('\').Replace('\', '/')
+
+        # Add result
+        $sarif.runs[0].results += @{
```
mazhelez: Sometimes, there are multiple build jobs that build the same app (in different modes, for examples).
It's annoying when there are issues that are the same but reported from different jobs. Can this be handled in some way?

mazhelez: Alternatively, can `trackCodeCopIssuesInGitHub` be a project setting instead?

Templates/AppSource App/.github/workflows/CICD.yaml
```diff
@@ -197,6 +205,37 @@ jobs:
       useArtifactCache: true
       needsContext:  ${{ toJson(needs) }}
 
+  CodeAnalysisUpload:
```
mazhelez, line 208: Should `PostProcess` depend on this job?

Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
```diff
@@ -107,8 +115,43 @@ jobs:
...
+        uses: github/codeql-action/upload-sarif@v3
+        if: always()
+        with:
+          sarif_file: '${{ github.workspace }}/ErrorLogs/output.sarif.json'
+          category: "ALCodeAnalysis"
+          ref: ${{ format('refs/pull/{0}/head', github.event.pull_request.number) }}
+          sha: ${{ github.event.pull_request.head.sha }}
+
+  StatusCheck:
+    needs: [ Initialization, Build, CodeAnalysisUpload ]
```
mazhelez: Hm... I'm actually not sure if this is a good idea.
If the code analysis upload fails for whatever reason, then the whole build would fail.

Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
```diff
@@ -66,7 +68,13 @@ jobs:
...
         with:
           shell: powershell
-          get: shortLivedArtifactsRetentionDays
+          get: shortLivedArtifactsRetentionDays,trackCodeCopIssuesInGitHub
+
+      - name: Set AL Code Analysis Var output
+        id: SetALCodeAnalysisVar
+        run: |
+          Write-Host "trackCodeCopIssuesInGitHub environment variable: '$($env:trackCodeCopIssuesInGitHub)'"
+          Add-Content -Encoding UTF8 -Path $env:GITHUB_OUTPUT -Value "trackCodeCopIssuesInGitHub=$($env:trackCodeCopIssuesInGitHub)"
```
mazhelez: How about making this directly as an output from `ReadSettings`?

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
spetersenms: Currently, the alerts have no context of which project / job they came from, but we can probably just solve this with some logic to de-duplicate alerts.

Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
spetersenms: We could do that, but I chose this approach since it's mostly limited where we need this output, so I did not want to clutter the readsettings action more logic and outputs. It's also the same way we set the ArtifactsRetentionDays output.

spetersenms: Discussed offline. Will keep this approach in this PR and restructure the way outputs from ReadSettings are handled in a different PR.

Templates/AppSource App/.github/workflows/CICD.yaml
spetersenms: You are right, it should.

Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
spetersenms: Good point. I removed the dependency from StatusCheck.

mazhelez (dismissed):

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
```diff
@@ -0,0 +1,102 @@
+Param(
```
aholstrup1, line 1: LGTM! 🚀 Really nice improvement! 

I wonder if we should wait to merge it until we've shipped v7.3? Just so we don't risk regressions and we give this a bit of time to cook in preview before we deploy it to an official version. 

Feel free to close the comment if you disagree 👍

aholstrup1 (dismissed):

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
spetersenms: Yeah I agree, we should have this in preview for a while.

RELEASENOTES.md
```diff
@@ -13,6 +13,14 @@ Example
 }
 ```
 
+### AL Code Analysis tracked in GitHub
```

aholstrup1 (dismissed):

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
```diff
@@ -0,0 +1,102 @@
...
+                OutputDebug -message "Error: $_"
+            }
+        }
+
+        $sarifJson = $sarif | ConvertTo-Json -Depth 10 -Compress
+        OutputDebug -message $sarifJson
+        Set-Content -Path "$errorLogsFolderPath/output.sarif.json" -Value $sarifJson
+    }
+    else {
+        OutputWarning -message "ErrorLogs folder not found. You can manually inspect your artifacts for AL code alerts."
```
mazhelez: Why would that happen? What would be the solution for this warning?

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
```diff
@@ -0,0 +1,102 @@
...
+    }
+}
+catch {
+    OutputWarning -message "Unexpected error processing AL code analysis results. You can manually inspect your artifacts for AL code alerts."
```
mazhelez, line 100: Should something be traced in AL-Go telemetry?

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
spetersenms: Main reason would be if you use the param with the folder not existing. Even if this step fails, your artifacts would still exist with potential alerts, so the purpose of the warning is to let people know to manually inspect their alerts and not blindly merge their PR.

mazhelez: Ok. That makes sense. But maybe at least add the folder name in the warning.

mazhelez:
```suggestion
        OutputWarning -message "ErrorLogs $errorLogsFolder folder not found. You can manually inspect your artifacts for AL code alerts."
```

Templates/AppSource App/.github/workflows/_BuildALGoProject.yaml
```diff
@@ -108,7 +108,7 @@ jobs:
           shell: ${{ inputs.shell }}
           project: ${{ inputs.project }}
           buildMode: ${{ inputs.buildMode }}
-          get: useCompilerFolder,keyVaultCodesignCertificateName,doNotSignApps,doNotRunTests,doNotRunBcptTests,doNotRunpageScriptingTests,artifact,generateDependencyArtifact,trustedSigning,useGitSubmodules
+          get: useCompilerFolder,keyVaultCodesignCertificateName,doNotSignApps,doNotRunTests,doNotRunBcptTests,doNotRunpageScriptingTests,artifact,generateDependencyArtifact,trustedSigning,useGitSubmodules,trackCodeCopIssuesInGitHub
```
mazhelez:
```suggestion
          get: useCompilerFolder,keyVaultCodesignCertificateName,doNotSignApps,doNotRunTests,doNotRunBcptTests,doNotRunpageScriptingTests,artifact,generateDependencyArtifact,trustedSigning,useGitSubmodules,trackALAlertsInGithub
```

Templates/AppSource App/.github/workflows/CICD.yaml
```diff
@@ -197,6 +205,37 @@ jobs:
...
+    needs: [ Initialization, Build ]
+    if: (!cancelled()) && (needs.Initialization.outputs.trackALAlertsInGithub == 'True')
+    runs-on: [ windows-latest ]
+    name: Code Analysis Processing
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
mazhelez, line 215: Is the checkout here really needed?

Templates/AppSource App/.github/workflows/PullRequestHandler.yaml
```diff
@@ -107,6 +115,41 @@ jobs:
       needsContext:  ${{ toJson(needs) }}
       useArtifactCache: true
 
+  CodeAnalysisUpload:
+    needs: [ Initialization, Build ]
+    if: (!cancelled()) && (needs.Initialization.outputs.trackALAlertsInGithub == 'True')
+    runs-on: [ windows-latest ]
+    name: Code Analysis Processing
+    steps:
+      - name: Checkout
```
mazhelez, line 124: Same here. I can't figure out why the checkout is needed.

spetersenms: During checkout, a diff file is generated behind the scenes with data on which files changed in the PR. This is needed by the SARIF upload in order to actually post any alerts in the PR. Without this, alerts would only go to the security tab.

Templates/AppSource App/.github/workflows/CICD.yaml
spetersenms: Here, SARIF upload is possible without checkout, but it is recommended to still do it. Part of the SARIF file uses the checkout path to correctly map alerts to files in your repo. This can technically work without checkout, but manual checkout enforces consistency, as the SARIF upload would have to "guess" the checkout path without.

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
spetersenms: Added a Trace-Exception in the latest commit, is it something like that you were thinking?

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
```diff
@@ -93,10 +93,11 @@ try {
...
     }
     else {
-        OutputWarning -message "ErrorLogs folder not found. You can manually inspect your artifacts for AL code alerts."
+        OutputWarning -message "ErrorLogs $errorLogsFolder folder not found. You can manually inspect your artifacts for AL code alerts."
     }
 }
 catch {
     OutputWarning -message "Unexpected error processing AL code analysis results. You can manually inspect your artifacts for AL code alerts."
     OutputDebug -message "Error: $_"
+    Trace-Exception -Message "Unexpected Error" -ActionName "ProcessALCodeAnalysisLogs" -ErrorRecord $_
```
aholstrup1:
```suggestion
    Trace-Exception -ActionName "ProcessALCodeAnalysisLogs" -ErrorRecord $_
```

If you log it like this, it will follow the same "naming convention" as other places

aholstrup1 (approved):

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
mazhelez: `Trace-Exception` is exactly what I had in mind.

mazhelez (dismissed):

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
spetersenms: Alright, I'll make the change. We might want to edit the param for Trace-Exception though as it marks $Message as Mandatory.

PR #1882, opened 2025-08-25
ReadSettings output JSON structured based on get input parameter.

spetersenms:
### ❔What, Why & How

This feature changes how we handle output variables from ReadSettings. The current solution is to add a setting to the get parameter of ReadSettings, and then add a step like this:
```
- name: Add Output From Setting
   id: AddOutputFromSetting
   run: |
     Add-Content -Encoding UTF8 -Path $env:GITHUB_OUTPUT -Value "SettingName=$($env:settingName)"
```

This is then added to the outputs of a job like this:
```
Initialization:
  needs: [ PregateCheck ]
  if: (!failure() && !cancelled())
  runs-on: [ windows-latest ]
  outputs:
    <...>
    SettingName: ${{ steps.AddOutputFromSetting.outputs.SettingName }}
  steps:
    <...>
```

This new solution takes all settings selected in the get parameter of ReadSettings and puts them in a single JSON structure. This can then be used directly in the job outputs like this:
```
Initialization:
  needs: [ PregateCheck ]
  if: (!failure() && !cancelled())
  runs-on: [ windows-latest ]
  outputs:
    <...>
    SettingName: ${{ fromJson(steps.ReadSettings.outputs.SelectedSettingsJson).settingName }}
  steps:
    <...>
```

### ✅ Checklist

- [ ] Update RELEASENOTES.md
- [ ] Add telemetry

### DIFF
--- a/Actions/.Modules/ReadSettings.psm1
+++ b/Actions/.Modules/ReadSettings.psm1
@@ -310,7 +310,7 @@ function ReadSettings {
 
     $repoName = $repoName.SubString("$repoName".LastIndexOf('/') + 1)
     $githubFolder = Join-Path $baseFolder ".github"
-    $workflowName = $workflowName.Trim().Split([System.IO.Path]::getInvalidFileNameChars()) -join ""
+    $workflowName = SanitizeWorkflowName -workflowName $workflowName
 
     # Start with default settings
     $settings = GetDefaultSettings -repoName $repoName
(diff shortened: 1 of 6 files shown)


Templates/Per Tenant Extension/.github/workflows/CICD.yaml
```diff
@@ -44,8 +44,8 @@ jobs:
       buildOrderJson: ${{ steps.determineProjectsToBuild.outputs.BuildOrderJson }}
       baselineWorkflowRunId: ${{ steps.determineProjectsToBuild.outputs.BaselineWorkflowRunId }}
       baselineWorkflowSHA: ${{ steps.determineProjectsToBuild.outputs.BaselineWorkflowSHA }}
-      workflowDepth: ${{ steps.DetermineWorkflowDepth.outputs.WorkflowDepth }}
-      powerPlatformSolutionFolder: ${{ steps.DeterminePowerPlatformSolutionFolder.outputs.powerPlatformSolutionFolder }}
+      workflowDepth: ${{ env.workflowDepth }}
+      powerPlatformSolutionFolder: ${{ fromJson(steps.ReadSettings.outputs.SelectedSettingsJson).powerPlatformSolutionFolder }}
```
aholstrup1, line 48: If we can just assign workflowDepth to ${{ env.workflowDepth }} couldn't we do the same for powerPlatformSolutionFolder. 

The settings defined in the "get" parameter are already assigned to an environment variable AFAIK

spetersenms: This only works for workflowDepth because it's a global environment variable defined on line 23. The environment variables created by readsettings for things in the get param can't be used for job outputs, hence the need for putting everything in "get" into a separate output variable which can then be used for the job output.

That said, it seems the workflowDepth output is not actually used. In this PR I kept it for now, but we can probably remove it completely as the whole workflow should have access to env.workflowDepth given it's defined on the global level.

PR #1887, opened 2025-08-27
Sanitize the workflow name provided in conditional settings

aholstrup1:
### ❔What, Why & How

Related to: #1885 

In AL-Go it is possible to define conditional settings based on workflows. In the Read Settings action we will look if the workflow name of the current workflow matches a workflow name defined in the conditional settings. 

However, right now we sanitize the workflow name of the current workflow so AL-Go users would have to add the sanitized workflow name to their settings rather than the actual workflow name e.g. CI/CD doesn't work but CICD does. 

We should make that a bit easier so AL-Go users can just specify the name of the workflow. 

### ✅ Checklist

- [x] Update RELEASENOTES.md


mazhelez (approved):

spetersenms (approved):

RELEASENOTES.md
```diff
@@ -48,6 +48,7 @@ Please note, that due to certain security limitations, the properties `runs-on`,
 - Issue 1837 Deployment from PR builds fail if PR branch name includes forward slashes (e.g., `feature/branch-name`).
 - Issue 1852 Page Scripting Tests are not added to build summary
 - Issue 1829 Added custom jobs cannot be removed
+- Discussion 1885 Conditional settings for CI/CD are not applied
```

PR #1923, opened 2025-09-24
Remove required list in settings.schema.json

mazhelez:
### ❔What, Why & How

Remove required list in settings.schema.json.
Details on the issue at: #1899

### ✅ Checklist

- [X] Update RELEASENOTES.md


cegekaJG:
With Copilot, it's important to remove inaccurate problems. If someone were to use Copilot to troubleshoot their AL Go project, it would likely use the warnings generated by the missing required properties first to propose a solution.

mazhelez:
> This is good fix, on short term. 👍 One could argue for implementing the required checks later.

For settings files the `required` properties doesn't make sense (as agreed in the linked discussion).

For validating the settings after _ReadSettings_ actions, there is value (especially after #1924). However, there was a suggestion to make _ReadSettings_ work the same way as _ReadSecrets_, i.e. not all settings are read all the time, only the ones that are needed in the current job. As far as I can recall, this is because someone noticed half of the AL-Go logs are output settings 😃

jwikman:
> For validating the settings after _ReadSettings_ actions, there is value (especially after #1924). However, there was a suggestion to make _ReadSettings_ work the same way as _ReadSecrets_, i.e. not all settings are read all the time, only the ones that are needed in the current job. As far as I can recall, this is because someone noticed half of the AL-Go logs are output settings 😃

Sounds as a good way forward. Then we only need to verify that the requested settings do exist. 👍

spetersenms (approved):

PR #1924, opened 2025-09-24
Add mechanism to overwrite complex settings type

mazhelez:
### ❔What, Why & How

By default, AL-Go merges settings from various places (see [settings levels](http://aka.ms/algosettings#where-are-the-settings-located)). Basic setting types such as `string` and `integer` are overwritten, but settings with complex types such as `array` and `object` are merged.

However, sometimes it is useful to avoid merging complex types. 
This PR adds support for `overwriteSettings` property on a settings object. The purpose of the property is to list settings, for which the value will be overwritten, instead of merged.

Related to discussion: #1913

### ✅ Checklist

- [x] Update RELEASENOTES.md
- No telemetry. Note: telemetry cannot be easily added in the process of reading settings.


Scenarios/settings.md
```diff
@@ -232,6 +232,73 @@ Please read the release notes carefully when installing new versions of AL-Go fo
...
+    "country": "de"
+    "buildModes": ["Default"]
+}
+```
+
+and `.github\AL-Go-Settings.json` contains the following values
+
+```json
+{
+    "resetSettings": "buildModes"
```
jwikman: `resetSettings` should be an array, shouldn't it?
```suggestion
    "resetSettings": ["buildModes"]
```

mazhelez: Thanks! Good catch!

Actions/.Modules/ReadSettings.psm1
```diff
@@ -10,11 +10,27 @@ function MergeCustomObjectIntoOrderedDictionary {
         [PSCustomObject] $src
     )
 
+    # If the src object contains property 'resetSettings' (list of settings), remove these settings from the dst object
```
aholstrup1: This doesn't work for e.g. an array inside an object, right? E.g. 
In the original example, there is a branches array inside the DeliverToGitHubPackages object. You wouldn't be able to use resetSettings _only_ on the branches array, right? You'd have to reset the whole structure? 

```
    "DeliverToGitHubPackages": {
        "ContinuousDelivery": true,
        "Branches": [
            "main",
            "releases/*"
        ]
    }
}
```

I think this is fine btw. Just maybe worth adding a note of resetSettings in the docs.

aholstrup1 (dismissed):

Actions/.Modules/ReadSettings.psm1
jwikman: @aholstrup1 this would work, wouldn't it? At least that was what I expected...
```
    "DeliverToGitHubPackages": {
        "resetSettings": ["Branches],
        "ContinuousDelivery": true,
        "Branches": [
            "main",
            "releases/*"
        ]
    }
```

aholstrup1: After testing, it appears it does yes :) My mistake

jwikman: Phew! 😅

mazhelez: Thanks for testing, Alex.
Your comment prompt me to check and adjust the settings schema accordingly.

Also, unrelated, I decided that `overwriteSettings` match better the intention of the new property.

jwikman:
I think that the "resetSetting" name was better... Because that is what it does.
Then you _can_ use this to overwrite/override settings, but that is optional.

mazhelez:
> I think that the "resetSetting" name was better... Because that is what it does. Then you _can_ use this to overwrite/override settings, but that is optional.

That's the other new thing I added: if a setting is added in `overwriteSettings` but the settings isn't defined with the new value, then the setting's value isn't reset, it stays the same.
This is to avoid removing settings, without providing a new value. If the logic allows removing settings, there might be side effects. AL-Go relies on the settings being there.

jwikman:
> > I think that the "resetSetting" name was better... Because that is what it does. Then you _can_ use this to overwrite/override settings, but that is optional.
> 
> 
> 
> That's the other new thing I added: if a setting is added in `overwriteSettings` but the settings isn't defined with the new value, then the setting's value isn't reset, it stays the same.
> 
> This is to avoid removing settings, without providing a new value. If the logic allows removing settings, there might be side effects. AL-Go relies on the settings being there.

I don't think I like that change...
So now we cannot revert to default settings?

It is the responsibility of the one who configuring settings that it works. With this change it only get less flexibility.

mazhelez:
> I don't think I like that change... So now we cannot revert to default settings?
> 
> It is the responsibility of the one who configuring settings that it works. With this change it only get less flexibility.

There must have been some sort of confusion on my side. The ask was never to revert to default settings, was it?
From the disccussion #1913 :
> Hence, I suggest implementing a mechanism where we also can override array and object settings.

With the new `overwriteSettings`, compex settings (arrays and objects) can be overwritten the same way as simple type settings.

jwikman:
> There must have been some sort of confusion on my side. The ask was never to revert to default settings, was it? From the disccussion #1913 :
> 
> > Hence, I suggest implementing a mechanism where we also can override array and object settings.

Yes, true. But later in that discussion I wrote this: 

> I first thought about this with an "OverrideProperties" setting, but removing setting from a higher level would also open up for totally removing a setting, not only override with new values.

That was why I landed on the `RemoveProperties` approach (but your naming with `resetSettings` is probably better).

I thought your first implementation was simple, flexible, straight forward and just brilliant. 🙂

mazhelez:
> Yes, true. But later in that discussion I wrote this:
> 
> > I first thought about this with an "OverrideProperties" setting, but removing setting from a higher level would also open up for totally removing a setting, not only override with new values.
> 
> That was why I landed on the `RemoveProperties` approach (but your naming with `resetSettings` is probably better).

I guess maybe I failed to understand the scenario in which you'd want a setting to be removed altogether.
The initial implementation of `resetSettings` wasn't _emptying_ the setting, it was removing it from the settings object, and there was no verification that the settings was added back. The way I see it, this can only lead to issues (e.g. if a setting is removed, because someone thought it wasn't used, but it ended up being used in a subtle way).

jwikman:
> I guess maybe I failed to understand the scenario in which you'd want a setting to be removed altogether. The initial implementation of `resetSettings` wasn't _emptying_ the setting, it was removing it from the settings object, and there was no verification that the settings was added back. The way I see it, this can only lead to issues (e.g. if a setting is removed, because someone thought it wasn't used, but it ended up being used in a subtle way).

And when a setting was removed on some level, it can be added back on a lower level, or, if not added back, it would default to AL-Go default values. Or am I wrong?

I don't think that your change to `overwriteSettings` will help anyone to not mess up their settings if they do not know what they are doing. To almost achieve the `removeSettings` approach, the user can still add an empty array or object. And still mess things up.
The only thing achieved with the change to `overwriteSettings` is that we cannot remove a setting completely, reverting to AL-Go default. That also applies to simple types.

I rather see a feature that clearly can show the user how the settings object is built when running a pipeline. Maybe output to pipeline when running in verbose mode? In that way we help the user to troubleshoot if they messed up their settings. 🙂

mazhelez:
> And when a setting was removed on some level, it can be added back on a lower level, or, if not added back, it would default to AL-Go default values. Or am I wrong?

You are indeed mistaken here. The default values are the initial settings. The settings are then applied (merged or overwritten) level by level (org settings, repo settings, project settings etc.). But if a setting is removed (via `resetSettings`) at certain point and it is not re-added later on, it's just gone from the result object.

> I don't think that your change to `overwriteSettings` will help anyone to not mess up their settings if they do not know what they are doing. To almost achieve the `removeSettings` approach, the user can still add an empty array or object. And still mess things up. The only thing achieved with the change to `overwriteSettings` is that we cannot remove a setting completely, reverting to AL-Go default. That also applies to simple types.

I agree it's not that hard to mess up settings. My approach is not to make it even easier 😄 

> I rather see a feature that clearly can show the user how the settings object is built when running a pipeline. Maybe output to pipeline when running in verbose mode? In that way we help the user to troubleshoot if they messed up their settings. 🙂

That's a good point! There is some logging already, but it can definitely be improved. Let me tackle that in a separate PR though.

jwikman:
> You are indeed mistaken here. The default values are the initial settings. The settings are then applied (merged or overwritten) level by level (org settings, repo settings, project settings etc.). But if a setting is removed (via `resetSettings`) at certain point and it is not re-added later on, it's just gone from the result object.

Oh, I see - my bad. NOW I finally understand why you were doing this change! 😅

One solution, though, would be to instead add missing, and only missing, default values afterwards. Adding them first felt a bit strange IMHO, but probably had its reasons. (Maybe because that was the easiest way to implement default values? 🙂)
That being said, I assume that you are not into changing that behavior? 🤔


> That's a good point! There is some logging already, but it can definitely be improved. Let me tackle that in a separate PR though.

A separate PR sounds as a very good idea! 🙂

mazhelez:
> Oh, I see - my bad. NOW I finally understand why you were doing this change! 😅

Glad I was finally clear enough 😅

> One solution, though, would be to instead add missing, and only missing, default values afterwards. Adding them first felt a bit strange IMHO, but probably had its reasons. (Maybe because that was the easiest way to implement default values? 🙂) That being said, I assume that you are not into changing that behavior? 🤔

Your assumption is correct at this point. I actually did give a thought about _resetting a setting to its default value_ but it's the same as _overwriting it with its default value_.
My decision to keep it simple was also affected by https://github.com/microsoft/AL-Go/discussions/1854 where the ask is to support _custom settings schema_. It means that some repos will be using custom settings and `ReadSettings` action wouldn't know about the default values of such settings, at least not by default.

jwikman:
> My decision to keep it simple was also affected by #1854 where the ask is to support _custom settings schema_. It means that some repos will be using custom settings and `ReadSettings` action wouldn't know about the default values of such settings, at least not by default.

I see. Then I'll stop bugging you on this, please go ahead and merge 🙂

elbardel (approved):

y0m0:
Hi @mazhelez, I am having some issues with overwriteSettings not working as I expect it to

In my ALGoOrgSettings i have the following
``` 
  "CICDPullRequestBranches": [
    "main",
    "master"
  ],
  "CICDPushBranches": [
    "main",
    "master",
    "release/*"
  ],
```

And in a project repository i have the following in my AL-Go-Settings.json
``` 
  "overwriteSettings": [
    "CICDPullRequestBranches",
    "CICDPushBranches"
  ],
  "CICDPullRequestBranches": [
    "__disabled__"
  ],
  "CICDPushBranches": [
    "__disabled__"
  ]
```

Then I commit and push my change to my repo
Then I run "Update AL-Go System Files"

and here's the result from the PR

image

what am I doing wrong here?

mazhelez:
First off, are you running on the latest preview version?

If yes, can you run the AL-Go update in debug mode to gather more info?

y0m0:
My bad, I had templateUrl set to @main. I changed it to point to @preview instead of @main and it's now working as expected.
Thank you.

PR #1942, opened 2025-10-01
Track AL Alerts in GitHub - Check if sarif file was produced and standardize casing 

aholstrup1:
### ❔What, Why & How

Related to issue: AB#1937

This PR does two things:
* It adds an additional check in the upload-sarif step that ensures the file exists before trying to upload it
* It standardizes the casing of the setting. trackALAlertsInGit**hub** -> trackALAlertsInGit**Hub**


Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
```diff
@@ -13,6 +13,73 @@ if (Test-Path $sarifPath) {
...
+    .DESCRIPTION
+    Given an absolute path, this function searches the workspace for a file with the same name and returns its relative path.
+    If multiple files with the same name are found, it returns the one with the longest matching suffix to the absolute path.
+    If no file is found, it returns $null.
+    .PARAMETER AbsolutePath
+    The absolute path of the file to find.
+    .PARAMETER WorkspacePath
+    The workspace path to search in. Defaults to the GITHUB_WORKSPACE environment variable.
+#>
+function Get-FileFromAbsolutePath {
```
mazhelez: I'd really appreciate a couple of tests for this function.

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
```diff
@@ -29,36 +96,57 @@ function GenerateSARIFJson {
...
-        })) {
-            $sarif.runs[0].results += @{
+            $_.message.text -eq $message -and
+            $_.level -eq ($issue.properties.severity).ToLower() -and
+            ($_.locations[0].physicalLocation.artifactLocation.uri -eq $relativePath) -and
+            ($_.locations[0].physicalLocation.region | ConvertTo-Json) -eq ($issue.locations[0].analysisTarget[0].region | ConvertTo-Json)
+        }
+
+        # Add result if it does not already exist
+        if (-not $existingResults)
```
mazhelez: Minor: I think the following fits better in this function:

```
if ($existingResults) {
    # Skip if existing
    continue
}
```

That way you can remove the final `if ($null -ne $newResult)` check as well.

mazhelez (dismissed):
Looks good! More tests would be useful.

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
aholstrup1: That's also exactly why this PR is still in draft 👍 More tests coming!

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.psm1
```diff
@@ -0,0 +1,88 @@
...
+        # If the file exists at the absolute path, return the relative path from the workspace root
+        return NormalizePath -Path ([System.IO.Path]::GetRelativePath($workspacePath, $normalizedPath))
+    }
+
+    # Extract the file name from the absolute path
+    $fileName = [System.IO.Path]::GetFileName($normalizedPath)
+
+    # Search the workspace path for a file with that name
+    $matchingFiles = @(Get-ChildItem -Path $WorkspacePath -Filter $fileName -File -Recurse -ErrorAction SilentlyContinue)
+    if($matchingFiles.Count -eq 1) {
```
spetersenms, line 94: This does not seem to match the description. Description indicates that it's ok to have multiple files, but here we only continue if a single file is found.

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
```diff
@@ -25,11 +26,48 @@ if (Test-Path $sarifPath) {
...
+            ($_.locations[0].physicalLocation.artifactLocation.uri -eq $relativePath) -and
+            ($_.locations[0].physicalLocation.region | ConvertTo-Json) -eq ($issue.locations[0].analysisTarget[0].region | ConvertTo-Json)
+        }
+
+        if ($existingResults) {
+            # Skip if existing
+            continue
+        }
+
+        # Add rule to the sarif object if not already added
```
spetersenms, line 70: Should we use the new `Get-IssueMessage` for the `shortDescription` and `fullDescription` of the rule too? If so, the function might have to be modified with a parameter to select if the short or full message is more appropriate.

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.psm1
aholstrup1: I'll update the description. It used to be in there, but I think for this release let's start without it.

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
aholstrup1: I still think we should keep the full description if it is there. I've made a small adjustment. Can you take another look?

PR #1957, opened 2025-10-06
Make ReadSettings less chatty

mazhelez:
### ❔What, Why & How

Make `ReadSettings` less chatty as it outputs a lot of information that is mainly needed for debugging.
This PR also corrects the order of the logs for applying the settings.

Related to comment https://github.com/microsoft/AL-Go/pull/1924#issuecomment-3338567205

### ✅ Checklist

- [x] Update RELEASENOTES.md

### DIFF
--- a/Actions/ReadSettings/ReadSettings.ps1
+++ b/Actions/ReadSettings/ReadSettings.ps1
@@ -81,7 +81,7 @@ $settings.Keys | ForEach-Object {
     }
 }
 
-Write-Host "AL-Go settings: $($outSettings | ConvertTo-Json -Depth 99 -Compress)"
+Write-Host "AL-Go settings: $($outSettings | ConvertTo-Json -Depth 99)"
 
 Add-Content -Encoding UTF8 -Path $env:GITHUB_ENV -Value "Settings=$($outSettings | ConvertTo-Json -Depth 99 -Compress)"
 


Actions/ReadSettings/ReadSettings.ps1
```diff
@@ -74,8 +81,8 @@ $settings.Keys | ForEach-Object {
     }
 }
 
-Write-Host "SETTINGS:"
-$outSettings | ConvertTo-Json -Depth 99 | Out-Host
+OutputDebug "Final settings: $($outSettings | ConvertTo-Json -Depth 99 -Compress)"
```
aholstrup1: Tbh I think this is pretty useful to have in the logs. I would personally prefer to keep it

mazhelez: The settings are set as an environment variable, so they are visible on every action that runs after `ReadSettings` and that is more important to see (in case something has meddled with the settings)
E.g. https://github.com/microsoft/BCApps/actions/runs/18342944499/job/52242672202#step:5:50

It's really just a lot of duplication around the logs.

aholstrup1: This PR won't get rid of that duplication, I guess? I think having one nicely formatted copy of the settings is nice
https://github.com/microsoft/BCApps/actions/runs/18342944499/job/52242672202#step:3:37

It's for sure a nice-to-have and not a need-to-have. 

My preference would be to keep it but if you feel strongly for cleaning this up from the logs we can also do that :)

mazhelez: My preference isn't that strong, so reverted back to outputting all settings, nicely formatted.

spetersenms (approved):

PR #1969, opened 2025-10-09
Run CI action tests on both PowerShell 5 and PowerShell 7

aholstrup1:
Run CI action tests on both PowerShell 5 and PowerShell 7

Fixes [AB#609581](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/609581)

### DIFF
--- a/Actions/.Modules/ReadSettings.psm1
+++ b/Actions/.Modules/ReadSettings.psm1
@@ -1,3 +1,5 @@
+Import-Module (Join-Path -Path $PSScriptRoot "DebugLogHelper.psm1")
+
 $ALGoFolderName = '.AL-Go'
 $ALGoSettingsFile = Join-Path '.AL-Go' 'settings.json'
 $RepoSettingsFile = Join-Path '.github' 'AL-Go-Settings.json'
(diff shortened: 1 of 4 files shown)


.github/workflows/CI.yaml
```diff
@@ -7,16 +7,25 @@ on:
...
+        include:
+          - os: windows-latest
+            shell: powershell
+            psVersion: PS5
+          - os: ubuntu-latest
+            shell: pwsh
+            psVersion: PS7
+      fail-fast: false
+    runs-on: ${{ matrix.os }}
+    defaults:
```
mazhelez, line 27:
```suggestion
    name: Run tests on ${{ matrix.os }} (${{ matrix.psVersion }})
    defaults:
```

.github/workflows/CI.yaml
```diff
@@ -26,16 +35,16 @@ jobs:
...
-      - name: Run AL-Go Actions Tests
+      - name: Run AL-Go Actions Tests (${{ matrix.psVersion }})
         run: |
           . (Join-Path "." "Tests/runtests.ps1") -Path "Tests"
 
-      - name: Test AL-Go Workflows
+      - name: Test AL-Go Workflows (${{ matrix.psVersion }})
         if: github.repository_owner == 'microsoft'
         run: |
           . (Join-Path "." "Tests/runtests.ps1") -Path "Tests/WorkflowSanitation"
```
mazhelez, line 46: Minor: "Test AL-Go Workflows" and "Test aka.ms links in markdown files" don't need to run on all configurations. These might be better off in a job of their own.

aholstrup1: Yeeeah I mean that is true. On the other hand, they're so quick to run that it would likely take longer if we had to spin up another runner. 
I'd keep them there for now but if you want, we can also move them out.

PR #1978, opened 2025-10-14
Add support for reportSuppressedDiagnostics

jwikman:
### ❔What, Why & How

- Implement the idea in discussion: #1911 
- Add a fix for the schema locations for several settings files. I guess the location was changed at some point.

### ✅ Checklist

- [X] Update RELEASENOTES.md
- [X] Update documentation (e.g. for new settings or scenarios)
- [ ] Add telemetry

### DIFF
--- a/Actions/.Modules/ReadSettings.psm1
+++ b/Actions/.Modules/ReadSettings.psm1
@@ -240,7 +240,7 @@ function GetDefaultSettings
         "useGitSubmodules"                              = "false"
         "gitSubmodulesTokenSecretName"                  = "gitSubmodulesToken"
         "shortLivedArtifactsRetentionDays"              = 1  # 0 means use GitHub default
-        "reportSuppressedDiagnostics"                   = $false;
+        "reportSuppressedDiagnostics"                   = $false
     }
 }
 


jwikman:
@mazhelez I just discovered that `Run-AlPipeline` in BcContainerHelper do _not_ support the `reportSuppressedDiagnostics` parameter 🤦‍♂️

And I cannot find any generic way of passing that parameter, without overriding the compile script. But that does not feel right...

Would it make sense to add the parameter to BCH?

mazhelez:
> @mazhelez I just discovered that `Run-AlPipeline` in BcContainerHelper do _not_ support the `reportSuppressedDiagnostics` parameter 🤦‍♂️
> 
> And I cannot find any generic way of passing that parameter, without overriding the compile script. But that does not feel right...
> 
> Would it make sense to add the parameter to BCH?

Yes, that would make sense.

jwikman:
> > Would it make sense to add the parameter to BCH?
> 
> Yes, that would make sense.

Ok, good. I'll start in that end instead then! :)

jwikman:
Now https://github.com/microsoft/navcontainerhelper/pull/4025 is ready for review.
And then this one _should_ work.

Actions/.Modules/ReadSettings.psm1
```diff
@@ -240,6 +240,7 @@ function GetDefaultSettings
         "useGitSubmodules"                              = "false"
         "gitSubmodulesTokenSecretName"                  = "gitSubmodulesToken"
         "shortLivedArtifactsRetentionDays"              = 1  # 0 means use GitHub default
+        "reportSuppressedDiagnostics"                   = $false;
```
aholstrup1:
```suggestion
        "reportSuppressedDiagnostics"                   = $false
```

aholstrup1 (commented):
One small comment but other than that, it looks great! Let's see if we can get this shipped in preview soon 🚀

Actions/.Modules/ReadSettings.psm1
jwikman: Must have been my AL genes that added that semi colon! 😅

PR #1994, opened 2025-10-22
Refactor Deployment settings & logic in Deploy and DetermineDeploymentEnvironments actions.

spetersenms:
### ❔What, Why & How

Refactor that puts all handling of DeployTo settings into the ReadSettings action instead of letting Deploy and DetermineDeploymentEnvironments handle it.

Related to issue: #

### ✅ Checklist

- [ ] Update RELEASENOTES.md
- [ ] Add telemetry


Actions/.Modules/ReadSettings.psm1
```diff
@@ -549,6 +549,64 @@ function ReadSettings {
         $settings.projectName = $project # Default to project path as project name
     }
 
+    $ghEnvironments = @(GetGitHubEnvironments)
```
mazhelez: This will be called on every `ReadSettings` call, right? Given that `GetGitHubEnvironments` makes an HTTP call, it seems excessive and unnecessary.

PR #2010, opened 2025-11-07
Add copilot instructions

jonasblunck:
Add basic copilot instructions that makes it possible to ask Copilot about settings.


Templates/AppSource App/.github/.agents/algo-settings.agent.md
```diff
@@ -0,0 +1,9 @@
+# AL-Go Copilot instructions
```
aholstrup1, line 1: Putting it here would override any copilot instructions the owner already has. I don't think we should do that 

Maybe we can make it path-specific instead so it can be its own separate file under the `.github/instructions` folder
https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions?tool=vscode#creating-path-specific-custom-instructions-1

jwikman: Exactly my thought when I saw this PR.

Please do not add these files with the default names, as we already have them in our App repos.

spetersenms (approved):

PR #2018, opened 2025-11-12
Refactor branch fetching logic to exclude 'origin' from remote branches

### LINKED ISSUE #2016: [Bug]: Running Update AL-Go system files with branches wildcard `*` tries to update `origin`
### AL-Go version

8.0

### Describe the issue

When running "Update AL-Go system files" with `*` as "list of branches to update" it tries to update `origin` which fails.

### Expected behavior

When running "Update AL-Go system files" with `*` as "list of branches to update" it should filter out `origin`.

### Steps to reproduce

1. Run "Update AL-Go System Files" with `*` as branches:
    Image
2. See that the workflow fails because it tries to update `origin`:
    Image

The step "Get Workflow Multi-Run Branches" within the "Initialize" job should likely filter out `origin`:

Image

Logs:
[logs_49612972209.zip](https://github.com/user-attachments/files/23472532/logs_49612972209.zip)
Internal work item: [AB#612685](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/612685)

mazhelez:
This pull request addresses an issue where using a wildcard (`*`) for branch selection in the workflow update process incorrectly included the `origin`. The changes ensure that only valid remote branches are considered, excluding the `origin` reference itself.

Bug fix for branch filtering:

* Updated `GetWorkflowMultiRunBranches.ps1` to exclude the `origin` reference and correctly strip the `origin/` prefix from remote branch names when listing branches.

Fixes #2016

### DIFF
--- a/Tests/GetWorkflowMultiRunBranches.Test.ps1
+++ b/Tests/GetWorkflowMultiRunBranches.Test.ps1
@@ -76,6 +76,23 @@ Describe "GetWorkflowMultiRunBranches Action" {
             $outputName | Should -Be "Result"
             $outputValue | Should -Be "{`"branches`":[`"test-branch`",`"some-other-branch`"]}"
         }
+
+        It 'Action filters out HEAD symbolic reference when using wildcard' {
+            $env:GITHUB_EVENT_NAME = "workflow_dispatch"
+            $env:Settings = ""
+            $env:GITHUB_REF_NAME = "main"
+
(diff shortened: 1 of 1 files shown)


mazhelez:
@copilot Can you add a test to make sure `HEAD` is never included?

aholstrup1 (approved):

PR #2044, opened 2025-11-27
Various Al Alert fixes

spetersenms:
### ❔What, Why & How

Bundle of improvements for AL alerts

* Fixes issue where Al alerts are not uploaded if build modes are used
* Allows error severity alerts to be displayed in PRs
* Add rule id to alert title
* Correctly set rule title

See PR alerts here: https://github.com/spetersenmsorg/AL-Go_BugBash/pull/16

### ✅ Checklist

- [x] Update RELEASENOTES.md
- [ ] Add telemetry

### DIFF
--- a/Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
+++ b/Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
@@ -33,7 +33,7 @@ function GenerateSARIFJson {
 
     foreach ($issue in $errorLogContent.issues) {
         # Skip issues without locations as GitHub expects at least one location
-        if (($issue.PSObject.Properties.Name -notcontains "locations" ) -or ($issue.locations.Count -eq 0)) {
+        if (($issue.PSObject.Properties.Name -notcontains "locations" ) -or ($issue.locations.Count -eq 0) -or $issue.PSObject.Properties.Name -notcontains "ruleId") {
             continue
         }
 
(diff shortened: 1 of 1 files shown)


Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
```diff
@@ -74,10 +74,17 @@ function GenerateSARIFJson {
             if ($issue.PSObject.Properties.Name -contains "fullMessage") {
                 $fullMessage = $issue.fullMessage
             }
+            if ($issue.PSObject.Properties.Name -contains "ruleId") {
+                $fullMessage = "$($issue.ruleId): $fullMessage"
+            }
```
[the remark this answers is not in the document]
spetersenms: While the rule does have an id field, this is not actually shown in the comment added to PRs, only in the security tab. Adding the rule to the tile makes it quick to see which rule is used in the PR.

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
```diff
@@ -74,10 +74,17 @@ function GenerateSARIFJson {
             if ($issue.PSObject.Properties.Name -contains "fullMessage") {
                 $fullMessage = $issue.fullMessage
             }
+            if ($issue.PSObject.Properties.Name -contains "ruleId") {
```
aholstrup1: I wonder if we should add a defensive check a bit higher up in this function and completely skip the issue if the ruleId is not there. Then down here you can assume it exists. 

I also think we'll end up running into other issues if the rule id is empty.

aholstrup1 (dismissed):

mazhelez (dismissed):

Actions/ProcessALCodeAnalysisLogs/ProcessALCodeAnalysisLogs.ps1
spetersenms: Good point, a missing ruleid would definitely be critical.

PR #2051, opened 2025-11-30
Rename tile from 'New tile' to 'Runner Environments'

jmanuelmartinez:
Rename the “New tile” caption in Scenarios/resources/telemetrydashboard.json to “Runner environments” so the title matches the actual content of the tile. 

Removed the md h1 to avoid duplication of title.

### ❔What, Why & How

Related to issue: #

### ✅ Checklist

- [N/A] Add tests (E2E, unit tests)
- [N/A] Update RELEASENOTES.md
- [N/A] Update documentation (e.g. for new settings or scenarios)
- [N/A] Add telemetry

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -1,3 +1,5 @@
+## v8.1
+
 ### Custom AL-Go files
 
 AL-Go for GitHub now supports updating files from your custom templates via the new `customALGoFiles` setting. Read more at [customALGoFiles](https://aka.ms/algosettings#customALGoFiles).


Scenarios/resources/telemetrydashboard.json
```diff
@@ -715,7 +715,7 @@
                 "width": 20,
                 "height": 3
             },
-            "markdownText": "## Runner Environments\r\nAL-Go supports both GitHub-hosted runners and self-hosted runners. In the tiles below you can get an overview over which environments AL-Go actions are running in. ",
+            "markdownText": "AL-Go supports both GitHub-hosted runners and self-hosted runners. In the tiles below you can get an overview over which environments AL-Go actions are running in. ",
```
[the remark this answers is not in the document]
aholstrup1: I tend to agree with Copilot. I think we should keep it

Scenarios/resources/telemetrydashboard.json
aholstrup1: There's a similar "New tile" title a little further down (as part of the Runner OS tile). We should probably update them both while we're at it :)

Scenarios/resources/telemetrydashboard.json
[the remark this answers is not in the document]
jmanuelmartinez: Yep, after reading copilot's suggestion it does make sense to leave it there in the content and then replace the "New tile" with the ## Runner OS.

Scenarios/resources/telemetrydashboard.json
jmanuelmartinez: Good catch! I didn't notice that one!

jmanuelmartinez: @aholstrup1 I’ve tested another workaround - leaving a space in the tile name field keeps the field blank and avoids repetitive titles. It’s not the most elegant solution, but it does the trick 😊

image

aholstrup1: Let's go with that then :)

PR #2075, opened 2026-01-05
Bug fixes for running a large/complicated solution on AL-Go

### LINKED ISSUE #2084: [Bug]: Multiple artifacts failure if you re-run failed jobs after flaky tests
### AL-Go version

latest

### Describe the issue

If a build is cancelled during tests or failing due to instability - and you subsequently re-run failed jobs (not all jobs) you will get multiple artifacts for succeeding projects and subsequent jobs cannot locate these artifacts (they get an array of artifacts). The "double" artifacts are not visible in the UI, but found when artifacts are queried. You could see this as a bug in GitHub, but we should fix this in AL-Go in a way that works, even if GitHub fixes their bug.

This build: https://github.com/Freddy-DK/DependenciesProblems/actions/runs/21090442931

Has double artifacts, see: https://api.github.com/repos/Freddy-DK/DependenciesProblems/actions/runs/21090442931/artifacts

image

When than downloading artifacts from prior builds for incremental builds, it fails with
```
throw "Multiple artifacts found with mask $artifactMask for project $projectName"
```
See: https://github.com/Freddy-DK/DependenciesProblems/actions/runs/21090506962/job/60661024240

### Expected behavior

No failure - latest re-run build should be used

### Steps to reproduce

Fork Freddy-DK/DependenciesProblems
Run CI/CD - cancel after artifacts of P1 has been published (during test analyzer)
Re-run failed jobs
Modify P0/app.json description
Build fails due to multiple artifacts

_No response_

### LINKED ISSUE #2085: [Bug]: Projects that doesn't contain both Apps and TestApps are wrongly seen as not built.
### AL-Go version

latest

### Describe the issue

When downloading artifacts from projects build earlier in the same workflow, it flags projects as missing if either Apps or TestApps are missing. This means that projects, which only contains TestApps or Apps will be downloaded but also looked for in the last known good build later on. It doesn't fail, because it cannot find the artifacts in that built either.

See https://github.com/Freddy-DK/DependenciesProblems/actions/runs/21079606240/job/60630194330

image

### Expected behavior

AL-Go shouldn't try to download artifacts from a prior build when the project is build in this run

### Steps to reproduce

Fork Freddy-DK/DependenciesProblems
Run CI/CD
See the output of Download Project Dependencies of P0 (thinks that P1 isn't built)

_No response_

### LINKED ISSUE #2086: [Bug]: Postpone jobs, which doesn't have any dependants to the end of the build order.
### AL-Go version

latest

### Describe the issue

I want to build all projects first and then run tests afterwards, this is not really possible today.

Today, projects are sorted in order of dependencies and as soon as projects are built, the project which tests the apps in that project will be built in the next round. This behavior causes projects which runs tests to be performed side-by-side with build projects, delaying the builds unnecessarily.

We are aiming to have build projects to complete in 1-2 minutes and test projects will often take much longer. When test projects are run side-by-side with build projects, the subsequent build projects will wait for the test projects even though it isn't necessary.

We want this:
image

to become this:
image

Which completes in 2:01, saving a full 45 minutes from the first run, just by re-ordering and allowing people to see compile errors up-front - fail early!

The key difference here is just the order in which projects are built/tested. In the first run, build3 had to wait 30 minutes until the test run was done before building the next jobs, and 1 hour and 44 until the last build jobs would run.
In the second run - all test jobs are running in parallel at the very end, getting compile errors quicly and postponing test runs until the very last.

If necessary, I can put the modified behavior behind a setting, so that you can get the old behavior if you like.

### Expected behavior

No unnecessary delays when building projects.

### Steps to reproduce

Create a multi-project repository, with 4 levels of depth like:
P0 contains app and test app
P1 contains app (with dependency on P0) and test app
P2 contains app (with dependency on P1) and test app
P0.test points to test folder in P0
P1.test points to test folder in P1
P2.test points to test folder in P2

Run CI/CD, you will see:

P0 built first
Then P1 + P0.Test
Then P2 + P1.Test
Then P2.Test

Should be

P0
Then P1
Then P2
Then P0.Test + P1.Test + P2.Test

_No response_

freddydk:
### ❔What, Why & How

This PR fixes a few minor bugs and 3 major, all changes required for moving Bunker Holding Group to AL-Go for GitHub
 
[Bug] Fix a small spelling mistake in DownloadProjectDependencies.Action.ps1‎

[Bug] Avoid failing entire workflows due to cleanup issues (f.ex. if no Container is used) in PipelineCleanup.ps1

Fixes #2086 - In AL-Go-Helper.ps1

The sort order is changed from this:
image

to this:
image

Which completes in 2:01, saving 45 minutes from the first run, just by re-ordering and allowing people to see compile errors up-front - fail early!
The key difference here is just the order in which projects are built/tested. In the first run, build3 had to wait 30 minutes until the test run was done before building the next jobs, and 1 hour and 44 until the last build jobs would run.
In the second run - all test jobs are running in parallel at the very end, getting compile errors quicly and postponing test runs until the very last.

Subsequently, when using GitHub hosted Custom runners, we can achieve this:
image

Note that the custom runners are running tests in 40 minutes, where the GitHub hosted runners takes 104 minutes. This is because the GitHub hosted custom runners are using SQL Server Developer edition and no docker.

Saving a full 2 hours, bringing a complete build from 2 hours 46 minutes down to 46 minutes.

Fixes #2085 - In GetDependencies in GitHub-Helper.ps1

When downloading artifacts from projects build earlier in the same workflow, it flags projects as missing if either Apps or TestApps are missing. This means that projects, which only contains TestApps or Apps will be downloaded but also looked for in the last known good build later on. It doesn't fail, because it cannot find the artifacts in that built either.
See https://github.com/Freddy-DK/DependenciesProblems/actions/runs/21079606240/job/60630194330

image

Fixes #2084 in GetArtifactsFromWorkflowRun in GitHub-Helper.ps1

If a build is cancelled during tests or failing due to instability - and you subsequently re-run failed jobs (not all jobs) you will get multiple artifacts for succeeding projects and subsequent jobs cannot locate these artifacts (they get an array of artifacts). The "double" artifacts are not visible in the UI, but found when artifacts are queried. You could see this as a bug in GitHub, but we should fix this in AL-Go in a way that works, even if GitHub fixes their bug.

This build: https://github.com/Freddy-DK/DependenciesProblems/actions/runs/21090442931
Has double artifacts: https://api.github.com/repos/Freddy-DK/DependenciesProblems/actions/runs/21090442931/artifacts

image

When than downloading artifacts from prior builds for incremental builds, it fails with
```
throw "Multiple artifacts found with mask $artifactMask for project $projectName"
```
See: https://github.com/Freddy-DK/DependenciesProblems/actions/runs/21090506962/job/60661024240

### ✅ Checklist

- [x] Update RELEASENOTES.md
- [ ] Add telemetry


freddydk:

freddydk:

mazhelez:
Interesting idea on postponing jobs to actually save time overall. 
However, if there is a build job with no tests, it will be artificially pushed to the back

E.g., the following build order
```mermaid
flowchart TD
    A[Build apps, Project A] --> B[Build apps, Project B]
    A[Build apps, Project A] --> C[Build apps, Project C]
    B --> D[Test apps, Project A]
    B --> E[Test apps, Project B]
```
    
  will become
```mermaid
flowchart TD
    A[Build apps, Project A] --> B[Build apps, Project B]
    B --> D[Test apps, Project A]
    B --> E[Test apps, Project B]
    B --> C[Build apps, Project C]
```

That way compilation errors might have been earlier, saving time and compute.

My suggestion is to try out adding project dependencies _explicitly_.
AL-Go calculates the projects dependencies, the build depth and the build order automagically, based on the apps in each project.
That way "Build apps, Project C" (in my example above) ends up on the second build level. But if you set it to depend on "Build apps, Project B", then it will end up on the next build level.
Not sure how this would look like in code, but seems like a neater way to control the build matrix.

Let me know what you think about this idea.

freddydk:
> Let me know what you think about this idea.

Your example above is wrong, if you had that order of dependencies, the matrix would look like this in the current version of AL-Go:

```mermaid
flowchart TD
    A[Build apps, Project A] --> B[Build apps, Project B]
    A[Build apps, Project A] --> C[Build apps, Project C]
    A[Build apps, Project A] --> D[Test apps, Project A]
    B --> E[Test apps, Project B]
```

Causing the test of B to be delayed until the test of A is done, that is exactly what I want to avoid.
But you are right that it would postpone C to later, which actually also is the case at Bunker (see images above)

I would very much warn against explicit project dependency definitions in AL-Go, it complicates things a lot - I think build order needs to be calculated based on dependencies - else adding a dependency to an app.json or a new app will cause you to need to manually re-think build order in some structure.

In Bunker we have many projects and many apps - this could become a nightmare.

What I could add to this PR instead would be to add a project setting called: **PostponeProjectInBuildOrder**, which you could set to true on projects, which would be allowed to be scheduled later (if possible, i.e. if nobody depends on it). This way, I would set that on all test projects and the build projects would not be deferred, but built at the earliest time, essentially causing the result to be:

```mermaid
flowchart TD
    A[Build apps, Project A] --> B[Build apps, Project B]
    A[Build apps, Project A] --> C[Build apps, Project C]
    B --> D[Test apps, Project A]
    B --> E[Test apps, Project B]
```

Which is exactly what we want (and better than my current suggestion)

I have added this property to the PR - can remove again if needed. @mazhelez - what do you think about that property? (default is false causing the build order to be like before)
If we keep this, then I will add a few tests for the property and the impact on the build order.

freddydk:
Adding the postponeProjectInBuildOrder property and "postponeProjectInBuildOrder": true to my test projects and causes build order to be like:
image

Which is much better than before - and exactly what we want.

freddydk:

freddydk:
There is something wrong with your @microsoft-github-policy-service agree thingy - I am constantly asked to agree with this - it doesn't make any sense

Actions/AL-Go-Helper.ps1
```diff
@@ -1851,11 +1865,49 @@ Function AnalyzeProjectDependencies {
...
+                            $hasRemainingDependents = $true
+                        }
+                    }
+                }
+            }
+            if (!$hasRemainingDependents) {
+                Write-Host "Project $_ has no remaining dependents, can be built later"
+            }
+            return -not $hasRemainingDependents
+        })
```
aholstrup1, line 1937: Could the logic inside the Where-Object be its own helper function? Something like Test-HasRemainingDependents

Would help keep the code inside AnalyzeProjectDependencies  to a minimal

freddydk: I will extract that to a function of its own.

freddydk: done

Actions/AL-Go-Helper.ps1
```diff
@@ -1723,6 +1723,38 @@ function CheckAndCreateProjectFolder {
     }
 }
 
+function TestIfProjectHasRemainingDependents {
```
mazhelez: Minor: naming


```suggestion
function TestIfProjectHasDependents {
```

Actions/AL-Go-Helper.ps1
```diff
@@ -1723,6 +1723,38 @@ function CheckAndCreateProjectFolder {
...
+        [string] $project,
+        [string[]] $projects,
+        [hashtable] $appDependencies,
+        [array] $projectsOrder
+    )
+
+    $hasRemainingDependents = $false
+    foreach($otherProject in $projects) {
+        if ($otherProject -ne $project) {
+            # Grab dependencies from other project, which haven't been built yet
```
mazhelez: Minor: comment accuracy


```suggestion
            # Grab dependencies from other project, which haven't been included in the build order yet
```

Actions/AL-Go-Helper.ps1
```diff
@@ -1723,6 +1723,38 @@ function CheckAndCreateProjectFolder {
...
+            $otherDependencies = $appDependencies."$otherProject".dependencies | Where-Object {
+                $dependency = $_
+                $alreadyBuilt = ($projectsOrder | ForEach-Object { $_.Projects | Where-Object { $appDependencies."$_".apps -contains $dependency } })
+                return -not $alreadyBuilt
+            }
+            Write-Host "Other project $otherProject has unbuilt dependencies: $($otherDependencies -join ", ")"
```
mazhelez:
```suggestion
            Write-Host "Other project $otherProject has dependencies that are not in the build order yet: $($otherDependencies -join ", ")"
```

Actions/AL-Go-Helper.ps1
```diff
@@ -1723,6 +1723,38 @@ function CheckAndCreateProjectFolder {
...
+            foreach($dependency in $otherDependencies) {
+                if ($appDependencies."$project".apps -contains $dependency) {
+                    Write-Host "Project $project is still a dependency for project $otherProject"
+                    $hasRemainingDependents = $true
+                }
+            }
+        }
+    }
+    if (!$hasRemainingDependents) {
+        Write-Host "Project $project has no remaining dependents, can be built later"
```
mazhelez:
```suggestion
        Write-Host "Project $project has no dependents, can be built later"
```

Actions/AL-Go-Helper.ps1
```diff
@@ -1851,11 +1897,28 @@ Function AnalyzeProjectDependencies {
...
+
+        # Remove projects in this job from the list of projects to be built (including the projects without dependents)
+        $projects = @($projects | Where-Object { $thisJob -notcontains $_ })
+
+        # Do not build jobs without dependents until the last job, remove them from this job
+        $thisJob = @($thisJob | Where-Object { $projectsWithoutDependents -notcontains $_ })
+
+        if ($projects.Count -eq 0) {
+            # Last job, add jobs without dependents
+            Write-Host "Adding jobs without dependents to last build job"
```
mazhelez:
```suggestion
            Write-Host "Adding projects without dependents to last build job"
```

Actions/AL-Go-Helper.ps1
```diff
@@ -1785,6 +1821,11 @@ Function AnalyzeProjectDependencies {
     # }
     $no = 1
     $projectsOrder = @()
+    # Collect projects without dependants, which can be build later
```
mazhelez:
```suggestion
    # Collect projects without dependants, which can be built later
```

Actions/Github-Helper.psm1
```diff
@@ -1015,6 +1029,11 @@ function GetArtifactsFromWorkflowRun {
     # Get sanitized project names (the way they appear in the artifact names)
     $projectArr = @(@($projects.Split(',')) | ForEach-Object { $_.Replace('\','_').Replace('/','_') })
 
+    # Get branch used in workflowRun
+    $workflowRunInfo = (InvokeWebRequest -Headers $headers -Uri "$api_url/repos/$repository/actions/runs/$workflowRun").Content | ConvertFrom-Json
+    $branch = $workflowRunInfo.head_branch.Replace('\', '_').Replace('/', '_')
```
mazhelez: Why is the branch filter needed at all?

Scenarios/settings.md
```diff
@@ -48,6 +48,7 @@ When running a workflow or a local script, the settings are applied by reading s
...
+| <a id="postponeProjectInBuildOrder"></a>postponeProjectInBuildOrder | Setting this project setting to true will cause the project to be postponed until the last build job when possible (i.e. if no other projects depends on it). If set on test projects, then all tests can be deferred until all builds have succeeded. | false |
```
mazhelez: Copilot can help you rewrite this doc to be more descriptive.
Also, `postponeProjectInBuildOrder` makes sense only when `useProjectDependencies` is set to _true_ as well, right?

Tests/DetermineProjectsToBuild.Test.ps1
```diff
@@ -747,6 +747,131 @@ Describe "Get-ProjectsToBuild" {
         { Get-ProjectsToBuild -baseFolder $baseFolder -maxBuildDepth 1 } | Should -Throw "The build depth is too deep, the maximum build depth is 1. You need to run 'Update AL-Go System Files' to update the workflows"
     }
 
+    It 'postpones projects if postponeProjectInBuildOrder is set to true' {
```
mazhelez, line 852: Can you also include tests for the following scenarios:
- Postponing a project that is already in the last build job matrix does nothing.
- If `useProjectDependencies` is _false_, then `postponeProjectInBuildOrder` does nothing as well.

mazhelez (changes requested):
PR looks good.
Left minor comments about naming, doc and test.

It isn't clear why adding the branch filter when getting workflow run artifacts is needed.

Actions/Github-Helper.psm1
freddydk: Issue was that the former artifact pattern was:
```
"$project-*-$mask-*"
```
and we had projects called:

BHG
and
BHG-Foundation

(and a similar pattern in other cases)

meaning that it would find artifacts called:

BHG-main-Apps
AND
BHG-Foundation-main-Apps

with the former mask, when searching for artifacts for BHG artifacts.

mazhelez: That makes sense. I see there might be a similar problem when dealing with releases.

freddydk: Will check that, thanks (we will run into that when releasing:-) )

Tests/DetermineProjectsToBuild.Test.ps1
freddydk: I have added some tests to two existing tests (setting the property and testing that it doesn't modify anything)

Actions/Github-Helper.psm1
freddydk: Indeed, when downloading releases, the same problem might occur - but there we have a different problem.
The problem is that we cannot see what branch was used to create the release.

so project-*-mask will match

BHG-main-Apps
and
BHG-Foundation-main-Apps
and any other BHG-project

I have modified the downloadRelease function to use regular expressions instead:

"^$escapedProject-[^-]+-$escapedMask-.+\.zip$"

Only difference here is that branch name used to create the release cannot contain a -, which I would consider a reasonable tradeoff (the branches used for creating a release is either main or a release branch (i.e. release/v1.0.0))

mazhelez (dismissed):

freddydk:

RELEASENOTES.md
```diff
@@ -1,5 +1,8 @@
 ### Issues
```
aholstrup1, line 17: A new version of AL-Go (v8.2) has been released.

Please move your release notes changes to above the `## v8.1` section in the RELEASENOTES.md file.

RELEASENOTES.md
```diff
@@ -52,6 +55,10 @@ Previously, when running the "Publish To Environment" workflow with an environme
 
 Now, the workflow will fail with a clear error message if the specified environment doesn't exist. If you intentionally want to deploy to a new environment that hasn't been configured yet, you can check the **Create environment if it does not exist** checkbox when running the workflow.
 
+### New Settings
```
mazhelez: This part needs to be moved above `## v8.3`

mazhelez (changes requested):
Looks good. Just the release notes need to be adjusted.

Actions/PipelineCleanup/PipelineCleanup.ps1
```diff
@@ -3,10 +3,15 @@
...
-if ($project -eq ".") { $project = "" }
+    if ($project -eq ".") { $project = "" }
 
-$containerName = GetContainerName($project)
-Remove-Bccontainer $containerName
+    $containerName = GetContainerName($project)
+    Remove-Bccontainer $containerName
+}
+catch {
+    Write-Host "Pipeline Cleanup failed: $($_.Exception.Message)"
```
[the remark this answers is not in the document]
freddydk: This should not be visible in the UI

Actions/Github-Helper.psm1
```diff
@@ -796,10 +810,14 @@ function DownloadRelease {
         # GitHub replaces series of special characters with a single dot when uploading release assets
         $project = [Uri]::EscapeDataString($project.Replace('\','_').Replace('/','_').Replace(' ','.')).Replace('%2A','*').Replace('%3F','?').Replace('%','')
         Write-Host "project '$project'"
-        $assetPattern1 = "$project-*-$mask-*.zip"
-        $assetPattern2 = "$project-$mask-*.zip"
+        # Pattern 1: project-branch-mask-version.zip (branch used for release creation cannot contain -)
+        # Pattern 2: project-mask-version.zip (no branch)
+        $escapedProject = [regex]::Escape($project)
+        $escapedMask = [regex]::Escape($mask)
+        $assetPattern1 = "^$escapedProject-[^-]+-$escapedMask-.+\.zip$"
```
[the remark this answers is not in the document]
freddydk: Comment says that branch cannot contain -

Actions/AL-Go-Helper.ps1
```diff
@@ -27,12 +27,13 @@ $RepoSettingsFile = Join-Path '.github' 'AL-Go-Settings.json'
 $defaultCICDPushBranches = @( 'main', 'release/*', 'feature/*' )
 [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSUseDeclaredVarsMoreThanAssignments', 'defaultCICDPullRequestBranches', Justification = 'False positive.')]
 $defaultCICDPullRequestBranches = @( 'main' )
-$defaultBcContainerHelperVersion = "preview" # Must be double quotes. Will be replaced by BcContainerHelperVersion if necessary in the deploy step - ex. "https://github.com/organization/navcontainerhelper/archive/refs/heads/branch.zip"
+$defaultBcContainerHelperVersion = "https://github.com/microsoft/navcontainerhelper/archive/refs/heads/main.zip" # Must be double quotes. Will be replaced by BcContainerHelperVersion if necessary in the deploy step - ex. "https://github.com/organization/navcontainerhelper/archive/refs/heads/branch.zip"
```
mazhelez:
```suggestion
$defaultBcContainerHelperVersion = "preview"
```

The new _BCContainerHelper_ preview was just released.

mazhelez (changes requested):

Actions/AL-Go-Helper.ps1
freddydk: I can change this back to preview, but it actually makes sense to have this be the URL for BcContainerHelper from the main branch (talked to Alex about this).

The latest AL-Go will use the latest BcContainerHelper
The preview AL-Go will use the preview BcContainerHelper
The unreleased AL-Go will use the unreleased BcContainerHelper.

Please advice whether you want preview or the microsoft / main reference as here?

Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -358,8 +335,12 @@ try {
             Write-Host "Add override for $scriptName"
             Trace-Information -Message "Using override for $scriptName"
 
+            $scriptblock = (Get-Command $scriptPath | Select-Object -ExpandProperty ScriptBlock)
+            if (-not $scriptblock) {
+                throw "Failed to get scriptblock for $scriptName.ps1, please check the override for validity."
```
aholstrup1:
```suggestion
                OutputError -message "Failed to get scriptblock for $scriptName.ps1, please check the override for validity."
```

Actions/AL-Go-Helper.ps1
aholstrup1: We discussed that this should get replaced with either "preview" or "latest" when we deploy AL-Go. While we're developing AL-Go it might be good to also test against latest BCContainerHelper. 

I'm fine with either honestly 👍

aholstrup1: **🔵 defaultBcContainerHelperVersion changed to main branch zip** (low) — Claude Sonnet

**What?** Changed from `"preview"` to a direct URL to navcontainerhelper's main branch. This means all AL-Go builds will pull from whatever is currently on `main` rather than a versioned release. If `main` is unstable, this could break unrelated builds. Also, line 381's `-notlike 'https://*'` check will prevent repo settings from overriding `BcContainerHelperVersion`.

**Recommendation:** If this is needed for development/testing, revert before merging or document the intent in RELEASENOTES.md.

Actions/Github-Helper.psm1
```diff
@@ -796,10 +810,14 @@ function DownloadRelease {
         # GitHub replaces series of special characters with a single dot when uploading release assets
         $project = [Uri]::EscapeDataString($project.Replace('\','_').Replace('/','_').Replace(' ','.')).Replace('%2A','*').Replace('%3F','?').Replace('%','')
         Write-Host "project '$project'"
-        $assetPattern1 = "$project-*-$mask-*.zip"
-        $assetPattern2 = "$project-$mask-*.zip"
+        # Pattern 1: project-branch-mask-version.zip (branch used for release creation cannot contain -)
+        # Pattern 2: project-mask-version.zip (no branch)
+        $escapedProject = [regex]::Escape($project)
```
aholstrup1: When project is `"*"` $escapedProject becomes `\*.` Is that intentional? Won't that mean when we match further down, we'll look for patterns that start with a literal asterisk instead of using `"*"` as a wildcard?

Actions/AL-Go-Helper.ps1
mazhelez: Let's revert to using _preview_ for now.

Actions/Github-Helper.psm1
FreundDK: Good catch

RELEASENOTES.md
```diff
@@ -10,6 +10,13 @@ The `DownloadProjectDependencies` action now downloads app files from URLs speci
 
 - Attempt to start docker service in case it is not running
 - NextMajor (v28) fails when downloading dependencies from NuGet-feed
+- Issue 2084 Multiple artifacts failure if you re-run failed jobs after flaky tests
+- Issue 2085 Projects that doesn't contain both Apps and TestApps are wrongly seen as not built.
+- Issue 2086 Postpone jobs, which doesn't have any dependents to the end of the build order.
```
mazhelez, line 23: Isn't this the new feature?
It deserves a section of its own, not just a line in the `Issues` section.

mazhelez (dismissed):
Looks good. Just a small comment for adjusting the release notes.

RELEASENOTES.md
FreundDK: Added a section about overall build performance

aholstrup1 (approved):

mazhelez (approved):

freddydk:

freddydk:
Do I need to do more to get this merged and in preview?

aholstrup1:
Merged - Running E2E tests on it now. Once that is done, we'll ship it to preview

waldo1001:
🤔

PR #2105, opened 2026-01-29
Improved options for dependency handling

spetersenms:
### ❔What, Why & How

Additional options for handling dependencies. This is mainly needed if you build the same app with different code but use the same id, such as when you have country specific changes.

The issue you might face is that if you have a project B that depends on project A and both A and B builds the same app, but B adds country specific changes, the dependency resolution of AL-Go will take the app from A. This can cause failures.

Additionally a setting has been added to force dependency on a specific project. Imagine you have project A, B and C. A and B builds the same app, but B has country specific changes. C depends on that app, but only expect the base version without country specific changes. In this case, AL-Go will think C depends on both A and B since the app id is the same in both projects. The new setting allows you to manually set C to only depend on A.

Note: if you don't use the same app id for apps with country specific changes, this is not relevant for you.

Related to issue: #

### ✅ Checklist

- [x] Update RELEASENOTES.md
- [ ] Add telemetry

### DIFF
--- a/Actions/DownloadProjectDependencies/DownloadProjectDependencies.Action.ps1
+++ b/Actions/DownloadProjectDependencies/DownloadProjectDependencies.Action.ps1
@@ -200,6 +200,9 @@ function FilterDependenciesByAppId {
 
 . (Join-Path -Path $PSScriptRoot -ChildPath "..\AL-Go-Helper.ps1" -Resolve)
 
+# Import BcContainerHelper module for Sort-AppFoldersByDependencies function
+DownloadAndImportBcContainerHelper
+
 Write-Host "Downloading dependencies for project '$project'. BuildMode: $buildMode, Base Folder: $baseFolder, Destination Path: $destinationPath"
 
 $downloadedDependencies = @()


freddydk:
Isn't this something we strongly advise NOT to do?
I would not recommend allowing this. AppId + version should yield a unique app - else things like NuGet will also have to be reconsidered.
We have special handling of NuGet because Microsoft are "allowed" to have the same appid for country specific apps, but this should be something we are going away from having - not something the encourage people to do.

Please reconsider this.

freddydk:
> Isn't this something we strongly advise NOT to do? I would not recommend allowing this. AppId + version should yield a unique app - else things like NuGet will also have to be reconsidered. We have special handling of NuGet because Microsoft are "allowed" to have the same appid for country specific apps, but this should be something we are going away from having - not something the encourage people to do.
> 
> Please reconsider this.

If AL-Go should support stuff like this - we should enable mechanisms to get new sticky appids for different buildmodes / countrycodes to keep the uniqueness.

Vjeko recently had a good blogpost describing some of the problems of having the same appids for different apps: https://vjeko.com/2026/01/27/when-guids-collide-the-app-id-problem-nobody-expected/ - please don't increase support for something we should not support.

spetersenms:
> > Isn't this something we strongly advise NOT to do? I would not recommend allowing this. AppId + version should yield a unique app - else things like NuGet will also have to be reconsidered. We have special handling of NuGet because Microsoft are "allowed" to have the same appid for country specific apps, but this should be something we are going away from having - not something the encourage people to do.
> > Please reconsider this.
> 
> If AL-Go should support stuff like this - we should enable mechanisms to get new sticky appids for different buildmodes / countrycodes to keep the uniqueness.
> 
> Vjeko recently had a good blogpost describing some of the problems of having the same appids for different apps: https://vjeko.com/2026/01/27/when-guids-collide-the-app-id-problem-nobody-expected/ - please don't increase support for something we should not support.

Yes, this is indeed not something meant to be used. This is purely for testing with BCApps, as we still internally have situations with identical id's for different apps. No worries, this will not be merged as is, I am just exploring different options for how to handle this situation.

PR #2154, opened 2026-03-04
Rework inputs handling of update workflow for trigger 'workflow_call'

OleWunschmann:
#### What

The `workflow_call` inputs in the update workflow are handled incorrectly.
They must be accessed as `inputs.*`, not `github.event.inputs.*`. 
Also, `github.event_name` will always reflect the trigger of the *calling* workflow, not the reusable workflow itself.

In my tests of the original contribution that added the trigger to the update workflow, my use case tests passed by coincidence because the parent workflow had an input named `directCommit` itself and all other inputs used their default values.

#### Why

When the update workflow is invoked as a reusable workflow, it should use the values defined in its own inputs.
These should be treated the same as when the workflow is triggered by `workflow_dispatch` (for example, when determining branches).

#### How

- Modified the action `GetWorkflowMultiRunBranches`:
  - Added a new optional input `workflowEventName`
    - Default: `github.event_name`
    - Purpose: override the event name used to determine which branches to include
    - Event `workflow_call` resuls in the same output as event `workflow_dispatch`
- Modified the update workflow:
  - Replaced all uses of `github.event.inputs.*` with `inputs.*`
  - Added a new required workflow_call input `caller` (string):
    - Marks that the update workflow was called from another workflow
  - Added a new environment variable `WorkflowEventName`:
    - Set to `github.ref_name` or to `workflow_call` when the `caller` input is present
    - Passed to input `workflowEventName` of action `GetWorkflowMultiRunBranches`
    - Used to determine the commit options

Related to original discussion #1855 and PR #2031

### ✅ Checklist

- [x] Update RELEASENOTES.md

### DIFF
--- a/Tests/DetermineProjectsToBuild.Test.ps1
+++ b/Tests/DetermineProjectsToBuild.Test.ps1
@@ -1044,7 +1044,7 @@ Describe "Get-ProjectsToBuild" {
     }
 
     It 'throws error when test project has buildable app folders' {
-        # TestProject has both projectsToTest setting AND an app folder — this should fail
+        # TestProject has both projectsToTest setting AND an app folder - this should fail
 
         $appFile = @{ id = '83fb8305-4079-415d-a25d-8132f0436fd1'; name = 'First App'; publisher = 'Contoso'; version = '1.0.0.0'; dependencies = @() }
         New-Item -Path "$baseFolder/Project1/.AL-Go/settings.json" -type File -Force
@@ -1052,7 +1052,7 @@ Describe "Get-ProjectsToBuild" {
 
         New-Item -Path "$baseFolder/TestProject/.AL-Go/settings.json" -type File -Force
         @{ projectsToTest = @("Project1") } | ConvertTo-Json -Depth 99 -Compress | Out-File (Join-Path $baseFolder "TestProject/.AL-Go/settings.json") -Encoding UTF8
-        # Add an app folder to the test project — this should be forbidden
+        # Add an app folder to the test project - this should be forbidden
         $testAppFile = @{ id = '83fb8305-4079-415d-a25d-8132f0436fd2'; name = 'Bad App'; publisher = 'Contoso'; version = '1.0.0.0'; dependencies = @() }
         New-Item -Path "$baseFolder/TestProject/app/app.json" -Value (ConvertTo-Json $testAppFile -Depth 10) -type File -Force
 
@@ -1066,7 +1066,7 @@ Describe "Get-ProjectsToBuild" {
     }
 
     It 'throws error when test project has buildable test folders' {
-        # TestProject has both projectsToTest setting AND a test folder — this should fail
+        # TestProject has both projectsToTest setting AND a test folder - this should fail
 
         $appFile = @{ id = '83fb8305-4079-415d-a25d-8132f0436fd1'; name = 'First App'; publisher = 'Contoso'; version = '1.0.0.0'; dependencies = @() }
         New-Item -Path "$baseFolder/Project1/.AL-Go/settings.json" -type File -Force
@@ -1089,7 +1089,7 @@ Describe "Get-ProjectsToBuild" {
 
     It 'throws error when one test project depends on another test project' {
         # Project1 is a normal project, TestProject1 and TestProject2 are both test projects
-        # TestProject2 tries to depend on TestProject1 — this should fail
+        # TestProject2 tries to depend on TestProject1 - this should fail
         Mock OutputError {} -ModuleName DetermineProjectsToBuild
 
         $appFile = @{ id = '83fb8305-4079-415d-a25d-8132f0436fd1'; name = 'First App'; publisher = 'Contoso'; version = '1.0.0.0'; dependencies = @() }


Actions/GetWorkflowMultiRunBranches/GetWorkflowMultiRunBranches.ps1
```diff
@@ -20,8 +22,8 @@ switch ($env:GITHUB_EVENT_NAME) {
         $branchPatterns = @()
       }
     }
-    'workflow_dispatch' {
-      Write-Host "Event is workflow_dispatch: getting branches from input"
+    { $_ -in 'workflow_dispatch', 'workflow_call' } {
+      Write-Host "Event is $($_): getting branches from input"
       $branchPatterns = @($includeBranches.Split(',') | ForEach-Object { $_.Trim() })
```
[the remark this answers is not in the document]
OleWunschmann: Not part of this change.

It’s been working so far because the action always invokes this script with at least an empty string. 
The check `if (-not $branchPatterns) { ... }` therefore behaves as expected and returns `$true` when the array only contains a single empty-string element.

@mazhelez @aholstrup1 Do you want me to add the suggested null-check in this PR, just to be safe?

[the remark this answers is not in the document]
mazhelez: No, leave it as is for now.

Templates/AppSource App/.github/workflows/UpdateGitHubGoSystemFiles.yaml
```diff
@@ -133,12 +139,12 @@ jobs:
...
-          directCommit: '${{ github.event.inputs.directCommit }}'
-          downloadLatest: '${{ github.event.inputs.downloadLatest }}'
+          directCommit: '${{ inputs.directCommit }}'
+          downloadLatest: '${{ inputs.downloadLatest }}'
         run: |
           $errorActionPreference = "Stop"; $ProgressPreference = "SilentlyContinue"; Set-StrictMode -Version 2.0
-          if('${{ github.event_name }}' -in 'workflow_dispatch', 'workflow_call') {
-            Write-Host "Using inputs from ${{ github.event_name }} event"
+          if('${{ env.WorkflowEventName }}' -in 'workflow_dispatch', 'workflow_call') {
+            Write-Host "Using inputs from ${{ env.WorkflowEventName }} event"
```
aholstrup1: Do you need the ${{ }} expansion in this case or could you just reference the envs directly? 

```suggestion
          if($env:WorkflowEventName -in 'workflow_dispatch', 'workflow_call') {
            Write-Host "Using inputs from $($env:WorkflowEventName) event"
```

Templates/AppSource App/.github/workflows/UpdateGitHubGoSystemFiles.yaml
```diff
@@ -21,6 +21,10 @@ on:
         default: ''
   workflow_call:
     inputs:
+      __caller:
```
aholstrup1: Could this be callingWorkflow or something like that? Or caller? Or is there a reason for the double underscore?

Tests/GetWorkflowMultiRunBranches.Test.ps1
```diff
@@ -141,4 +141,74 @@ Describe "GetWorkflowMultiRunBranches Action" {
             $outputValue | Should -Be "{`"branches`":[`"test-branch`",`"some-other-branch`"]}"
         }
     }
+
+    Context 'workflow_call event' {
+        It 'Action sets the current branch as result when no branch patterns are specified' {
+            $env:GITHUB_REF_NAME = "main"
+
```
aholstrup1, line 159: I can see we mock invoke-git in all the other tests. Is that needed here too before invoking the action?

Templates/AppSource App/.github/workflows/UpdateGitHubGoSystemFiles.yaml
OleWunschmann: I wanted to mark this input as an internal input that is only used for the call trigger.
I’m not sure why I used a double underscore instead of a single underscore.
But I can rename the input to "caller" if you prefer.

Templates/AppSource App/.github/workflows/UpdateGitHubGoSystemFiles.yaml
OleWunschmann: My idea was to highlight the difference between workflow envs and step envs. 
But I agree that it's clearer and more consistent to reference the envs directly.

I implemented your suggested change for the workflow in both templates.

Tests/GetWorkflowMultiRunBranches.Test.ps1
OleWunschmann: The tests were based on the existing tests for the workflow_dispatch event, and the first test there also lacked the mock.

I made the following changes:
- Added the mock to the first tests for workflow_dispatch and workflow_call.
- Aligned the workflow_call tests to match the workflow_dispatch tests.

Tests/GetWorkflowMultiRunBranches.Test.ps1
```diff
@@ -39,6 +39,8 @@ Describe "GetWorkflowMultiRunBranches Action" {
             $env:Settings = ""
             $env:GITHUB_REF_NAME = "main"
 
+            Mock -CommandName invoke-git -ParameterFilter { $command -eq 'for-each-ref'}  -MockWith  { return @("origin/main") }
+
```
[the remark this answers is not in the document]
OleWunschmann: Added mocks for 'fetch' to all tests in this file.

OleWunschmann:
@aholstrup1
I added a fix for the failing workflow "PSScriptAnalyzer".
The problem seems to be the usage of "—". I replaced them with "-".

PR #2162, opened 2026-03-09
Test Project feature

spetersenms:
### ❔What, Why & How

This new feature brings the concept of test projects to AL-Go for GitHub. The purpose of this is to allow the user to split building and testing into separate projects. This can be especially useful if some tests are unstable and might require a re-run to succeed. In such scenarios, it's nice not to have to wait for compilation again. 

A test project is a special type of project that depends on one or more regular projects. It will download and install artifacts from those dependent projects and run all test apps it finds. It allows any customization or settings used in regular projects, such as build modes or conditional settings. The only limitation is test projects are not allowed to build apps.

Related to issue: #

### ✅ Checklist

- [x] Update RELEASENOTES.md
- [ ] Add telemetry

### DIFF
--- a/Actions/AL-Go-Helper.ps1
+++ b/Actions/AL-Go-Helper.ps1
@@ -1959,11 +1959,14 @@ Function AnalyzeProjectDependencies {
                 }
             }
             # If this project is a test project, add its target projects as direct dependencies
+            # Only add target projects that haven't been built yet (still in $projects), matching the normal dependency resolution pattern
             if ($testProjectTargets.Keys -contains $project) {
                 foreach($targetProject in $testProjectTargets."$project") {
-                    $foundDependencies += $targetProject
-                    if ($projectDependencies.Keys -contains $targetProject) {
-                        $foundDependencies += $projectDependencies."$targetProject"
+                    if ($projects -contains $targetProject) {
+                        $foundDependencies += $targetProject
+                        if ($projectDependencies.Keys -contains $targetProject) {
+                            $foundDependencies += $projectDependencies."$targetProject"
+                        }
                     }
                 }
             }


Actions/.Modules/ReadSettings.psm1
```diff
@@ -190,6 +190,7 @@ function GetDefaultSettings
         "templateBranch"                                = ""
         "appDependencyProbingPaths"                     = @()
         "useProjectDependencies"                        = $false
+        "testProject"                                   = @()
```
aholstrup1: Since this is an array, it feels like this should be plural no? So `testProjects`

Actions/AL-Go-Helper.ps1
```diff
@@ -1852,6 +1870,55 @@ Function AnalyzeProjectDependencies {
...
+    # Handle testProject settings: inject explicit dependencies from target project names
+    # First pass: collect all test projects so we can validate that no project depends on a test project
+    $testProjectNames = @{}
+    foreach($project in $projects) {
+        $projectSettings = ReadSettings -project $project -baseFolder $baseFolder
+        if ($projectSettings.testProject -and $projectSettings.testProject.Count -gt 0) {
+            # Validate that test projects do not contain buildable code
+            $buildableFolders = @($appDependencies."$project".apps)
+            if ($buildableFolders.Count -gt 0) {
+                OutputError "Test project '$project' must not contain buildable code. Remove appFolders, testFolders, and bcptTestFolders or remove the testProject setting."
```
mazhelez: Isn't that case already handled in `AnalyzeRepo`?

Actions/AL-Go-Helper.ps1
```diff
@@ -717,12 +717,30 @@ function AnalyzeRepo {
...
+    }
     if (!$settings.doNotRunTests -and -not $settings.testFolders) {
-        if (-not ($doNotIssueWarnings -or $settings.doNotBuildTests)) { OutputWarning -message "No test apps found in testFolders in $ALGoSettingsFile" }
-        $settings.doNotRunTests = $true
+        if ($isTestProject) {
+            # Test projects run tests from upstream projects via installTestApps, not from local testFolders
+            Write-Host "Test project: tests will be run from installed test apps"
+            $settings.installTestRunner = $true
+            $settings.installTestFramework = $true
+            $settings.installTestLibraries = $true
```
aholstrup1: Can we not just take these from the repos settings file? You may not need all these 🤷‍♂️

Actions/AL-Go-Helper.ps1
```diff
@@ -1852,6 +1870,55 @@ Function AnalyzeProjectDependencies {
...
+                OutputError "Test project '$project' must not contain buildable code. Remove appFolders, testFolders, and bcptTestFolders or remove the testProject setting."
+            }
+            $testProjectNames[$project] = $projectSettings.testProject
+        }
+    }
+    foreach($project in $testProjectNames.Keys) {
+        Write-Host "Project '$project' is a test project targeting: $($testProjectNames[$project] -join ', ')"
+        $testProjectDeps = @()
+        foreach($targetProject in $testProjectNames[$project]) {
+            # Resolve target project name: support both full paths (path/to/projectName) and short names (projectName)
```
mazhelez: `projectName` here is ambiguous as I thought it the [projectName](http://aka.ms/algosettings#projectName) setting.

I suggest supporting only the full path for now.

RELEASENOTES.md
```diff
@@ -1,3 +1,32 @@
...
+{
+  "testProject": ["MyProject"]
+}
+```
+
+AL-Go will automatically:
+
+- Resolve the dependency so the test project always builds after its target project(s).
+- Install the Test Runner, Test Framework, and Test Libraries into the container.
+- Run all tests from the installed test apps (`runTestsInAllInstalledTestApps`).
```
aholstrup1:
```suggestion
- Run all tests from the installed test apps.
```

Actions/.Modules/ReadSettings.psm1
spetersenms: Maybe? The thought behind it is that the current project is a test project (singular) if this setting is set. testProjects might not be a good name since that would indicate it depends on several test projects? Maybe testProjectDependencies instead?

mazhelez: How about `projectsToTest`?

spetersenms: Went with Marias suggestion :)

Actions/AL-Go-Helper.ps1
spetersenms: The first one is mandatory for running tests, so I kept that. For the other two I now write a log mentioning they are false.

Actions/AL-Go-Helper.ps1
spetersenms: You are right, the first check is redundant.

Actions/AL-Go-Helper.ps1
spetersenms: Yeah that might be the safer choice for now. I'll change to only allow full paths.

Tests/DetermineProjectsToBuild.Test.ps1
```diff
@@ -974,6 +974,198 @@ Describe "Get-ProjectsToBuild" {
...
+
+        $allProjects, $modifiedProjects, $projectsToBuild, $projectDependencies, $buildOrder = Get-ProjectsToBuild -baseFolder $baseFolder
+
+        $allProjects | Should -BeExactly @("Project1", "Project2", "TestProject")
+
+        $projectDependencies['Project1'] | Should -BeExactly @()
+        $projectDependencies['Project2'] | Should -BeExactly @("Project1")
+        # TestProject should depend on both Project2 AND Project1 (transitive)
+        $projectDependencies['TestProject'] | Should -Contain "Project1"
+        $projectDependencies['TestProject'] | Should -Contain "Project2"
```
mazhelez, line 1037: Interesting. Doesn't that mean that `TestProject` will run tests from `Project1` as well?
I wouldn't expect that if `projectsToTest` is only `Project2`.

spetersenms: Yes. The thought process is you might have some shared functionality, in this case in `Project1`. So the tests you build in `Project2` might require the stuff from `Project1`, which would then fail to run in `TestProject` if it doesn't get a dependency to `Project1` as well.

mazhelez: Agreed to keep it as is for now and maybe potentially changing in the future.

Actions/AL-Go-Helper.ps1
```diff
@@ -1852,6 +1873,49 @@ Function AnalyzeProjectDependencies {
...
+            if (-not $resolvedTarget) {
+                OutputError "Test project '$project' references project '$targetProject' which does not exist in the repository. Use the full project path (e.g. 'projects/MyProject')."
+                throw
+            }
+            # Validate that the target is not itself a test project
+            if ($testProjectNames.Keys -contains $resolvedTarget) {
+                OutputError "Test project '$project' references '$resolvedTarget' which is also a test project. A test project cannot depend on another test project."
+                throw
+            }
+            # Insert 'fake' app to force dependency from test project to target project
```
mazhelez: Not a great fan of such shenanigans, but ok 😄

Scenarios/settings.md
```diff
@@ -77,6 +77,7 @@ The repository settings are only read from the repository settings file (.github
...
+| <a id="projectsToTest"></a>projectsToTest | An array of project paths that this project depends on for testing. Setting this marks the project as a **test project** — it will not build any apps itself but will install the apps produced by the referenced projects and run tests against them. This lets you separate test execution from compilation so you can re-run tests without a full rebuild.<br />The project must not contain any buildable code (`appFolders`, `testFolders`, or `bcptTestFolders`). A test project cannot depend on another test project. Use full project paths (e.g. `"build/projects/MyApp"`).<br />Default: `[]` |
```
mazhelez: This should actually be under "Basic Project settings"

Actions/AL-Go-Helper.ps1
spetersenms: Yeah maybe not the best approach. I did a small update that injects test dependencies into `$foundDependencies`.

Scenarios/settings.md
spetersenms: Yep you are right.

Actions/AL-Go-Helper.ps1
```diff
@@ -717,12 +717,33 @@ function AnalyzeRepo {
...
+            if (-not $settings.installTestFramework) {
+                Write-Host "Test project: installTestFramework is not set. Add it to the project settings if upstream test apps depend on the test framework."
+            }
+            if (-not $settings.installTestLibraries) {
+                Write-Host "Test project: installTestLibraries is not set. Add it to the project settings if upstream test apps depend on test libraries."
+            }
+            if (-not $settings.runTestsInAllInstalledTestApps) {
+                Write-Host "runTestsInAllInstalledTestApps is false, but will be forced to true as no tests would be run otherwise."
+            }
+            $settings.runTestsInAllInstalledTestApps = $true
```
aholstrup1, line 738: I'm considering if we should just remove this setting again? It was kinda made to solve the same problem you are solving. Your solution is just more complete. What do you think?

Actions/AL-Go-Helper.ps1
```diff
@@ -1852,6 +1879,49 @@ Function AnalyzeProjectDependencies {
...
+            # Resolve target project name: must be the full project path
+            # Normalize slashes to match how project keys are stored (OS-dependent)
+            $normalizedTarget = $targetProject.Replace('/', [System.IO.Path]::DirectorySeparatorChar).Replace('\', [System.IO.Path]::DirectorySeparatorChar)
+            $resolvedTarget = $null
+            if ($appDependencies.Keys -contains $normalizedTarget) {
+                $resolvedTarget = $normalizedTarget
+            }
+            if (-not $resolvedTarget) {
+                OutputError "Test project '$project' references project '$targetProject' which does not exist in the repository. Use the full project path (e.g. 'projects/MyProject')."
+                throw
```
aholstrup1, line 1913: Is throw needed when you already do OutputError? 🤔

Actions/AL-Go-Helper.ps1
spetersenms: It can still cover a different scenario, right? You might have `ProjectB` that depends on `ProjectA`, both with tests. If you want to install the dependencies from `ProjectA` in `ProjectB`, but not run the tests, you would need that setting.

Actions/AL-Go-Helper.ps1
spetersenms: We need both. `OutputError` only formats the error with `::error::` for GitHub, but does not itself terminate the script right away. It should still fail the actual step in the pipeline, but the throw means we can fail a bit faster.

Actions/AL-Go-Helper.ps1
```diff
@@ -1852,6 +1879,49 @@ Function AnalyzeProjectDependencies {
...
+    # Second pass: resolve and validate target projects, build the testProjectTargets mapping
+    # testProjectTargets maps each test project to an array of resolved target project names
+    $testProjectTargets = @{}
+    foreach($project in $testProjectNames.Keys) {
+        Write-Host "Project '$project' is a test project targeting: $($testProjectNames[$project] -join ', ')"
+        $resolvedTargets = @()
+        foreach($targetProject in $testProjectNames[$project]) {
+            # Resolve target project name: must be the full project path
+            # Normalize slashes to match how project keys are stored (OS-dependent)
+            $normalizedTarget = $targetProject.Replace('/', [System.IO.Path]::DirectorySeparatorChar).Replace('\', [System.IO.Path]::DirectorySeparatorChar)
```
mazhelez, line 1906: I don't see this "normalization" above when `$appDependencies` is filled with the apps/dependencies per project.

spetersenms: It would not need it there? the project names are just the names detected by AL-Go, so that works as normal. The paths are the content of the new setting which needs to be normalized to work with both Windows/Linux.

Actions/AL-Go-Helper.ps1
aholstrup1: We'll leave it for now and remove it later

Actions/AL-Go-Helper.ps1
mazhelez: Ahh, I think I got it.

Actions/AL-Go-Helper.ps1
aholstrup1: Mmh odd.. Maybe that's something we should change. Once we know there's an error, we should probably just terminate 🤷‍♂️ 

Anyway.. Doesn't have to be in this PR

PR #2216, opened 2026-04-19
Add Test Isolation support

Drakonian:
### Why

BC runtime 16 (2025 W2) added the `RequiredTestIsolation` property on test codeunits. The standard BC test runner has a single, fixed `TestIsolation` value, so any project that mixes codeunits with different isolation requirements cannot run them all under one runner, those tests get excluded from CI.

AL-Go today makes one blanket `Run-TestsInBcContainer` call per test app under the default runner. Test codeunits that need a non-default runner (e.g. `TestIsolation = Disabled` for long-running integration tests that must not roll back) have to be placed in `disabledtests.json` and run manually, losing CI coverage.

This contribution closes that gap.

### Discussion reference

https://github.com/microsoft/AL-Go/discussions/1565

### How

A new opt-in `testIsolation` setting lets projects partition the test stage:

```json
"testIsolation": {
  "enabled": true,
  "defaultRunnerCodeunitId": 0,
  "partitions": [
    { "runnerCodeunitId": 130451, "codeunits": "99209" }
  ]
}
```

When `enabled`, `RunPipeline` installs a `-RunTestsInBcContainer` scriptblock into `Run-AlPipeline` that, per test app:

1. Invokes `Run-TestsInBcContainer` once per entry in `partitions`, with `-testRunnerCodeunitId` and `-testCodeunitRange` set from the entry.
2. Issues one trailing call under `defaultRunnerCodeunitId` whose `-testCodeunitRange` is the negation of every explicit partition's filter, so codeunits not in any partition run exactly once under the default runner.

The scriptblock forwards all other Run-AlPipeline parameters verbatim, so `disabledTests.json` handling, JUnit result appending, container lifecycle, and `useCompilerFolder` support continue to work unchanged. Feature is strictly additive and off by default.

### Validation

- Pester suite passes locally and in CI.
- End-to-end on a real BC 27 project ([parallel-worker-bc](https://github.com/Drakonian/parallel-worker-bc/actions/runs/24638819749), green): 33 integration tests that had been permanently disabled via `disabledtests.json` for lack of a `TestIsolation = Disabled` runner now execute successfully under standard BCApps runner `130451`, while the remaining test codeunits (99201–99203) run under the default runner in the same build — proving both the partition call and the negated-filter fallback work against real BcContainerHelper.


Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -487,6 +487,18 @@ try {
...
+    if ($settings.testIsolation.enabled -and -not $settings.doNotRunTests) {
+        Import-Module (Join-Path $PSScriptRoot '../.Modules/TestIsolation.psm1' -Resolve)
+        Write-Host "Test isolation enabled - $($settings.testIsolation.partitions.Count) explicit partition(s) + default runner ($($settings.testIsolation.defaultRunnerCodeunitId))"
+
+        $testIsolationTelemetry = [System.Collections.Generic.Dictionary[[System.String], [System.String]]]::new()
+        Add-TelemetryProperty -Hashtable $testIsolationTelemetry -Key 'PartitionCount' -Value "$($settings.testIsolation.partitions.Count)"
+        Add-TelemetryProperty -Hashtable $testIsolationTelemetry -Key 'DefaultRunnerCodeunitId' -Value "$($settings.testIsolation.defaultRunnerCodeunitId)"
+        Trace-Information -Message "Test Isolation enabled" -AdditionalData $testIsolationTelemetry
+
+        $runAlPipelineParams["RunTestsInBcContainer"] = New-PartitionedTestRunnerScriptBlock -Settings $settings.testIsolation
```
mazhelez: What if there already is `RunTestsInBcContainer` override for other purposes?
Is there a way to add test isolation w/o using `RunTestsInBcContainer`?

Drakonian: @mazhelez

Great questions!

1) I can combine the user scriptblock with the new hashtable `testCodeunitRange` + `testRunnerCodeunitId`. In that case, nothing will be overwritten.

2) In the current design of `bccontainerhelper`, I see an alternative, in recent versions, `RequiredTestIsolation` was added, and it looks like this can be used as an alternative.

https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/properties/devenv-requiredtestisolation-property

https://github.com/microsoft/navcontainerhelper/blob/c5fd2aa4e7a36405e50d552a3947102a05931003/AppHandling/Run-TestsInNavContainer.ps1#L532

I think I should look into point 2, I will try testing it and let you know if it works or not.

Drakonian: @mazhelez the main problem with `RequiredTestIsolation` that it will work only for Runtime 16.0 +

Drakonian:
@mazhelez Thanks for the review, both points are addressed in the latest commit, plus one issue I found while re-checking the filter logic:

**1. Existing `RunTestsInBcContainer` override is now wrapped, not replaced.** `New-PartitionedTestRunnerScriptBlock` accepts an `InnerScriptBlock`; when a project ships `.AL-Go/RunTestsInBcContainer.ps1`, each partitioned call is routed through that override with `testCodeunitRange`/`testRunnerCodeunitId` added to the parameter hashtable. Documented in the scenario doc (the override must splat `$parameters` through for the filter to take effect).

**2. Regarding an alternative without the override hook:** `Run-TestsInBcContainer` has `-requiredTestIsolation`, but `Run-AlPipeline` doesn't expose it (and it requires runtime 16+ plus tests that actually declare `RequiredTestIsolation`). Manual partitions work on any runtime and runner. A future `partitionBy: "requiredTestIsolation"` mode could be layered on top of this setting without breaking its shape.

**3. Fixed the default-runner exclusion filter for range partitions.** The BC filter grammar only supports `<>` against single values, so `<>60200..60299` would have been an invalid filter at runtime. AL-Go now computes the complement of all partitions instead (e.g. `..60199|60302..`), which is plain documented filter syntax. Along with this:

- `codeunits` is schema-restricted to single IDs and closed ranges joined by `|` (the subset a complement can be computed for)
- overlapping partitions emit a pipeline warning
- the trailing default-runner call is skipped when partitions cover the entire ID space

RELEASENOTES.md
```diff
@@ -1,3 +1,25 @@
+### Test Isolation - run selected test codeunits under a custom test runner
```

Drakonian:
@spetersenms should be good now

mazhelez:

<details>
<summary>Example</summary>

```markdown

- Your change here ✅

## v9.2

- Already released changes
```

</details>

Drakonian:
@mazhelez done

Drakonian:
@mazhelez 

### Fix failing `TestIsolation` test (PS5 and PS7)

`returns $false when the InnerScriptBlock reports a failed partition` failed on both CI legs with:

> PropertyNotFoundException: The property 'testRunnerCodeunitId' cannot be found on this object.

**Cause:** the test's inner script block used `$parameters.testRunnerCodeunitId`, but the default-runner call intentionally omits that key when `defaultRunnerCodeunitId = 0` (BC default runner). `runtests.ps1` runs under `Set-StrictMode -Version 2.0`, where dot-access of a missing hashtable key throws — on Windows PowerShell 5.1 and PowerShell 7 alike.

**Fix (test only, no product code change):**
- Use index access `$parameters['testRunnerCodeunitId']`, which returns `$null` for a missing key under strict mode on both engines.
- Add the standard `Set-StrictMode -Version 2.0` header to `Tests/TestIsolation.Test.ps1`, matching every other `*.Test.ps1`, so the file is strict regardless of how it is invoked.

Verified locally: PS 5.1 → 31 passed / 0 failed, pwsh 7 → 43 passed / 0 failed (`TestIsolation.Test.ps1` + `ReadSettings.Test.ps1`).

PR #2221, opened 2026-04-22
Download only relevant project dependencies

spetersenms:
Currently, when an AL-Go project depends on another AL-Go project, all artifacts from an upstream matrix is downloaded, even if a given project only depends on a subset of the projects in the upstream matrix. For smaller repos with few projects this is not much of an issue, but on larger repos with many projects, this can result in a lot of waster resources spent on downloading things that are not needed.

This features updates the download logic to use glob patterns in the download artifact action, to only download relevant artifacts. This is based on the dependency json we calculate earlier.

### ✅ Checklist

- [x] Update RELEASENOTES.md
- [ ] Add telemetry

### DIFF
--- a/Actions/DownloadProjectDependencies/DownloadProjectDependencies.psm1
+++ b/Actions/DownloadProjectDependencies/DownloadProjectDependencies.psm1
@@ -362,12 +362,7 @@ function Get-DependencyArtifactPattern {
         return $null
     }
 
-    # Determine the branch name using the same logic as CalculateArtifactNames.ps1
-    $branchName = $ENV:GITHUB_HEAD_REF
-    if (-not $branchName) {
-        $branchName = $ENV:GITHUB_REF_NAME
-    }
-    $branchName = $branchName.Replace('\', '_').Replace('/', '_')
+    $branchName = Get-CurrentBranchName
 
     # Build brace-expansion entries: 2 per dependency project (*Apps covers Apps+TestApps+buildMode variants)
     $entries = @()
--- a/Actions/DownloadProjectDependencies/DownloadProjectDependencies.psm1
+++ b/Actions/DownloadProjectDependencies/DownloadProjectDependencies.psm1
@@ -328,4 +328,56 @@ function Get-DependenciesFromInstallApps {
     return $install
 }
 
-Export-ModuleMember -Function Get-AppFilesFromUrl, Get-AppFilesFromLocalPath, Get-DependenciesFromInstallApps
+<#
+    .SYNOPSIS
+    Computes a minimatch-compatible glob pattern for downloading only dependency-project artifacts.
+    .DESCRIPTION
+    Given a project and its dependency map (from projectDependenciesJson), constructs a brace-expansion
+    pattern that matches only the Apps, TestApps, and Dependencies artifacts for the dependency projects.
+    This pattern is designed for use with the actions/download-artifact 'pattern' input (which uses minimatch).
+
+    The pattern uses '*Apps' to match both 'Apps' and 'TestApps' as well as build-mode-prefixed variants
+    (e.g. 'CleanApps', 'CleanTestApps'). Similarly '*Dependencies' matches 'Dependencies' and variants.
+    .PARAMETER Project
+    The name of the current AL-Go project.
+    .PARAMETER ProjectDependencies
+    A hashtable mapping project names to arrays of their dependency project names (direct + transitive).
+    .OUTPUTS
+    A minimatch glob pattern string, or $null if the project has no dependencies.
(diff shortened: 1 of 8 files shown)


Actions/DownloadProjectDependencies/action.yaml
```diff
@@ -29,9 +29,32 @@ outputs:
...
   steps:
-    - name: Download artifacts from current build
+    - name: Compute dependency artifact pattern
+      id: computePattern
+      shell: ${{ inputs.shell }}
+      env:
+        _project: ${{ inputs.project }}
+        _projectDependenciesJson: ${{ inputs.projectDependenciesJson }}
+      run: |
+        . (Join-Path ${{ github.action_path }} "../AL-Go-Helper.ps1" -Resolve)
```
mazhelez: Suggestion: place this script in a file of its own.

mazhelez (approved):

Actions/DownloadProjectDependencies/DownloadProjectDependencies.psm1
```diff
@@ -328,6 +328,58 @@ function Get-DependenciesFromInstallApps {
...
+
+    # Build brace-expansion entries: 2 per dependency project (*Apps covers Apps+TestApps+buildMode variants)
+    $entries = @()
+    foreach ($dep in $dependencyProjects) {
+        $sanitizedDep = $dep.Replace('\', '_').Replace('/', '_')
+        $entries += "$sanitizedDep-$branchName-*Apps-*"
+        $entries += "$sanitizedDep-$branchName-*Dependencies-*"
+    }
+
+    return "{$($entries -join ',')}"
```
mazhelez, line 375: Could this become too big?

Actions/DownloadProjectDependencies/DownloadProjectDependencies.psm1
```diff
@@ -328,6 +328,58 @@ function Get-DependenciesFromInstallApps {
...
+    $dependencyProjects = @()
+    if ($ProjectDependencies.Keys -contains $Project) {
+        $dependencyProjects = @($ProjectDependencies."$Project")
+    }
+
+    if ($dependencyProjects.Count -eq 0) {
+        return $null
+    }
+
+    # Determine the branch name using the same logic as CalculateArtifactNames.ps1
```
mazhelez: Can the logic be reused?

Actions/DownloadProjectDependencies/action.yaml
aholstrup1: +1

aholstrup1 (approved):

Actions/DownloadProjectDependencies/DownloadProjectDependencies.psm1
spetersenms: There is no documented max size.  We do return it as an environment variable, which GitHub limits to 48KB max per variable, which would give us just under 50000 characters. So that is probably plenty large.

Actions/DownloadProjectDependencies/DownloadProjectDependencies.psm1
spetersenms: Moved the shared logic into AL-Go helper.

Actions/DownloadProjectDependencies/action.yaml
spetersenms: Script is in it's own file now, though still in the same action. We might want to consider if it should be moved to a separate action though to stay consistent with other AL-Go functionaltiy.

Actions/DownloadProjectDependencies/action.yaml
```diff
@@ -36,19 +36,7 @@ runs:
...
-        if ($pattern) {
-          Write-Host "Dependency artifact pattern: $pattern"
-          Add-Content -Encoding UTF8 -Path $env:GITHUB_OUTPUT -Value "hasPattern=true"
-          Add-Content -Encoding UTF8 -Path $env:GITHUB_OUTPUT -Value "pattern=$pattern"
-        }
-        else {
-          Write-Host "No dependency projects found, skipping artifact download"
-          Add-Content -Encoding UTF8 -Path $env:GITHUB_OUTPUT -Value "hasPattern=false"
-        }
+        ${{ github.action_path }}/ComputeDependencyArtifactPattern.ps1
```
aholstrup1: Consider wrapping in Invoke-AlGoAction

aholstrup1 (dismissed):

mazhelez (dismissed):

aholstrup1 (approved):

mazhelez (approved):

PR #2233, opened 2026-05-01
Preserve overwriteSettings when template project settings are merged after user repo settings

### LINKED ISSUE #2228: [Bug]: Orverwritesettings fail with custom Al-Go Template
### AL-Go version

9.0

### Describe the issue

the "overwriteSettings" setting fails to overwrite if the setting is coming from a Custom Al-Go template. 

### Expected behavior

"overwriteSettings" will overwrite a setting instead of it being merged.

### Steps to reproduce

1. Create a new Custom AL-Go-template from https://github.com/microsoft/AL-Go-AppSource. 
3. Create a new repository from the template Repo-A  
4. in the new template - add appDependencyProbingPaths for Repo-A intp .AL-Go/settings.json
5. update both repository's in order template --> Repo-A 
Repo-A should now have a probing path in .github/AL-Go-TemplateProjectSettings.doNotEdit-json

.6 in Repo-A add into .github/AL-Go-Settings.json
  "appDependencyProbingPaths"
  ],  "appDependencyProbingPaths": [ 
  ]  

when CI/CD is running for Repo-A appDependencyProbingPaths is merged instead of being overwritten

The reason for this weird setup is that our custom template contains functionality that is common for all of ours app-source repository's including Repo-A.
Else we will need to have 2 Custom AL-Go template to maintain with the same code/ functionality

jeffreybulanadi:
## Problem

When a repository uses `overwriteSettings` in `.github/AL-Go-Settings.json` to clear or replace an array property (such as `appDependencyProbingPaths`), a subsequent template-managed settings source (`AL-Go-TemplateProjectSettings.doNotEdit.json`) re-merges its own values for that property. This causes the user intent to be silently ignored.

The root cause is in `MergeCustomObjectIntoOrderedDictionary`: the function is called independently for each source. The `overwriteSettings` mechanism removes a property from the destination only at the moment of the user source merge. Any template source that appears later in the ordered list has no knowledge that the property was already overwritten, so it re-applies its values unconditionally.

Closes: microsoft/AL-Go#2228

## Solution

Track which settings names have been explicitly overwritten by any user-controlled source as `ReadSettings` iterates the ordered sources list. Pass that accumulated list into `MergeCustomObjectIntoOrderedDictionary` alongside a flag that identifies whether the current source is a template-managed file. When both conditions are true, skip the merge for that property.

### Changes to `Actions/.Modules/ReadSettings.psm1`

`MergeCustomObjectIntoOrderedDictionary` receives two new optional parameters:

- `[string[]] overwrittenSettings`: names of settings already declared as overwritten by a prior user-controlled source.
- `[bool] isTemplateSource`: true when the source file is a template-managed `*.doNotEdit.json` file.

When `isTemplateSource` is true and the property name appears in `overwrittenSettings`, the merge for that property is skipped in both the "add new properties" loop and the "update existing properties" loop.

The `ReadSettings` foreach loop now:

1. Determines `isTemplateSource` from the source file name.
2. Passes `-overwrittenSettings` and `-isTemplateSource` to both `MergeCustomObjectIntoOrderedDictionary` calls (main merge and conditional settings merge).
3. After each non-template source is merged, accumulates property names declared in its `overwriteSettings` list so subsequent template sources cannot re-introduce those values.

### Changes to `Tests/ReadSettings.Test.ps1`

Two integration tests added:

1. Verifies that `appDependencyProbingPaths` remains empty after a user overwrite even when the template project settings file contains values for that property.
2. Verifies that a user project settings file can still add values to a property after it was overwritten at the repo level, and that template values remain excluded.

## Testing

1. Create a repository that uses an AL-Go template.
2. In `.github/AL-Go-TemplateProjectSettings.doNotEdit.json`, add an entry to `appDependencyProbingPaths`.
3. In `.github/AL-Go-Settings.json`, set `overwriteSettings: ["appDependencyProbingPaths"]` and `appDependencyProbingPaths: []`.
4. Run any AL-Go workflow that reads settings.
5. Before this change, `appDependencyProbingPaths` in the resolved settings contains the template entry. After this change, it is empty as the user specified.

Unit tests can be run locally with Pester v5:

```powershell
Invoke-Pester -Path Tests/ReadSettings.Test.ps1
```


jeffreybulanadi:
The review caught a real gap. The accumulation was adding entries to the overwritten tracking list based on what was declared in overwriteSettings alone, without checking whether the source object actually provided the property value. MergeCustomObjectIntoOrderedDictionary has that check already (line 30: it only clears a destination property when the source contains both the declaration and the value), so the tracking should follow the same rule.

The fix adds the property presence check to the accumulation step: a setting is only recorded as overwritten when the source object contains the property in addition to declaring it in overwriteSettings. A source that lists a setting but provides no value no longer silently blocks template sources from contributing that setting.

A test was added for this case: user source declares overwriteSettings for appDependencyProbingPaths but provides no value, and the template source value should still reach the final settings.

PR #2245, opened 2026-05-11
Add GH_TOKEN env variable to Build step in _BuildALGoProject workflow

mazhelez:
Pass `github.token` as `GH_TOKEN` environment variable to the RunPipeline action in the `_BuildALGoProject` workflow template.

This allows script overrides (e.g. `RunTestsInBcContainer`) to use the GitHub CLI (`gh`) to access repository artifacts during the build step — for example, to download unstable test lists for test tolerance checks.

Changes made to both templates:
- `Templates/Per Tenant Extension/.github/workflows/_BuildALGoProject.yaml`
- `Templates/AppSource App/.github/workflows/_BuildALGoProject.yaml`

### DIFF
--- a/Templates/Per Tenant Extension/.github/workflows/_BuildALGoProject.yaml
+++ b/Templates/Per Tenant Extension/.github/workflows/_BuildALGoProject.yaml
@@ -200,6 +200,7 @@ jobs:
           Secrets: '${{ steps.ReadSecrets.outputs.Secrets }}'
           BuildMode: ${{ inputs.buildMode }}
           NeedsContext: ${{ env.NeedsContext }}
+          GH_TOKEN: ${{ github.token }}
         with:
           shell: ${{ inputs.shell }}
           artifact: ${{ env.artifact }}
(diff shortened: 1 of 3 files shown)


spetersenms (approved):

Templates/Per Tenant Extension/.github/workflows/_BuildALGoProject.yaml
```diff
@@ -200,6 +200,7 @@ jobs:
           Secrets: '${{ steps.ReadSecrets.outputs.Secrets }}'
           BuildMode: ${{ inputs.buildMode }}
           NeedsContext: ${{ env.NeedsContext }}
+          GH_TOKEN: ${{ github.token }}
```
aholstrup1, line 203: We do already pass github.token as a parameter to RunPipeline
https://github.com/microsoft/AL-Go/blob/2c60bf2b8a164b0c0a956088ea850b7d1d0e27cf/Actions/RunPipeline/action.yaml#L8-L11

We don't login though but presumably you could look for $ENV:_token in your override script to do that? 🤷‍♂️ Or we could rename $ENV:_token to $ENV:GH_Token in the action? That way we both pass the token as a parameter and authenticate by setting the right environment variable

mazhelez: Good point! Let me see how it would work.

mazhelez:
Closing: the _token env variable already set by RunPipeline action is sufficient — using it directly in the consuming script instead.

PR #2249, opened 2026-05-12
AL-Go native overrides - BuildInitialize

spetersenms:
### ❔What, Why & How

Adding first iteration of the new experimental feature "AL-Go hooks" which will replace the BcContainerHelper overrides used in RunPipeline. The intention is to eventually add multiple places in AL-Go for custom scrpts to hook in. This first iteration introduces a single hook called BuildInitialize which runs after the ReadSettings step in build workflows.

Related to issue: #

### ✅ Checklist

- [x] Update RELEASENOTES.md
- [ ] Add telemetry


Actions/RunOverride/RunOverride.ps1
```diff
@@ -0,0 +1,28 @@
...
+        if ($null -ne $parsed) {
+            $parameters = ConvertTo-HashTable $parsed -recurse
+        }
+    }
+    catch {
+        throw "Failed to parse parametersJson as JSON: $($_.Exception.Message)"
+    }
+    if ($parameters -isnot [hashtable]) {
+        throw "parametersJson must deserialize to a JSON object (hashtable), not $($parameters.GetType().Name)."
+    }
```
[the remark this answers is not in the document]
spetersenms: Fixed

Actions/AL-Go-Helper.ps1
```diff
@@ -89,6 +98,87 @@ function Get-ScriptOverrides() {
...
+        [Parameter(Mandatory = $true)]
+        [string] $ALGoFolderName,
+        [Parameter(Mandatory = $true)]
+        [string] $OverrideName,
+        [Parameter(Mandatory = $false)]
+        [hashtable] $Parameters = @{}
+    )
+    $overrides = Get-ScriptOverrides -ALGoFolderName $ALGoFolderName -OverrideScriptNames @($OverrideName)
+    if (-not $overrides.ContainsKey($OverrideName)) {
+        Write-Host "No override script '$OverrideName.ps1' found in '$ALGoFolderName' - skipping."
```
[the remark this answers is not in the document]
spetersenms: Changed to debug output

Actions/AL-Go-Helper.ps1
```diff
@@ -89,6 +98,87 @@ function Get-ScriptOverrides() {
...
+        return
+    }
+    Trace-Information -Message "Using override for $OverrideName"
```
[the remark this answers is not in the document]
spetersenms: Telemetry helper is always imported by invoke-algoaction.

Actions/RunHook/README.md
```diff
@@ -0,0 +1,58 @@
...
+argument (the same calling convention used by BCH script overrides).
+
+Example `.AL-Go/BuildInitialize.ps1`:
+
+```powershell
+Param(
+    [Hashtable] $parameters
+)
+Write-Host "BuildInitialize running for project '$($parameters.project)'"
+```
```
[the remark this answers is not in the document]
spetersenms: Defaults added

Templates/AppSource App/.github/workflows/_BuildALGoProject.yaml
```diff
@@ -103,6 +103,13 @@ jobs:
           ref: ${{ inputs.checkoutRef }}
           lfs: true
 
+      - name: BuildInitialize
+        uses: microsoft/AL-Go-Actions/RunOverride@main
```
aholstrup1: Is there any way to check here if the override exists so we don't run the action if it doesn't?

Templates/AppSource App/.github/workflows/_BuildALGoProject.yaml
```diff
@@ -103,6 +103,13 @@ jobs:
           ref: ${{ inputs.checkoutRef }}
           lfs: true
 
+      - name: BuildInitialize
```
aholstrup1: Could we make it a bit more user-friendly? "Run Build Initialize" or "Run Build Initialize override" or something like that?

Actions/AL-Go-Helper.ps1
```diff
@@ -89,6 +98,98 @@ function Get-ScriptOverrides() {
...
+        The folder where AL-Go override scripts are located (typically the
+        project's .AL-Go folder).
+    .PARAMETER OverrideName
+        The name of the override script to invoke (without the .ps1 extension).
+    .PARAMETER Parameters
+        Optional hashtable of parameters to pass to the override script.
+    .EXAMPLE
+        Invoke-ScriptOverride -ALGoFolderName '.AL-Go' -OverrideName 'BuildInitialize' -Parameters @{ project = '.' }
+#>
+function Invoke-ScriptOverride() {
```
aholstrup1: Does this need to live in AL-Go-Helper? Why not in a module in the RunOverride folder?

Actions/AL-Go-Helper.ps1
```diff
@@ -89,6 +98,98 @@ function Get-ScriptOverrides() {
...
+        $baseFolder = (Get-Location).Path
+    }
+    $projectPath = Join-Path $baseFolder $Project
+    $alGoFolder = Join-Path $projectPath $ALGoFolderName
+
+    # Populate default context keys so override authors can rely on them being
+    # present. Caller-supplied values in $Parameters take precedence.
+    $effectiveParameters = @{
+        project      = $Project
+        overrideName = $OverrideName
```
aholstrup1: The caller should know the override name already considering they need to name the script file after it. So I'm not sure that parameter is needed.

Actions/AL-Go-Helper.ps1
```diff
@@ -89,6 +98,98 @@ function Get-ScriptOverrides() {
...
+    }
+    foreach ($key in $Parameters.Keys) {
+        $effectiveParameters[$key] = $Parameters[$key]
+    }
+
+    Invoke-ScriptOverride -ALGoFolderName $alGoFolder -OverrideName $OverrideName -Parameters $effectiveParameters
```
aholstrup1: Which path are we on when we run the script? 

Would it make sense to do a 
Push-Location $ENV:GITHUB_WORKSPACE
...
Pop-Location

To make sure we're always in the root path when calling the overrides?

Actions/AL-Go-Helper.ps1
spetersenms: Not for this specific override, but if we expand this with a similar selection to what is currently offered in BcContainerHelper, we might need to support overrides from other actions. Eg a PreCompile override that is called from the workspace compilation action.

Actions/AL-Go-Helper.ps1
spetersenms: Yeah pinning to a specific location might be a good idea. Would it make more sense to Push-Location to the project folder instead of repo root though? That is where the override itself lives, so might be more intuitive.

Actions/AL-Go-Helper.ps1
spetersenms: True, you always know which override you are in.

Actions/AL-Go-Helper.ps1
aholstrup1: I was thinking the same actually. Wasn't sure which one made most sense. 

Maybe the project folder makes more sense? That's where the existing overrides are called from today as well I believe

Actions/AL-Go-Helper.ps1
```diff
@@ -89,6 +98,103 @@ function Get-ScriptOverrides() {
...
+    if (-not $baseFolder) {
+        $baseFolder = (Get-Location).Path
+    }
+    $projectPath = Join-Path $baseFolder $Project
+    $alGoFolder = Join-Path $projectPath $ALGoFolderName
+
+    # Populate default context keys so override authors can rely on them being
+    # present. Caller-supplied values in $Parameters take precedence.
+    $effectiveParameters = @{
+        project = $Project
```
aholstrup1, line 179: Nitpick: Now that we push location to the location of the project, I wonder if we need to pass the project as a parameter too 🤷‍♂️

aholstrup1 (dismissed):

Actions/AL-Go-Helper.ps1
spetersenms: I guess you could check the path you are in to detect the project, but this just provides a quick way to get it. It might not be needed, but also does not really hurt to have?

Templates/Per Tenant Extension/.github/workflows/_BuildALGoProject.yaml
```diff
@@ -103,6 +103,14 @@ jobs:
           ref: ${{ inputs.checkoutRef }}
           lfs: true
 
+      - name: Run Build Initialize override
```
mazhelez: Doesn't it make more sense to call this after `ReadSettings`?

Actions/RunOverride/action.yaml
```diff
@@ -0,0 +1,34 @@
+name: Run AL-Go Override
```
mazhelez: Minor: naming

Technically speaking it's not an override as nothing is being overridden.
How about something like "Run AL-Go extension script". Maybe Copilot will come up with something even better.

Actions/AL-Go-Helper.ps1
```diff
@@ -89,6 +98,103 @@ function Get-ScriptOverrides() {
...
+        [string] $Project = ".",
+        [Parameter(Mandatory = $true)]
+        [string] $OverrideName,
+        [Parameter(Mandatory = $false)]
+        [hashtable] $Parameters = @{}
+    )
+    if ($alGoOverrides -notcontains $OverrideName) {
+        throw "Override name '$OverrideName' is not a recognized AL-Go override. Allowed values: $($alGoOverrides -join ', ')."
+    }
+    $baseFolder = $ENV:GITHUB_WORKSPACE
```
mazhelez: `Get-BasePath` here?

mazhelez (dismissed):

Templates/Per Tenant Extension/.github/workflows/_BuildALGoProject.yaml
spetersenms: Yep, it does.

Actions/RunOverride/action.yaml
spetersenms: I went with 'AL-Go Hook' for now.

Actions/RunHook/action.yaml
```diff
@@ -0,0 +1,34 @@
...
+    description: Compressed JSON string with parameters to pass to the hook script as a hashtable
+    required: false
+    default: '{}'
+runs:
+  using: composite
+  steps:
+    - name: run
+      shell: ${{ inputs.shell }}
+      env:
+        _project: ${{ inputs.project }}
```
mazhelez, line 25: Minor: project might not be required (if the hook is in the initialization job, for example)

mazhelez: Oh, but not with the current design, I guess.

mazhelez (approved):

aholstrup1 (approved):

Actions/RunHook/action.yaml
spetersenms: True in the init job there would be no project, but in that case we can pass $null as project.

PR #2265, opened 2026-05-27
move the BuildInitialize hook to after determine build project

freddydk:
### ❔What, Why & How

I would very much like to use the new BuildInitialize hook with my Fkh integration to AL-Go, but I need to know whether the project is going to be build or not:-)


freddydk:

Templates/AppSource App/.github/workflows/_BuildALGoProject.yaml
```diff
@@ -128,6 +120,19 @@ jobs:
           project: ${{ inputs.project }}
           baselineWorkflowRunId: ${{ inputs.baselineWorkflowRunId }}
 
+      - name: Run Build Initialize hook
+        if: hashFiles(format('{0}/.AL-Go/BuildInitialize.ps1', inputs.project)) != ''
```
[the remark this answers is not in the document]
FreundDK: Placing it before DetermineBuildProject always run without the option of skipping or doing things different due to the value of buildIt

Templates/Per Tenant Extension/.github/workflows/_BuildALGoProject.yaml
```diff
@@ -128,6 +120,19 @@ jobs:
...
+        if: hashFiles(format('{0}/.AL-Go/BuildInitialize.ps1', inputs.project)) != ''
+        uses: microsoft/AL-Go-Actions/RunHook@main
+        with:
+          shell: ${{ inputs.shell }}
+          project: ${{ inputs.project }}
+          hookName: BuildInitialize
+          parametersJson: | 
+            {
+              "buildMode": "${{ inputs.buildMode }}",
+              "buildIt": "${{ steps.DetermineBuildProject.outputs.BuildIt }}"
```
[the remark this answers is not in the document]
FreundDK: Use boolean instead of string

mazhelez (dismissed):

Templates/AppSource App/.github/workflows/_BuildALGoProject.yaml
```diff
@@ -128,6 +120,19 @@ jobs:
...
+        uses: microsoft/AL-Go-Actions/RunHook@main
+        with:
+          shell: ${{ inputs.shell }}
+          project: ${{ inputs.project }}
+          hookName: BuildInitialize
+          parametersJson: | 
+            {
+              "buildMode": "${{ inputs.buildMode }}",
+              "buildIt": ${{ steps.DetermineBuildProject.outputs.BuildIt == 'True' }}
```
aholstrup1, line 133: Rather than passing BuildIt as a parameter, wouldn't it make more sense to just skip the hook altogether if BuildIt is false? Not sure what a hook would do in a job that doesnt build anyway. 

What do you think?

Templates/AppSource App/.github/workflows/_BuildALGoProject.yaml
```diff
@@ -128,6 +120,19 @@ jobs:
...
+      - name: Run Build Initialize hook
+        if: hashFiles(format('{0}/.AL-Go/BuildInitialize.ps1', inputs.project)) != ''
+        uses: microsoft/AL-Go-Actions/RunHook@main
+        with:
+          shell: ${{ inputs.shell }}
+          project: ${{ inputs.project }}
+          hookName: BuildInitialize
+          parametersJson: | 
+            {
+              "buildMode": "${{ inputs.buildMode }}",
```
aholstrup1, line 132: I'm wondering if buildmode should just be a standard parameter like project is

Templates/AppSource App/.github/workflows/_BuildALGoProject.yaml
freddydk: Well it actually will invoke the _BuildALGoProject (to copy the files), but it would be OK for my scenario.
I also added a Cleanup hook - with these two hooks, I can integrate Fkh to AL-Go and not need anything but that.
Obviously, I would copy some overrides to the .AL-Go folder during build, but I could modify those when you guys modify the build process and keep all changes centralized.

This is how my BuildInitialize and BuildCleanup looks:

```powershell
Param(
    [Hashtable] $parameters
)

# AL-Go Hook

$buildIt = $parameters.buildIt

if (-not $buildIt) {
    Write-Host "BuildIt is false, no need to copy fkh scripts"
    return
}

$ENV:FKH_BACKEND_URL = "https://fkh-freddydk-backend.azurewebsites.net/api"
$ENV:FKH_TIMEZONE = "Europe/Copenhagen"

Write-Host "Install Freddy's Kubernetes Helper & Business Central Development Tools"
dotnet tool install -g fkh --prerelease
dotnet tool install -g Microsoft.Dynamics.BusinessCentral.Development.Tools

fkh applyALGoOverrides --output $PSScriptRoot
```

and Cleanup:
```powershell
Param(
    [Hashtable] $parameters
)

# AL-Go Hook

$project = $parameters.project
$buildMode = $parameters.buildMode
$buildIt = $parameters.buildIt

if (-not $buildIt) {
    Write-Host "BuildIt is false, no need to remove container"
    return
}

$ENV:FKH_BACKEND_URL = "https://fkh-freddydk-backend.azurewebsites.net/api"
$ENV:FKH_TIMEZONE = "Europe/Copenhagen"

$parts = "$ENV:GITHUB_REPOSITORY". Split('/')
$containerName = "$($parts[0])-$($parts[1])-$($project)-$($buildMode)-$($ENV:GITHUB_RUN_ID)".ToLower() -replace "[^a-z0-9\-]"

fkh RemoveContainer --name $containerName --useOIDC
```

I can easily modify the PR to skip the hooks if buildIt is false - I am not going to use that.

Templates/AppSource App/.github/workflows/_BuildALGoProject.yaml
freddydk: Depends on whether hooks will be available in non-build workflows as well, but then project maybe shouldn't be a standard parameter? 
Let me know what you conclude?

PR #2293, opened 2026-06-26
Add support for protectedSettings to override lower priority settings

ChrisBlankDe:
### ❔What, Why & How

This pull request introduces a new `protectedSettings` mechanism that allows you to protect specific settings from being overridden by lower-priority sources in the AL-Go settings hierarchy. The implementation ensures that settings marked as protected can only be overridden by lower-priority sources if those sources also explicitly mark the same setting as protected. This applies to both standard and conditional settings, and integrates with the existing `overwriteSettings` mechanism. Documentation and schema files are updated to reflect this new feature.

**Protected settings**

* Added support for an `protectedSettings` array in settings files, enabling top-level settings to be protected from non-protected overrides by lower-priority sources. The merging logic in `MergeCustomObjectIntoOrderedDictionary` was updated to enforce this rule for both direct and conditional settings, and to ensure that array settings are always merged. [[1]](diffhunk://#diff-259d8325d7db8f6970bf404f6b56f1cc32a5280686f77571b6a4124ecc874532L19-R31) [[2]](diffhunk://#diff-259d8325d7db8f6970bf404f6b56f1cc32a5280686f77571b6a4124ecc874532L65-R87) [[3]](diffhunk://#diff-259d8325d7db8f6970bf404f6b56f1cc32a5280686f77571b6a4124ecc874532R499-R511) [[4]](diffhunk://#diff-259d8325d7db8f6970bf404f6b56f1cc32a5280686f77571b6a4124ecc874532L511-R540)

**Schema and documentation updates:**

* Updated `settings.schema.json` to add the `protectedSettings` property and clarified the behavior of `overwriteSettings` when protected settings are involved. [[1]](diffhunk://#diff-eb455665e386ecd7afac1dc52479c0dacaf6bc98de073f07cb8f3e08c10d3ae4L731-R739) [[2]](diffhunk://#diff-eb455665e386ecd7afac1dc52479c0dacaf6bc98de073f07cb8f3e08c10d3ae4L398-R398) [[3]](diffhunk://#diff-eb455665e386ecd7afac1dc52479c0dacaf6bc98de073f07cb8f3e08c10d3ae4L653-R653) [[4]](diffhunk://#diff-eb455665e386ecd7afac1dc52479c0dacaf6bc98de073f07cb8f3e08c10d3ae4L683-R683)
* Added detailed explanations, examples, and usage notes for `protectedSettings` in `RELEASENOTES.md` and `Scenarios/settings.md`. [[1]](diffhunk://#diff-310cf8dbf9f6a843674df7afb82cf592a81636c01f7c25221aba39cf86e8dcf1R1-R34) [[2]](diffhunk://#diff-ab2c7406f61eeb7a5bc9e0a07b6ddc964ef0415f77f24890b5e77acdb5f0a0f2R319-R395)

**Other improvements:**

* Added `protectedSettings` to the default settings returned by `GetDefaultSettings`.
* Minor parameter and comment cleanups for consistency in PowerShell scripts. [[1]](diffhunk://#diff-259d8325d7db8f6970bf404f6b56f1cc32a5280686f77571b6a4124ecc874532L112-R126) [[2]](diffhunk://#diff-259d8325d7db8f6970bf404f6b56f1cc32a5280686f77571b6a4124ecc874532L314-R328) [[3]](diffhunk://#diff-259d8325d7db8f6970bf404f6b56f1cc32a5280686f77571b6a4124ecc874532L336-R350) [[4]](diffhunk://#diff-259d8325d7db8f6970bf404f6b56f1cc32a5280686f77571b6a4124ecc874532L40-R46)

These changes make it easier to enforce organization- or repository-level settings and prevent accidental overrides from lower-priority sources, while maintaining flexibility for advanced scenarios.

### Use Case 1

We want to define conditionalSetting for different buildModes as ALGoOrgSettings as we need them in many Repos.
For example we have one BuildMode for MinVersion Validation:

```json
{
    "ConditionalSettings": [
        {
            "buildModes": [
                "minversion"
            ],
            "settings": {
                "artifact": "/sandbox/*//first",
                "nuGetFeedSelectMode": "EarliestMatching"
            }
        }
    ]
}
```
in the target repo its then enough to define the buildModes:
```json
{
    "buildModes": [
        "default",
        "minversion"
    ]
}
```

if we add the buildMode to any workflow in the repo/project the artifact get overwritten by the repo/project setting.
If we define the artifact setting in conditional setting as protected, this overwrite wouldnt happen.

### Use Case 2

Went to define a buildmode for testing isv solution against different localizations

```json
{
    "ConditionalSettings": [
        {
            "buildModes": [
                "at"
            ],
            "settings": {
                "country": "at"
            }
        }
    ]
}
```
in the target repo its then enough to define the buildModes:
```json
{
    "country": "de",
    "buildModes": [
        "default",
        "at"
    ]
}
```

the country of this build mode is overwritten by the country defined in the repo.
However, we can't omit the “country” definition from the repo settings because there are solutions that need to be built against ‘w1’ and others that need to be built against ‘de’.

Related to issue: #

### ✅ Checklist

- [x] Update RELEASENOTES.md
- [ ] Add telemetry


Actions/.Modules/ReadSettings.psm1
```diff
@@ -476,19 +491,27 @@ function ReadSettings {
...
             OutputDebug "Applying settings from $($settingsObject.Source) ($($settingsObject.Type))"
-            MergeCustomObjectIntoOrderedDictionary -dst $settings -src $settingsJson
+            $srcImportantSettings = @()
+            if ($settingsJson.PSObject.Properties.Name -contains "importantSettings") {
+                $srcImportantSettings = @($settingsJson.importantSettings)
+            }
+            MergeCustomObjectIntoOrderedDictionary -dst $settings -src $settingsJson -srcImportantSettings $srcImportantSettings -dstImportantSettings $currentImportantSettings
+            if ($settingsJson.PSObject.Properties.Name -contains "importantSettings") {
+                $currentImportantSettings = $srcImportantSettings
+            }
```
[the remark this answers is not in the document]
ChrisBlankDe: fixed

Tests/ReadSettings.Test.ps1
```diff
@@ -540,5 +540,358 @@ InModuleScope ReadSettings { # Allows testing of private functions
...
+        It 'Lower-priority important settings can override higher-priority important settings' {
+            Mock Write-Host { }
+            Mock Out-Host { }
+
+            Push-Location
+            $tempName = Join-Path ([System.IO.Path]::GetTempPath()) ([Guid]::NewGuid().ToString())
+            $githubFolder = Join-Path $tempName ".github"
+            $ALGoFolder = Join-Path $tempName $ALGoFolderName
+            $projectALGoFolder = Join-Path $tempName "Project/$ALGoFolderName"
+
```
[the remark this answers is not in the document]
ChrisBlankDe: fixed

Actions/.Modules/ReadSettings.psm1
```diff
@@ -508,7 +531,14 @@ function ReadSettings {
...
                             OutputDebug "Applying conditional settings for $($conditions -join ", ")"
-                            MergeCustomObjectIntoOrderedDictionary -dst $settings -src $conditionalSetting.settings
+                            $srcImportantSettings = @()
+                            if ($conditionalSetting.settings.PSObject.Properties.Name -contains "importantSettings") {
+                                $srcImportantSettings = @($conditionalSetting.settings.importantSettings)
+                            }
+                            MergeCustomObjectIntoOrderedDictionary -dst $settings -src $conditionalSetting.settings -srcImportantSettings $srcImportantSettings -dstImportantSettings $currentImportantSettings
+                            if ($conditionalSetting.settings.PSObject.Properties.Name -contains "importantSettings") {
+                                $currentImportantSettings = $srcImportantSettings
+                            }
```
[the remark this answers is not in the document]
ChrisBlankDe: fixed

Tests/ReadSettings.Test.ps1
```diff
@@ -540,5 +540,358 @@ InModuleScope ReadSettings { # Allows testing of private functions
...
+            New-Item $githubFolder -ItemType Directory | Out-Null
+            New-Item $projectALGoFolder -ItemType Directory | Out-Null
+
+            # Org settings: important settings includes "country", set country = "de"
+            @{ "importantSettings" = @("country"); "country" = "de" } | ConvertTo-Json -Depth 99 |
+            Set-Content -Path (Join-Path $githubFolder "AL-Go-Settings.json") -encoding utf8 -Force
+
+            # Project settings: try to override country = "ch"
+            @{ "country" = "ch" } | ConvertTo-Json -Depth 99 |
+            Set-Content -Path (Join-Path $projectALGoFolder "settings.json") -encoding utf8 -Force
```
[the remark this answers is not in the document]
ChrisBlankDe: fixed

mazhelez:
Interesting concept. We'd have to give it a bit of a thought, as it goes a bit against the way "important" settings are defined.
In AL-Go, project settings are higher in priority than org settings, for example.

For now, I can't tell I like `importantSettings` because it might mean a lot of things. How about `preventOverwrite`, or `protectedSettings`?

ChrisBlankDe:
I missed to add the use-case to description. added now. I think this will help understand what we're trying to achieve with this feature.

When naming it, we got inspired by the [!important](https://www.w3schools.com/css/css_important.asp) rule in CSS, since we want to implement a similar concept.

i wouldnt name it `preventOverwrite `since overwritting is still possible, if the overwriting source is also defined as important. And we already use that name for a special meachanism in handling AL-Go Settings.

Think about what you'd like to call it, and I'll be happy to change it.
In the end, the feature is what matters to me, not the name.

RELEASENOTES.md
```diff
@@ -1,3 +1,37 @@
+### Important settings protection
```

ChrisBlankDe: duplicate comment.

ChrisBlankDe: fixed

Actions/.Modules/ReadSettings.psm1
```diff
@@ -180,10 +193,10 @@ function GetDefaultSettings
         "configPackages"                                = @()
         "appSourceCopMandatoryAffixes"                  = @()
         "deliverToAppSource"                            = [ordered]@{
-            "mainAppFolder"                             = ""
-            "productId"                                 = ""
-            "includeDependencies"                       = @()
-            "continuousDelivery"                        = $false
+            "mainAppFolder"       = ""
```
mazhelez: Please leave the formatting as is.

Actions/.Modules/ReadSettings.psm1
```diff
@@ -62,15 +68,23 @@ function MergeCustomObjectIntoOrderedDictionary {
     # If the property exists in the source object, but is of a different type, throw an error
     # If the property exists in the source object:
     # If the property is an Object, call this function recursively to merge values
-    # If the property is an Object[], merge the arrays
+    # If the property is an Object[], merge the arrays (even if important - arrays always merge)
```
mazhelez, line 71: That's not intuitive, why would arrays always merge even the setting in the source is marked as important?

Actions/.Modules/ReadSettings.psm1
```diff
@@ -37,7 +43,7 @@ function MergeCustomObjectIntoOrderedDictionary {
     $src.PSObject.Properties.GetEnumerator() | ForEach-Object {
         $prop = $_.Name
 
-        # Skip overwriteSettings property as it's only used to remove settings from the destination object and is specific to the source object
+        # Skip overwriteSettings property as it's only used for configuration, not actual settings
```
mazhelez, line 46: Does `importantSettings` need to be transferred here? I see they are accumulated in `$currentImportantSettings`

mazhelez (commented):
Left some comments for you to look at.

Also, let's also address the name of the new setting. I understand the CSS origin, but the syntax is quite different.
Here are my suggestions:

1. `protectedSettings`
2. `enforcedSettings`

Actions/.Modules/ReadSettings.psm1
ChrisBlankDe: fixed.
code was autoformatted by vscode.
maybe we could add autoformat settings into the repo. to avoid this in future.

Actions/.Modules/ReadSettings.psm1
ChrisBlankDe: I think this is unnecessary from a technical standpoint, but I still think it makes sense to include them so that later on e.g., during debugging it’s clear why a setting was inherited in a certain way.
So far, we’ve been doing this with conditional settings as well. These remain part of the settings indefinitely, even after they’ve been completely resolved.

I've also considered no longer accumulating the $currentImportantSettings outside of the MergeCustomObjectIntoOrderedDictionary function, but instead just accessing the property within the function itself.
However, I think it's a cleaner design to keep the control state separate from the data state.

What do you think, keep the cleaner design or should we try to remove `$currentImportantSettings`?

Actions/.Modules/ReadSettings.psm1
ChrisBlankDe: Yeah, I wasn't sure about that either, but I couldn't think of a better solution.

I think it's important for arrays that the “important” values are preserved. For example, with the buildModes, the “important” ones should always be executed, but also it should be possible to add others.

If I were to change the behavior here so that they aren’t always merged, that would mean that whenever I want to add buildModes, I’d also have to set `overwriteSettings: ["buildModes"]` and include the earlier-level values in the overwrite. I think this is a bit clunky, since, in my opinion, “overwrite” should only be used to remove values from earlier levels.

freddydk: I think the biggest problem with this is that some settings are dependent of each other - shell, runs-on, useCompilerFolder, etc. and it becomes very hard to understand why a setting, which I explicitly set in my project now suddenly doesn't work because somebody has added a higher level importantSettings.

I would rather have a setup where a higher level could specify importantSettings - but it didn't affect the merge of settings, but instead - it would give an error when the setting was overridden lower down - then people can fix this - meaning a non-intrusive behavior really.

Another thing we could implement was a dump after readsettings, where it would print out specifically which settings where read from which files:

artifact - taken from project setting xxxx
runs-on - taken from .github/AL-Go-settings.json
.....

That would help people setup the right settings.
Adding overwrite and important settings seems to only complicate things, sorry.

ChrisBlankDe: @freddydk Your comment isn't about how to handle arrays, but is more of a general point, right?

> it would give an error when the setting was overridden lower down

If we look at Use Case 2 from the PR description, this would be disadvantageous. If I have to define the country in the repo again for the build mode, I might as well skip entering it in the organization settings altogether. Furthermore, I think that anyone who is allowed to edit earlier settings (org + repo) should have the foresight and sense of responsibility to be able to override the repo settings.

> Another thing we could implement was a dump after readsettings, where it would print out specifically which settings where read from which files:
> 
> artifact - taken from project setting xxxx
> runs-on - taken from .github/AL-Go-settings.json

i could do that, @mazhelez should i?
print always to the logs or just with `::debug::` annotation?

mazhelez:

<details>
<summary>Example</summary>

```markdown

- Your change here ✅

## v9.2

- Already released changes
```

</details>

ChrisBlankDe:
> Also, let's also address the name of the new setting. I understand the CSS origin, but the syntax is quite different. Here are my suggestions:
> 
> 1. `protectedSettings`
> 2. `enforcedSettings`

renamed everything to protectedSettings

Actions/.Modules/ReadSettings.psm1
mazhelez: In my opinion, `protectedSettings` isn't really an AL-Go setting. You can consider it a meta-setting: it's needed only while calculating the settings themselves.

 I think keeping `currentImportantSettings` is fine but definitely add some logging so that it is clear why a setting was calculated in a certain way.

RELEASENOTES.md
```diff
@@ -42,6 +75,7 @@ The `trackALAlertsInGitHub` setting now also works when `workspaceCompilation` (
 
 ## v9.1
 
+
```
mazhelez, line 78:
```suggestion
```

Actions/.Modules/ReadSettings.psm1
```diff
@@ -71,6 +77,13 @@ function MergeCustomObjectIntoOrderedDictionary {
             $srcProp = $src."$prop"
             $dstPropType = $dstProp.GetType().Name
             $srcPropType = $srcProp.GetType().Name
+
+            # For non-array properties: skip if this setting is marked as protected from higher priority source,
+            # unless the lower-priority source also marks this property as protected.
+            if ($dstProtectedSettings -contains $prop -and $srcPropType -ne "Object[]" -and $srcProtectedSettings -notcontains $prop) {
```
mazhelez, line 83: I don't get why arrays are excluded here.

Actions/.Modules/ReadSettings.psm1
ChrisBlankDe: As we discussed off-GitHub we should always write setting tracing to the log (not only in debug mode)
Also Arrays should be handled like every other datatype.

PR #2300, opened 2026-07-03
Make failOn: newWarning work with workspace compilation

aholstrup1:
## Why

The `failOn: newWarning` setting fails a pull request when it introduces new AL compiler warnings, compared against the last successful build of the target branch. It only ever worked for container / compiler-folder builds. With `workspaceCompilation` enabled it silently did nothing, so repositories using workspace compilation (including BCApps) had no protection against newly introduced warnings.

## What changed

The root cause was that the new-warning check (`Test-ForNewWarnings`) only ran in the `RunPipeline` action, but with workspace compilation the apps are compiled by the separate `CompileApps` action and `RunPipeline` runs against prebuilt apps.

- **Run the check where compilation happens.** `Test-ForNewWarnings` is now invoked from `CompileApps` when workspace compilation is enabled, and `RunPipeline` skips it in that mode to avoid a double run. `CheckForWarningsUtils.psm1` moved from `Actions/RunPipeline/` to `Actions/.Modules/` since it is now shared.
- **Preserve the compiler output for the baseline artifact.** BcContainerHelper's `Run-AlPipeline` deletes its `buildOutputFile` at startup (unconditionally, before it looks at whether anything needs building). In workspace mode that file is the compiler output written by `CompileApps` and contains the warnings, so `RunPipeline` now passes an empty `buildOutputFile` to keep it intact for the uploaded build-output artifact that future PRs compare against.

Two robustness issues surfaced while validating this against BCApps and are fixed here too:

- **Parse the workspace warning format.** Workspace compilation writes raw AL compiler output (`File(line,col): warning ID: description`) rather than the GitHub Actions annotation format (`::warning file=...::`). The parser now handles both, so warnings are actually detected under workspace compilation.
- **Ignore build-version churn in warning text.** Some diagnostics (e.g. `AL0523`) embed a referenced app's version in the message. AL-Go stamps build/revision from the run number, so that version differs between the baseline and PR builds and made unchanged warnings look new. Four-part version numbers are now normalized before comparison, while the original description is still shown in the error annotation.

## Testing

Added Pester coverage in `Tests/CheckForWarningsUtils.Test.ps1` for both warning formats, the new-vs-removed warning cases, the version-normalization case, and the pull-request/`failOn` guards.

Validated end to end on a real multi-project repo (microsoft/BCApps, workspace compilation enabled): a PR that introduces one unused-variable warning fails with "New warnings were introduced" via the `CompileApps` path, while projects whose only diff was version-churn in warning text pass instead of failing falsely.

## Note for reviewers

There is a one-time transition effect: the first PRs after this merges compare against baseline artifacts produced by the previous code, so for workspace publishing projects those baselines may be warning-less until `main` is rebuilt once. It self-heals after the next successful `main` build.


spetersenms (dismissed):

RELEASENOTES.md
```diff
@@ -1,3 +1,9 @@
+### `failOn: newWarning` now works with workspace compilation
```
spetersenms, line 9: A new version of AL-Go (v9.1) has been released.

spetersenms, line 9: A new version of AL-Go (v9.1) has been released.

mazhelez:
@copilot+claude-opus-4.8 resolve the merge conflicts in this pull request: move the release notes at the top of the file

mazhelez (dismissed):

Tests/CheckForWarningsUtils.Test.ps1
```diff
@@ -0,0 +1,201 @@
...
+                # File separators are normalized to forward slashes regardless of source format.
+                $warnings[0].File | Should -Be 'Codeunits/MyCodeunit.al'
+                $warnings[0].Description | Should -Be 'The variable is never used.'
+                $warnings[0].Line | Should -Be '10'
+                $warnings[0].Col | Should -Be '5'
+            }
+        }
+    }
+
+    Context 'Compare-Files' {
```
mazhelez: Minor: naming

`Compare-File` is quite broad name. Consider another, a bit more specific name.

mazhelez (dismissed):

mazhelez (approved):

spetersenms (approved):

PR #2301, opened 2026-07-03
Provide package cache to aldoc when building reference documentation

### LINKED ISSUE #883: Deploy Reference Documentation (ALDoc) action missing symbols and xml documentation not showing
I added ALDoc to my `settings.json` file (with `be`as supported country):

image

The CI/CD action deployed the generated documentation:

image

But the static site does not have all information I expected:

1. __MissingTypeSymbol__

> The symbols for the parameter of a function isn't found

This is my procedure in AL:
```AL
    procedure SetStatementNoWithYearPrefix(var CODAStatement: Record "CODA Statement")
```

And this is the output on the generated Docs:

image

2. AL XML Documentation does not show

This is the doccumentation in AL:
```AL
    /// <summary>
    /// Set Statement No. With Year Prefix.
    /// Banks reset Statement No. every year. To prevent duplicate Statement No. in BC, we need to add year prefix to Statement No.
    /// </summary>
    /// <param name="CODAStatement">VAR Record "CODA Statement".</param>
```

And this is the output on the generated Docs:

image

### LINKED ISSUE #2302: [Bug]: AlDoc does not use --packagecache when building reference documentation
### AL-Go version

preview

### Describe the issue

When building reference documentation, AlDoc does not use the --packageacache flag, meaning it does not have access to dependencies. 
This does not crash the code, but throws warnings about missing modules when generating documentation for an app depending on some other app.

### Expected behavior

AlDoc should not throw warnings about missing modules and instead use the --packagecache flag to pass dependencies (which are already downloaded in the action, but not used)

### Steps to reproduce

Run the flow to generate reference documentation in a repo with one app depending on some other app.

_No response_

spetersenms:
aldoc build was invoked with only --source, so it fell back to its default package cache ({source}/.alpackages) which does not exist for artifact-based apps. As a result every reference to another module was reported as 'Referenced module not loaded' and the corresponding links in the generated documentation were left unresolved.

Populate a package cache with all apps and their dependencies (the latter were already downloaded and computed by CalculateProjectsAndApps but never used) and pass it to aldoc via --packagecache so cross-module references resolve.

Fixes #2302 

### ✅ Checklist

- [x] Update RELEASENOTES.md
- [ ] Add telemetry


RELEASENOTES.md
```diff
@@ -68,6 +68,7 @@ The `DownloadProjectDependencies` action now downloads only artifacts from depen
 - Issue 2235 - Workspace compilation: only the first `customCodeCops` entry resolved when multiple relative paths were configured. Relative `customCodeCops` paths are now resolved against the project folder before being passed to the compiler.
 - Issue 2265 - Creating a Performance Test App fails on Linux due to case-sensitive path lookup for the Performance Toolkit sample app
 - Issue 2284 - GitHub App authentication fails with `401 (Unauthorized)` on runners with minor clock drift. The JWT `iat` claim is now backdated by 60 seconds instead of 10, as recommended by GitHub, to tolerate runners whose clock runs slightly ahead of GitHub.
+- Issue 2302 - AlDoc does not use --packagecache when building reference documentation
```
aholstrup1: Need to be updated now that we shipped 9.1

aholstrup1 (commented):
Releasenote needs to be updated but otherwise looks fine

aholstrup1 (approved):

mazhelez (approved):

PR #2304, opened 2026-07-06
Support framework-dependent (flat) AL Language extension layout

aholstrup1:
### ❔What, Why & How

Newer, framework-dependent AL Language extensions (for example marketplace-packaged builds) ship a **flat** internal layout: `altool`, `alc`, `aldoc`, and the analyzer DLLs all live directly under `extension/bin`, with no `win32`/`linux` or `Analyzers` subfolder. Setting such a VSIX via the `vsixFile` setting caused **workspace compilation** to fail with `Could not find AL tool in the compiler folder`, even though the same VSIX worked fine for container and `useCompilerFolder` builds (BcContainerHelper already handles both layouts).

This PR makes AL-Go's own path resolution layout-aware, mirroring how BcContainerHelper resolves these tools. Three spots assumed the nested layout and now fall back to the flat `bin` folder:

- **`Get-ALTool`** (`CompileFromWorkspace.psm1`): probes `bin/win32/altool.exe` (or `bin/linux/altool`), then falls back to the flat `bin/altool(.exe)`.
- **`Get-CustomAnalyzers`** (`CompileFromWorkspace.psm1`): URL-based `customCodeCops` download to `bin/Analyzers` when present, otherwise to the flat `bin`.
- **`DownloadAlDoc`** (`BuildReferenceDocumentation`): resolves `aldoc` in `bin/win32|linux` then the flat `bin`.

**Worth noting for reviewers:**
- The `aldoc` fix is pre-emptive. That path is fed by the VSIX bundled in the BC artifact (not the user's `vsixFile`), so it is not exercised by the `vsixFile` scenario today, but it carried the identical latent assumption and was cheap to fix alongside.
- This does **not** cover built-in analyzers (`CodeCop`/`PTECop`/etc.). Those are passed to `altool` by shorthand name, and `altool` itself resolves them to `<altoolDir>/../Analyzers`, which is wrong for the flat layout. That is an `altool` bug (its shorthand resolver is inconsistent with how it resolves `${CodeCop}`/custom analyzers) and is being tracked separately against the AL Language / compiler team. The fixes here are validated end-to-end up to that point: with them applied, a flat VSIX now gets past tool discovery and fails only on `altool`'s analyzer path.

### ✅ Checklist

- [x] Update RELEASENOTES.md
- [ ] Add telemetry


RELEASENOTES.md
```diff
@@ -1,3 +1,7 @@
+### Workspace compilation supports framework-dependent AL Language extensions
```

mazhelez (dismissed):

mazhelez (approved):

ankosiorek (approved):

PR #2343, opened 2026-08-06
RunTests action

spetersenms:
## Summary

Adds the preview `useSeparateTestAction` setting, defaulting to `false`, which moves normal test execution from `RunPipeline` into a new `RunTests` action using `al runtests`.

## Flow

1. `RunPipeline` builds and publishes the apps, creates the container, and keeps it alive.
2. It exports the container name, generated credentials, and a runtime signal.
3. The workflow conditionally invokes `RunTests`.
4. `RunTests` selects the test apps, executes them with AlTool, and produces JUnit results for the existing analysis step.
5. The existing cleanup step removes the container.

## Notes

- Only normal tests are moved. BCPT, page scripting, and multi-country tests remain in `RunPipeline`.
- `runTestsInAllInstalledTestApps` uses the existing `installTestAppsJson` handoff rather than discovering every installed server app.
- The optional `testType` setting supports `UnitTest`, `IntegrationTest`, and `Uncategorized`.
- A custom `RunTestsInBcContainer` override can still replace the built-in AlTool runner.
- BCH is still used for container creation, app installation and metadata, server and company discovery, test enumeration, and event-log collection. AlTool owns the actual test execution.

### ✅ Checklist

- [x] Update RELEASENOTES.md
- [ ] Add telemetry


Actions/RunTests/RunTests.ps1
```diff
@@ -0,0 +1,101 @@
...
+    The GitHub token running the action. It is exposed to test override scripts as _token.
+.PARAMETER project
+    Project folder.
+.PARAMETER installTestAppsJson
+    A path to a JSON-formatted list of test apps (produced by previous jobs) to run tests in.
+.EXAMPLE
+    RunTests.ps1 -project 'MyProject'
+#>
+
+# Make the token available to RunTestsInBcContainer overrides.
```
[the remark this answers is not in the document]
spetersenms: Implemented centrally in the run action script.

mazhelez:

<details>
<summary>Example</summary>

```markdown

- Your change here ✅

## v9.2

- Already released changes
```

</details>

Actions/.Modules/ReadSettings.psm1
```diff
@@ -174,6 +174,8 @@ function GetDefaultSettings
         "doNotBuildTests"                               = $false
         "doNotPerformUpgrade"                           = $false
         "doNotRunTests"                                 = $false
+        "useSeparateTestAction"                         = $false
```
mazhelez, line 177: Can we have a complex setting, similar to `workspaceCompilation`, that look like this

```json
{
    "enabled": false,
    "testType": "",
     "<other potential settings in the future>"
}
```

mazhelez (commented):
Mid-way review.

How about adding an e2e test(s) as well?

Actions/.Modules/ReadSettings.psm1
spetersenms: Agree on both. I'll push the changes when your review is done.

Actions/RunPipeline/RunPipeline.ps1
```diff
@@ -473,6 +487,30 @@ try {
...
+        # Surface a reusable credential so RunTests can reconnect to the kept-alive container.
+        $containerCredential = New-KeepAliveContainerCredential
+        $runAlPipelineParams["credential"] = $containerCredential
+
+        $containerCredentialPassword = $containerCredential.GetNetworkCredential().Password
+        $containerCredentialJson = @{ "username" = $containerCredential.UserName; "password" = $containerCredentialPassword } | ConvertTo-Json -Compress
+        $containerCredentialBase64 = [System.Convert]::ToBase64String([System.Text.Encoding]::UTF8.GetBytes($containerCredentialJson))
+        Write-Host "::add-mask::$containerCredentialPassword"
+        Write-Host "::add-mask::$containerCredentialBase64"
+        Add-Content -Encoding UTF8 -Path $env:GITHUB_ENV -Value "containerCredential=$containerCredentialBase64"
```
aholstrup1, line 508: This part doesn't need to be conditional does it? I guess we could always just create the credential before calling Run-ALPipeline. We'll probably need to do that soon anyway so maybe we could just make that change now?

Actions/RunTests/AlToolTestRunner.psm1
```diff
@@ -0,0 +1,945 @@
...
+<#
+.SYNOPSIS
+    Ensures the `al` CLI is available on PATH, installing the prerelease dotnet global tool.
+.DESCRIPTION
+    Installs the AL developer tools when unavailable. A named mutex prevents concurrent jobs from
+    modifying the shared tool store at the same time.
+.OUTPUTS
+    [string] The resolved `al` version string.
+#>
+function Install-AlTool {
```
aholstrup1, line 85: Could we move this function out into AL-Go helper or something? And maybe make it work a bit like https://github.com/microsoft/AL-Go/blob/f4387d14d67b25dee093a9bba6f09aea741c132e/Actions/AL-Go-Helper.ps1#L423-L426 where we can call it multiple times in a workflow but only run the installation once? 

I think the ALTool is something we'll be using more and more going forward. Would be nice to establish a good pattern for this that goes beyond running tests

Actions/RunTests/AlToolTestRunner.psm1
```diff
@@ -0,0 +1,945 @@
...
+            $installResult = Invoke-AlNativeCommand -FilePath "dotnet" -ArgumentList @(
+                "tool", "install", $script:AlToolPackageId, "--global", "--prerelease"
+            )
+            $installResult.Output | ForEach-Object { Write-Host $_ }
+            if ($installResult.ExitCode -ne 0) {
+                # A concurrent job may have installed it first; treat as success if `al` now resolves,
+                # otherwise fall back to an update.
+                if ($null -eq (Get-Command al -ErrorAction SilentlyContinue)) {
+                    $updateResult = Invoke-AlNativeCommand -FilePath "dotnet" -ArgumentList @(
+                        "tool", "update", $script:AlToolPackageId, "--global", "--prerelease"
```
aholstrup1, line 120: I'm not sure I would install it globally on the machine. I think dotnet also has the option to install into a tool path. I think that's the safer option to avoid situations where multiple runners on the same host are trying to update it or access it at the same time

Actions/RunTests/AlToolTestRunner.psm1
```diff
@@ -0,0 +1,945 @@
...
+        $allPassed = [bool] $batch.Succeeded
+
+        $idx = 0
+        foreach ($cu in $codeunits) {
+            $idx++
+            $methods = @($cu.Tests | ForEach-Object { "$_" })
+            $cuResults = $batchResults["$($cu.Id)"]
+            if ($null -eq $cuResults) { $cuResults = @() }
+
+            $failed = Add-JUnitTestSuite -Doc $doc -TestSuitesNode $suites -Codeunit $cu `
```
aholstrup1, line 919: Could ALTool support this natively? I'm thinking... AL-Go is probably not the only consumer that might be interested in getting test results in a well-known format

Actions/RunTests/AlToolTestRunner.psm1
```diff
@@ -0,0 +1,945 @@
...
+    Tenant used for test enumeration.
+.PARAMETER TestType
+    Optional platform test type. Supported values are UnitTest, IntegrationTest, and Uncategorized.
+    Blank enumerates all test types.
+.PARAMETER DisabledTests
+    Hashtable entries describing test methods or codeunits excluded from the run.
+.OUTPUTS
+    [object[]] Codeunit objects with .Id, .Name, .Tests (enabled method name array).
+#>
+function Get-AlToolTestCodeunits {
```
aholstrup1, line 280: I'm not sure I understand the contract between AL-Go and the AL tool for running tests. In my mind we should be able to supply the App ID for the test app, the test type and a list of disabled tests and then the ALTool should be able to figure out which tests to run from there. 

Right now, we seem to be doing a lot of parsing in AL-Go to determine the exact list of codeunits and methods that should run. IMO that seems more like the responsibility of the ALTool 🤷‍♂️ My concern is that AL-Go is just one consumer of the ALTool. Anyone else who want to run tests with the ALTool would need to implement all this logic as well.

aholstrup1 (commented):
A few comments here and there but I like where this is heading! :)

Actions/RunPipeline/RunPipeline.ps1
spetersenms: Yeah I guess so. It does not really matter much for this feature, but if you prefer I am fine with removing the condition.

Actions/RunTests/AlToolTestRunner.psm1
spetersenms: Yeah good point, I'll make it more general and move it to the helper module.

Actions/RunTests/AlToolTestRunner.psm1
spetersenms: Good point, I agree.

Actions/RunTests/AlToolTestRunner.psm1
spetersenms: Discussed offline.

Actions/RunTests/AlToolTestRunner.psm1
spetersenms: Discussed offline.

